PR #8, opened 2020-07-16
Add build registry instructions

### LINKED ISSUE #7: Move build_registry.go in to a separate directory to differentiate the purpose
In addition, add instructions in README on how to generate registry files. 
Remove the executable from the repo.

zyiou:
Fixes #7


srikartati:
> To fix issue [#7 ](url)

Use the following notation: Fixes #7  
This will automatically close the issue when PR is closed.

.gitignore
```diff
@@ -1,5 +1,3 @@
-pkg/registry/build_registry
```
srikartati, line 1: We still need this. Let's say user run build_registry.go to generate registry go files on his local setup, we do not want him to checkin the binary.

README.md
```diff
@@ -6,6 +6,16 @@ go-ipfix is an IPFIX library that can be used to implement an IPFIX exporter, wh
...
 
+## Build Registry
+To build the registry from [IANA registry](https://www.iana.org/assignments/ipfix/ipfix.xhtml) or [Antrea registry](pkg/registry/registry_antrea.csv), run
+```
+go run pkg/registry/build_registry/build_registry.go [REGISTRY_NAME]
+# REGISTRY_NAME: "Antrea", "IANA", ""(build both registries)
+```
+it will generate two files: `pkg/registry/registry_antrea.go` and/or `pkg/registry/registry_IANA.go` to enable local registry loading functions.
+
+Please make sure to reload registry if there are changes in IANA/Antrea registry.
```
srikartati, line 17: Making it explicit is better that is re-execute build_registry.go to account for changes in either registry and regenerate those go files.

pkg/registry/build_registry/build_registry.go
```diff
@@ -38,12 +38,12 @@ func initIANARegistry() {
 		klog.Errorf("main: %v", error)
 	}
 
-	registryFileName := "registry_IANA.go"
+	registryFileName := "./pkg/registry/registry_IANA.go"
```
srikartati, line 41: interesting.. how is the working directory root of repo and not pkg/registry/build_registry

zyiou: I tried that if we run the `go run` command in the root of repo, the working directory will change from where build_registry.go is to the root of repo. It is because of the way that os package get working directory.

srikartati: Oh ok. So we are restricting the user to run from root. Seems to be hard restriction. Maybe we could use filepath to get to the root of repo by removing pkg/registry/build_registry and use that, so it will work if we run the file from any place? https://stackoverflow.com/questions/31873396/is-it-possible-to-get-the-current-root-of-package-structure-as-a-string-in-golan

zyiou: Sure. That will be better. Thanks!

pkg/registry/build_registry/build_registry.go
```diff
@@ -78,17 +83,21 @@ func (registry *ianaRegistry) LoadRegistry() {
 }
 
 func initAntreaRegistry() {
-	fileName := "registry_antrea.csv"
+	// get root of current package
+	_, base, _, _ := runtime.Caller(0)
+	rootPath := filepath.Dir(base)
```
srikartati, line 88: rootPath may be mistook as root of the repo. Change the name to fileDir or basePath?

srikartati:
Looks like we should remove the signature option.

PR #13, opened 2020-08-11
Support IPFIX collecting process

### LINKED ISSUE #3: Implement IPFIX mediator as per RFC7119
Following the RFC below, we need to implement IPFIX mediator like IPFIX exporter.
https://tools.ietf.org/html/rfc7119

This will be needed if we want to build IPFIX based flow aggregator to aggregate/enhance flow records and send them to an IPFIX collector.

zyiou:
Partially fixes #3


pkg/collector/process.go
```diff
@@ -0,0 +1,283 @@
...
+	if _, exists := cp.templatesMap[obsDomainID]; !exists {
+		cp.templatesMap[obsDomainID] = make(map[uint16][]*TemplateField)
+	}
+	cp.templatesMap[obsDomainID][templateID] = fields
+	cp.templatesLock.Unlock()
+	// template lifetime management
+	if cp.address.Network() == "tcp" {
+		return
+	}
+	if cp.templateTTL == 0 {
```
srikartati, line 230: space here and comment that you are doing this for UDP will be useful.

pkg/collector/process.go
```diff
@@ -0,0 +1,283 @@
...
+	return fields, nil
+}
+
+func (cp *collectingProcess) addTemplate(obsDomainID uint32, templateID uint16, fields []*TemplateField) {
+	cp.templatesLock.Lock()
+	if _, exists := cp.templatesMap[obsDomainID]; !exists {
+		cp.templatesMap[obsDomainID] = make(map[uint16][]*TemplateField)
+	}
+	cp.templatesMap[obsDomainID][templateID] = fields
+	cp.templatesLock.Unlock()
```
srikartati, line 225: space here too.

pkg/collector/process.go
```diff
@@ -0,0 +1,283 @@
...
+	cp.templatesLock.Lock()
+	delete(cp.templatesMap[obsDomainID], templateID)
+	cp.templatesLock.Unlock()
+}
+
+func (cp *collectingProcess) getDataType(templateField *TemplateField) entities.IEDataType {
+	var registry registry.Registry
+	if templateField.EnterpriseID == 0 { // IANA Registry
+		registry = cp.ianaRegistry
+	} else if templateField.EnterpriseID == 55829 { // Antrea Registry
```
srikartati, line 266: making 55829 a constant would be better.

pkg/collector/process.go
```diff
@@ -0,0 +1,283 @@
...
+		registry = cp.antreaRegistry
+	}
+	fieldName, err := registry.GetIENameFromID(templateField.ElementID)
+	if err != nil {
+		klog.Errorf("Information Element with id %d cannot be found.", templateField.ElementID)
+	}
+	ie, _ := registry.GetInfoElement(fieldName)
```
srikartati, line 273: May be defining a single map with key as (entID, elementID) and value as information elements better than having a nameID map?
We will have two separate maps: elementName to element, (entID, elementID) to element. This is better than making them different. 
In future, we can turn this in to k8s indexer.

pkg/collector/process_test.go
```diff
@@ -0,0 +1,225 @@
+package collector
+
+import (
+	"bytes"
+	"github.com/stretchr/testify/assert"
```
srikartati, line 5: same as above

pkg/collector/process.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+// data struct of processed message
+type Packet struct {
+	Version      uint16
+	BufferLength uint16
+	SeqNumber    uint32
+	ObsDomainID  uint32
+	ExportTime   uint32
+	FlowSets     interface{}
```
srikartati, line 57: why not just "Sets"?

pkg/collector/process.go
```diff
@@ -0,0 +1,283 @@
...
+	// number of workers
+	workerNum int
+}
+
+type TCPCollectingProcess struct {
+	*collectingProcess
+}
+
+// data struct of processed message
+type Packet struct {
```
srikartati, line 51: Message is a better name? Also defining this in entities.Message would be good.

pkg/collector/process.go
```diff
@@ -0,0 +1,283 @@
...
+			packets:        make([]*Packet, 0),
+		},
+		workerList: workerList,
+		workerPool: workerPool,
+		workerNum:  workerNum,
+	}
+	return collectProc, nil
+}
+
+func (cp *collectingProcess) DecodeMessage(msgBuffer *bytes.Buffer) (*Packet, error) {
```
srikartati, line 136: DecodePacket better name for function as it essentially decodes packet and returns a message.

pkg/collector/process.go
```diff
@@ -0,0 +1,283 @@
...
+	*collectingProcess
+	// list of udp workers to process message
+	workerList []*Worker
+	// udp worker pool to receive and process message
+	workerPool chan chan *bytes.Buffer
+	// number of workers
+	workerNum int
+}
+
+type TCPCollectingProcess struct {
```
srikartati, line 46: could we define this and corresponding init in tcp.go? Same for udp process.

pkg/collector/process.go
zyiou: I was trying to differentiating it from Set. Since we will use Set as the interface, using Sets makes more sense.

pkg/collector/process.go
```diff
@@ -0,0 +1,283 @@
...
+	Version      uint16
+	BufferLength uint16
+	SeqNumber    uint32
+	ObsDomainID  uint32
+	ExportTime   uint32
+	FlowSets     interface{}
+}
+
+// Shared header fields for TemplateFlowSet and DataFlowSet
+type FlowSetHeader struct {
```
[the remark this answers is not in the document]
zyiou: Agreed. Thanks!

pkg/collector/process.go
```diff
@@ -0,0 +1,283 @@
...
+	EnterpriseID  uint32
+}
+
+type DataField struct {
+	DataType entities.IEDataType
+	Value    bytes.Buffer
+}
+
+func InitTCPCollectingProcess(address net.Addr, maxBufferSize uint16) (*TCPCollectingProcess, error) {
+	ianaReg := registry.NewIanaRegistry()
```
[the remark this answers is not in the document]
zyiou: you mean that we can let users to input the registry into collecting process rather than loading inside the init function?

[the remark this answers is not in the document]
srikartati: Hi Yiou, I meant like this: https://github.com/vmware-tanzu/antrea/blob/master/pkg/agent/flowexporter/exporter/exporter.go#L162
Loading registries should be independent of processes and let users initialize them.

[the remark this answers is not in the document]
zyiou: Got it. Thx!

zyiou:
After trials I found that keeping the workers for udp is still a better way to do. Reason being that different from tcp connection, for udp the Golang net package can distinguish from different exporters only when `ReadFromUDP` returns source address. Then I will create a worker to handle all the records sent from that connection (instead of using a fixed number of workers). The connection (worker) is then kinda manually kept in a map.

pkg/collector/tcp.go
```diff
@@ -0,0 +1,75 @@
...
+			if err == io.EOF {
+				klog.Infof("Connection %s has been closed.", cp.address.String())
+			} else {
+				klog.Errorf("Error in collecting process: %v", err)
+			}
+			break out
+		}
+		klog.Infof("Receiving %d bytes from %s", size, cp.address.String())
+		packet := make([]byte, size)
+		copy(packet, buff[0:size])
```
srikartati, line 56: why is copy needed again? Can't we use the buff?

pkg/collector/tcp.go
```diff
@@ -0,0 +1,75 @@
...
+			conn, err := listener.Accept()
+			if err != nil {
+				klog.Errorf("Cannot start collecting process on %s: %v", cp.address.String(), err)
+				return
+			}
+			go cp.handleConnection(conn)
+		}
+	}()
+	select {
+	case <-cp.stopChan:
```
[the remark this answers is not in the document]
srikartati: Can we pass this as parameter in go routine and close the connections gracefully? Use waitgroup to wait for go routines.

pkg/collector/process.go
```diff
@@ -0,0 +1,245 @@
...
+	// server information
+	address net.Addr
+	// maximum buffer size to read the record
+	maxBufferSize uint16
+	// chanel to receive stop information
+	stopChan chan bool
+	// packet list
+	messages []*entities.Message
+	// workers to deal with clients (client address -> packet channel)
+	workers map[string]*worker
```
srikartati, line 48: workers -> udpClients?

srikartati (commented):
Reviewed tcp and udp code. Both are using go routines in a way.
As there is no concept of worker pool, may be using the name worker is not advisable. udpConnHandler/udpHandler is probably an apt name.
In the case of tcp, we will need a mechanism to close the connections if the collector process is abruptly stopped.

pkg/collector/process.go
```diff
@@ -0,0 +1,245 @@
...
+
+	"github.com/vmware/go-ipfix/pkg/config"
+	"github.com/vmware/go-ipfix/pkg/entities"
+)
+
+type collectingProcess struct {
+	// for each obsDomainID, there is a map of templates
+	templatesMap map[uint32]map[uint16][]*templateField
+	// templatesLock allows multiple readers or one writer at the same time
+	templatesLock *sync.RWMutex
```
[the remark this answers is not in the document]
zyiou: I think we need to pass the pointer to lock and unlock method (https://golang.org/pkg/sync/#RWMutex). It is not shared across instances.

pkg/collector/process.go
zyiou: Makes sense. Thanks!

pkg/collector/tcp.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"bytes"
+	"io"
+	"net"
+
+	"k8s.io/klog/v2"
+)
+
+// number of connections
+var connCount = 0
```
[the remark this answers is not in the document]
zyiou: The only use of this variable is that it helps with unit tests for multiple clients handling. Do you think it is necessary to have a method to tell how many clients are connected?

pkg/collector/tcp.go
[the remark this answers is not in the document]
zyiou: Agreed that we don't need select here and I need to implement a graceful way to close connections.

pkg/collector/tcp.go
zyiou: Makes sense, I was thinking that can help cut the unused bytes. Will remove copy then

pkg/collector/udp.go
```diff
@@ -0,0 +1,110 @@
...
+					klog.Infof("Connection %s has been closed.", cp.address.String())
+				} else {
+					klog.Errorf("Error in collecting process: %v", err)
+				}
+				return
+			}
+			packet := make([]byte, size)
+			copy(packet, buff)
+			if _, exist := cp.workers[address.String()]; !exist {
+				worker := createWorker(address.String(), cp.decodePacket)
```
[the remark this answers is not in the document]
zyiou: That's a good suggestion. If we achieve packetChan and errChan, for tcp we will be able to pass stop message to goroutine to stop gracefully. I will have a try. Thanks!

pkg/collector/udp.go
```diff
@@ -0,0 +1,110 @@
...
+	return &worker{
+		address:    address,
+		packetChan: make(chan *bytes.Buffer),
+		errChan:    make(chan bool),
+		decodeFunc: decodeFunc,
+	}
+}
+
+func (w *worker) start() {
+	go func() {
```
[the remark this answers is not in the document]
zyiou: we can exit by calling stop() function. yes we can have a timeout to remove corresponding udp connection. how long do you suggest as timeout? Maybe 10 min?

pkg/collector/process.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+}
+
+func (cp *collectingProcess) deleteTemplate(obsDomainID uint32, templateID uint16) {
+	cp.templatesLock.Lock()
+	defer cp.templatesLock.Unlock()
+	delete(cp.templatesMap[obsDomainID], templateID)
+}
+
+func decode(buffer io.Reader, output ...interface{}) error {
```
srikartati, line 214: Comments will be good for this function. output->outputs?

pkg/entities/message.go
```diff
@@ -45,3 +55,40 @@ func (m *MsgBuffer) GetDataRecFlag() bool {
 func (m *MsgBuffer) SetDataRecFlag(flag bool) {
 	m.dataRecFlag = flag
 }
+
+type templateMessage struct {
```
srikartati, line 59: can this abstraction be moved as templateSet and dataSet? As message can have both data and templates? In addition, maybe its better to use decodedTemplateSet and decodedDataSet to differentiate.

srikartati (commented):
Some comments..Still reviewing. Thanks.

pkg/collector/tcp.go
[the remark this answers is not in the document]
srikartati: Maybe this can be used by the user of the library. Better to keep it I guess.

pkg/collector/udp.go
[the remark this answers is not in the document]
srikartati: Yes.. how about using the same timeout as templateRefreshTimeout? We expect udp exporter to send the refreshed template.

pkg/collector/process.go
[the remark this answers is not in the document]
zyiou: Yes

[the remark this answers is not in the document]
srikartati: Oh ok. It should not matter whether the receiver is a pointer or value. You can call methods from an object which is either value or pointer, the method accesses the object based on what the receiver is.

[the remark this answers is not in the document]
zyiou: Got it. Updated

pkg/collector/process.go
zyiou: I'm planning to use this worker/clientHandler errorChan to send stop message to tcp.

pkg/collector/process.go
```diff
@@ -0,0 +1,266 @@
...
+		} else {
+			length = int(field.elementLength)
+		}
+		val := dataBuffer.Next(length)
+		dataMsg.AddInfoElement(field.enterpriseID, field.elementID, val)
+	}
+	return dataMsg, nil
+}
+
+func (cp *collectingProcess) addTemplate(obsDomainID uint32, templateID uint16, elements []*templateField) {
```
srikartati, line 190: Here as well instead of templateField, may be we could use entities.infoElements. Any reason we need new struct?

pkg/collector/process.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+}
+
+func (cp *collectingProcess) deleteTemplate(obsDomainID uint32, templateID uint16) {
+	cp.templatesLock.Lock()
+	defer cp.templatesLock.Unlock()
+	delete(cp.templatesMap[obsDomainID], templateID)
+}
+
+// a common binary decoder to specified interfaces
```
srikartati, line 234: Convention for golang says that comment should start with function name.
https://blog.golang.org/godoc

pkg/collector/process.go
```diff
@@ -0,0 +1,266 @@
...
+	packet := entities.Message{}
+	var id, length uint16
+	err := decode(msgBuffer, &packet.Version, &packet.BufferLength, &packet.ExportTime, &packet.SeqNumber, &packet.ObsDomainID, &id, &length)
+	if err != nil {
+		return 0, fmt.Errorf("Cannot decode message: %v", err)
+	}
+	return int(packet.BufferLength), nil
+}
+
+// get string field length for data record
```
srikartati, line 254: Here as well. Please change the comments.

pkg/collector/process_test.go
```diff
@@ -0,0 +1,291 @@
...
+	assert.NotNil(t, err, "Error should be logged for malformed template record")
+	if _, exist := cp.templatesMap[uint32(1)]; exist {
+		t.Fatal("Template should not be stored for malformed template record")
+	}
+}
+
+func TestCollectingProcess_DecodeDataRecord(t *testing.T) {
+	cp := collectingProcess{}
+	cp.templatesMap = make(map[uint32]map[uint16][]*templateField)
+	cp.templatesLock = &sync.RWMutex{}
```
srikartati, line 243: will this work now?

pkg/collector/udp.go
```diff
@@ -0,0 +1,95 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collector
+
+import (
+	"bytes"
+	"github.com/vmware/go-ipfix/pkg/config"
```
srikartati, line 19: move this to end.

pkg/collector/udp.go
```diff
@@ -0,0 +1,95 @@
...
+		client := cp.createClient()
+		cp.clients[address.String()] = client
+		wg.Add(1)
+		defer wg.Done()
+		go func() {
+			ticker := time.NewTicker(time.Duration(config.TemplateRefreshTimeOut) * time.Second)
+			for {
+				select {
+				case <-client.errChan:
+					klog.Infof("UDP connection from %s has been closed.", address.String())
```
srikartati, line 75: "Collection Process has stopped" is the right error message here.

pkg/collector/process.go
```diff
@@ -0,0 +1,266 @@
...
+	templateTTL uint32
+	// server information
+	address net.Addr
+	// maximum buffer size to read the record
+	maxBufferSize uint16
+	// chanel to receive stop information
+	stopChan chan bool
+	// packet list
+	messages []*entities.Message
+	// channels to handle clients
```
srikartati, line 47: enhance the comment to say map is from address->client

pkg/collector/process.go
```diff
@@ -0,0 +1,266 @@
...
+	clients map[string]*client
+}
+
+type templateField struct {
+	elementID     uint16
+	elementLength uint16
+	enterpriseID  uint32
+}
+
+type client struct {
```
srikartati, line 57: Prefer clientHandler as all fields in the struct is used for handling.

pkg/collector/udp.go
```diff
@@ -0,0 +1,95 @@
...
+					cp.deleteClient(address.Network())
+					return
+				case packet := <-client.packetChan:
+					// get the message here
+					message, err := cp.decodePacket(packet)
+					if err != nil {
+						klog.Error(err)
+						return
+					}
+					klog.Info(message)
```
srikartati, line 88: Increase the debug log level to 2 or 4.  Seeing all messages in the log is not desirable.

pkg/entities/set.go
```diff
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"github.com/vmware/go-ipfix/pkg/config"
```
srikartati, line 21: space here

pkg/entities/message.go
```diff
@@ -20,6 +20,16 @@ const (
...
 )
 
+// data struct of processed message
+type Message struct {
+	Version      uint16
+	BufferLength uint16
+	SeqNumber    uint32
+	ObsDomainID  uint32
+	ExportTime   uint32
+	Record       interface{}
```
srikartati, line 30: Feel Sets is the right name as there can be multiple set inside a message.

srikartati (commented):
Thanks for working on resolving the comments.
The overall approach in handling multiple clients looks good to me. There are some more comments. Please have a look.

Thanks for the unit tests to cover all cases. What do you think of the integration test, where we can use both pkg/exporter and pkg/collector to see they integrate well? We could handle this as a separate PR.

pkg/entities/message.go
zyiou: Yes I'm considering merge MsgBuffer and Message together so that we can have uniform interface for exporter and collector, then Sets can be used here.

pkg/collector/process.go
zyiou: Agreed.

pkg/collector/process_test.go
zyiou: My mistake, should use `sync.RWMutex' here.

pkg/collector/process.go
```diff
@@ -0,0 +1,266 @@
...
+	var fieldCount uint16
+	err := decode(templateBuffer, &templateID, &fieldCount)
+	if err != nil {
+		return nil, fmt.Errorf("Error in decoding message: %v", err)
+	}
+	elements := make([]*templateField, 0)
+	templateMsg := entities.NewTemplateSet()
+
+	for i := 0; i < int(fieldCount); i++ {
+		element := templateField{}
```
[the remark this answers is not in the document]
zyiou: I was using it to separate registry from the collector. Since we are switching to global registry, we can change back to using element directly

pkg/collector/process.go
zyiou: Same as above. We don't need the struct any more. Thanks!

pkg/entities/message.go
srikartati: Sounds good.

zyiou: yes, maybe in the PR to move encoding to io stream or a separate PR.

pkg/collector/process.go
```diff
@@ -130,41 +129,50 @@ func (cp *collectingProcess) decodePacket(packetBuffer *bytes.Buffer) (*entities
...
 		} else {
-			err = decode(templateBuffer, &element.enterpriseID)
+			err = util.Decode(templateBuffer, &enterpriseID)
 			if err != nil {
 				return nil, fmt.Errorf("Error in decoding message: %v", err)
 			}
-			elementID[0] = elementID[0] ^ 0x80
-			element.elementID = binary.BigEndian.Uint16(elementID)
-			element.elementLength = elementLength
+			// encoding format: elementID[0] = elementID[0] | 0x80
```
srikartati, line 163: Something like below is more clear I think with bit numbers on top and providing link https://tools.ietf.org/html/rfc7011#appendix-A.2.2 would be good:
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |1|         Information Element id. = 15              |                    Field Length = 4       |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+++
  |                                                          Enterprise number                                         |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+

srikartati: also moving the comment up top in the else block is more appropriate

pkg/collector/process.go
```diff
@@ -231,36 +242,27 @@ func (cp *collectingProcess) deleteTemplate(obsDomainID uint32, templateID uint1
...
 		return 0, fmt.Errorf("Cannot decode message: %v", err)
 	}
 	return int(packet.BufferLength), nil
 }
 
-// get string field length for data record
+// getFieldLength returns string field length for data record
 func getFieldLength(dataBuffer *bytes.Buffer) int {
 	lengthBuff := dataBuffer.Next(1)
 	var length1 uint8
```
srikartati, line 259: Referencing the link here with comments will be good.
https://tools.ietf.org/html/rfc7011#appendix-A.5

srikartati, line 259: use one variable lengthVal?

pkg/collector/tcp.go
```diff
@@ -59,7 +59,7 @@ func (cp *collectingProcess) handleTCPClient(conn net.Conn, wg *sync.WaitGroup)
 				}
 				break out
 			}
-			klog.Infof("Receiving %d bytes from %s", size, cp.address.String())
+			klog.Infof("Receiving %d bytes from %s", size, address)
```
srikartati, line 62: I guess this is a repeated log. Will log-level 2 be good here?

pkg/collector/udp.go
```diff
@@ -42,12 +43,16 @@ func (cp *collectingProcess) startUDPServer() {
...
-			cp.handleUDPClient(address, &wg)
-			cp.clients[address.String()].packetChan <- bytes.NewBuffer(buff[0:size])
 			if err != nil {
+				if size == 0 { // received stop collector message
+					return
+				}
 				klog.Errorf("Error in collecting process: %v", err)
 				return
 			}
+			klog.Infof("Receiving %d bytes from %s", size, address.String())
```
srikartati, line 53: Here as well. Having a log with level 0, whenever new client is added would be good and rest should be either 2 or 4. What do you think?

pkg/config/config.go
```diff
@@ -15,10 +15,14 @@
 package config
```
srikartati, line 15: do we need a new pkg for this?
Maybe we could move template related info to entities.Set, and registry related const to pkg/registry.

pkg/util/util.go
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+package util
+
+import (
+	"encoding/binary"
+	"io"
+)
+
+// Decode decodes data from io reader to specified interfaces
```
srikartati, line 22: Maybe an example for interfaces like int or string slice with the predefined size is getting decoded here. As this is being used in multiple places, making it clear is better.

pkg/registry/global_registry.go
```diff
@@ -0,0 +1,25 @@
+package registry
```
srikartati, line 1: why do we need a new file? Can this be merged with registry.go as there is only one global map?

pkg/registry/global_registry.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"fmt"
+
+	"github.com/vmware/go-ipfix/pkg/entities"
+)
+
+var globalReg map[uint32]map[uint16]entities.InfoElement
+
+func LoadRegistry() {
```
srikartati, line 11: Create an issue in Antrea to make the change for loading registry accordingly for new version of IPFIX library.

srikartati (commented):
Thanks for addressing earlier comments.
Few more comments. Some of them are nits and others I hope are easy to address. Please take a look.

pkg/collector/process.go
```diff
@@ -130,41 +129,50 @@ func (cp *collectingProcess) decodePacket(packetBuffer *bytes.Buffer) (*entities
...
+			enterpriseID = uint32(0)
+			element, err = registry.GetInfoElementFromID(elementID, enterpriseID)
+			if err != nil {
+				return nil, err
+			}
 		} else {
-			err = decode(templateBuffer, &element.enterpriseID)
+			err = util.Decode(templateBuffer, &enterpriseID)
 			if err != nil {
 				return nil, fmt.Errorf("Error in decoding message: %v", err)
```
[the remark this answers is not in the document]
zyiou: The error is returned when decoding has problems. Do you mean that we should return the plain error without saying `Error in ...`?

pkg/collector/process.go
```diff
@@ -44,31 +45,25 @@ type collectingProcess struct {
...
-type client struct {
+type clientHandler struct {
 	packetChan chan *bytes.Buffer
 	errChan    chan bool
 }
 
 func InitCollectingProcess(address net.Addr, maxBufferSize uint16, templateTTL uint32) (*collectingProcess, error) {
 	collectProc := &collectingProcess{
-		templatesMap:  make(map[uint32]map[uint16][]*templateField),
+		templatesMap:  make(map[uint32]map[uint16][]entities.InfoElement),
```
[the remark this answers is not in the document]
zyiou: Makes sense. Changed back to pointer.

pkg/collector/tcp.go
zyiou: Agreed.

pkg/collector/process.go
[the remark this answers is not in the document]
srikartati: I meant why not `return nil, err` like the if block above.

pkg/collector/udp.go
zyiou: Yes I agreed. Log level 2 is better than 0 here

pkg/registry/global_registry.go
zyiou: you're right. I was thinking that it is just a temporary file that further changes to global registry can be made in another PR. Merging it with registry.go is better here. Thanks!

pkg/config/config.go
zyiou: Make sense. Thanks!

pkg/collector/process.go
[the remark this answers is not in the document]
zyiou: Since it is a decoding error, I didn't specify the error type in decoding function. Thanks for pointing this out. I will need to change the error log in decode function and just return `nil, err` here.

pkg/registry/global_registry.go
zyiou: Do you refer to the global registry refactoring?

pkg/collector/process.go
```diff
@@ -145,22 +144,32 @@ func (cp *collectingProcess) decodeTemplateSet(templateBuffer *bytes.Buffer, obs
...
+			enterpriseID = registry.IANAEnterpriseID
 			element, err = registry.GetInfoElementFromID(elementID, enterpriseID)
 			if err != nil {
 				return nil, err
 			}
 		} else {
+			/*
+					Encoding format for Enterprise-Specific Information Elements:
+				     0                   1                   2                   3
+					 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
```
srikartati, line 161: Format looks little off. You could view the file from github options "View File" at top right corner in the window.

pkg/registry/config.go
```diff
@@ -12,17 +9,11 @@
...
 
 const (
-	// TemplateRefreshTimeOut is the template refresh time out for exporting process
-	TemplateRefreshTimeOut uint32 = 1800
-	// TemplateTTL is the template time to live for collecting process
-	TemplateTTL uint32 = TemplateRefreshTimeOut * 3
-	// TemplateSetID is the setID for template record
-	TemplateSetID uint16 = 2
 	// AntreaEnterpriseID is the enterprise ID for Antrea Information Elements
 	AntreaEnterpriseID uint32 = 55829
```
srikartati, line 16: You could define these constants pkg/registry/registry.go. Do not need a separate file.

srikartati (commented):
Couple of nits.
Yiou, Wondering if you were able to test this collector code manually with Antrea flow exporter? If not, better to test once before checking in.

pkg/registry/global_registry.go
srikartati: Yes global registry change. We need to make corresponding Antrea change if we move to next version of go-ipfix library.

srikartati:
> Couple of nits.
> Yiou, Wondering if you were able to test this collector code manually with Antrea flow exporter? If not, better to test once before checking in.

I meant manual test as in a simple main function starting the collector process and printing stream of flow records coming from the flow exporter.

pkg/collector/process.go
```diff
@@ -156,15 +156,15 @@ func (cp *collectingProcess) decodeTemplateSet(templateBuffer *bytes.Buffer, obs
...
-					|1| Information Element id. = 15 | Field Length = 4 |
-					+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+++
-					| Enterprise number |
-					+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-					(Reference: https://tools.ietf.org/html/rfc7011#appendix-A.2.2)
+				Encoding format for Enterprise-Specific Information Elements:
+				 0                   1                   2                   3
+				 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+				+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+				|1| Information Element id. = 15 | Field Length = 4 |
```
srikartati, line 163: Sorry for nit picking. The number of bits/bytes used is still unclear. Instead of values, we can add bits in parentheses 
1 (1bit) | Information Element Id(15bits) Field Length (16bits)
Enterprised ID(32 bits)

zyiou: Due to space concerns, for first bit I didn't add the `(1bit)` to the diagram, but under it. And changed the element length for field length and enterprise id (previously I didn't copy the diagram correctly). Thanks!

pkg/collector/process.go
```diff
@@ -266,12 +270,13 @@ func getMessageLength(msgBuffer *bytes.Buffer) (int, error) {
 // (encoding reference: https://tools.ietf.org/html/rfc7011#appendix-A.5)
 func getFieldLength(dataBuffer *bytes.Buffer) int {
 	lengthBuff := dataBuffer.Next(1)
-	var lengthVal uint16
-	util.Decode(bytes.NewBuffer(lengthBuff), &lengthVal)
-	if lengthVal < 255 { // string length is less than 255
-		return int(lengthVal)
+	var lengthVal1 uint8
+	util.Decode(bytes.NewBuffer(lengthBuff), &lengthVal1)
+	if lengthVal1 < 255 { // string length is less than 255
```
srikartati, line 275: was there an issue using one variable?

pkg/registry/registry.go
```diff
@@ -44,6 +42,8 @@ const (
 	AntreaEnterpriseID uint32 = 55829
 	// IANAEnterpriseID is the enterprise ID for IANA Information Elements
 	IANAEnterpriseID uint32 = 0
+	// Enterprise ID for reverse Information Elements
```
srikartati, line 45: nit: move const and global variable to up top
Usually structs are defined after constants and global variables.

pkg/collector/process.go
srikartati: I see the difference. lengthOneByte and lengthTwoBytes are better names. What do you think?

srikartati (approved):
Please take care of the above nits before merging. Otherwise, LGTM.

pkg/collector/process.go
zyiou: Yes. Makes sense to me. Thanks!

PR #21, opened 2020-09-11
Add integration tests for exporter and collector

zyiou:


pkg/test/integration_test.go
```diff
@@ -0,0 +1,153 @@
+// Copyright 2020 VMware, Inc.
```
srikartati, line 1: Filename could be different. integration_test seems very generic. exporter_collector_test?

zyiou: Makes sense. Thanks

srikartati (commented):
LGTM. As this is dependent on other collector PR. Lets merge that one first and this can be merged later.

pkg/test/exporter_collector_test.go
```diff
@@ -0,0 +1,166 @@
...
+	assert.Equal(t, []uint16{101}, templateSet[registry.AntreaEnterpriseID], "TemplateSet does not store template elements (Antrea) correctly.")
+
+	dataSet, ok := dataMsg.Set.(entities.DataSet)
+	if !ok {
+		t.Error("Data packet is not decoded correctly.")
+	}
+	assert.Equal(t, []byte{1,2,3,4}, dataSet[registry.IANAEnterpriseID][8], "DataSet does not store elements (IANA) correctly.")
+	assert.Equal(t, uint64(12345678), dataSet[registry.ReverseEnterpriseID][1], "DataSet does not store reverse information elements (IANA) correctly.")
+	assert.Equal(t, "pod1", dataSet[registry.AntreaEnterpriseID][101], "DataSet does not store elements (Antrea) correctly.")
+}
```
srikartati, line 166: can you add a new line?

pkg/test/exporter_collector_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/stretchr/testify/assert"
+	"k8s.io/klog/v2"
+
+	"github.com/vmware/go-ipfix/pkg/collector"
+	"github.com/vmware/go-ipfix/pkg/entities"
+	"github.com/vmware/go-ipfix/pkg/exporter"
+	"github.com/vmware/go-ipfix/pkg/registry"
+)
+
+type Addr struct {
```
srikartati, line 31: why not directly use [net.Addr](https://golang.org/pkg/net/#Addr)?

zyiou: Because net.Addr is an interface, I just implement that here.

srikartati: ResolveUDPAddr and ResolveTCPAddr can help here. Those functions take a string as input and returns net.Addr interface.

zyiou: Got it. Updated. Thanks!

srikartati (approved):

PR #24, opened 2020-09-17
Change to global registry

### LINKED ISSUE #19: Support global registry
Refactor and improve current registry to support global registries for users. Once initialized, registry can be used in both exporter and collector.

zyiou:
Changed registry loading and removed registry interface so that users don't need to load IANA and Antrea registry separately.

Fixes #19


pkg/registry/registry.go
```diff
@@ -30,123 +30,116 @@ const (
...
-	if ie, exist = reg.registry[name]; !exist {
+	if ie, exist = antreaRegistry[name]; !exist {
 		err := fmt.Errorf("Antrea Registry: There is no information element with name %s", name)
 		return &ie, err
 	}
 	return &ie, nil
 }
 
-func (reg *antreaRegistry) GetReverseInfoElement(name string) (*entities.InfoElement, error) {
+func GetAntreaReverseIE(name string) (*entities.InfoElement, error) {
```
srikartati, line 106: This is not correct--I think we missed this before. We cannot use ReverseEnterpriseID for Antrea registry, that is only applicable for IANA registry. 
https://tools.ietf.org/html/rfc5103#section-6.2

We should define reverse elements as part of Antrea registry itself. I don't think we have any use case for Antrea reverse elements. Let me know if you think otherwise.

srikartati (commented):
Main comment is to reduce the functions by using enterpriseID and define two global registries one by ID and other one by name.

pkg/registry/registry.go
zyiou: Got it. I don't think we are using Antrea reverse elements. Thanks!

pkg/registry/build_registry/build_registry.go
```diff
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"encoding/csv"
 	"fmt"
+	"github.com/vmware/go-ipfix/pkg/registry"
```
srikartati, line 23: could you move this to proper position?

pkg/registry/registry.go
```diff
@@ -30,119 +30,74 @@ const (
...
+	globalRegistryByName map[uint32]map[string]entities.InfoElement
+)
 
 func LoadRegistry() {
-	globalReg = make(map[uint32]map[uint16]entities.InfoElement)
-	antreaReg := NewAntreaRegistry()
-	antreaReg.LoadRegistry()
-	ianaReg := NewIanaRegistry()
-	ianaReg.LoadRegistry()
+	globalRegistryByID = make(map[uint32]map[uint16]entities.InfoElement)
```
srikartati, line 41: pointer to entities.InfoElement should be good enough

pkg/registry/registry.go
```diff
@@ -30,119 +30,74 @@ const (
...
-}
-
-func NewAntreaRegistry() *antreaRegistry {
-	reg := make(map[string]entities.InfoElement)
-	globalReg[AntreaEnterpriseID] = make(map[uint16]entities.InfoElement)
-	return &antreaRegistry{
-		registry: reg,
+func GetInfoElement(name string, enterpriseID uint32) (*entities.InfoElement, error) {
+	if element, exist := globalRegistryByName[enterpriseID][name]; !exist {
+		return &element, fmt.Errorf("Information Element with name %s and enterpriseID %d cannot be found.", name, enterpriseID)
```
srikartati, line 65: -> with name %s in registry with enterpriseID %d

pkg/registry/registry.go
```diff
@@ -30,119 +30,74 @@ const (
...
-	var ie entities.InfoElement
-	if ie, exist = reg.registry[name]; !exist {
-		err := fmt.Errorf("IANA Registry: There is no information element with name %s", name)
-		return &ie, err
-	}
-	return &ie, nil
-}
-
-func (reg *ianaRegistry) GetReverseInfoElement(name string) (*entities.InfoElement, error) {
+func getIANAReverseIE(name string) (*entities.InfoElement, error) {
```
srikartati, line 90: getIANAReverseInfoElement just to be consistent with other function names.

srikartati (commented):
Since this change touches all unit tests and integration tests, please make sure all are passing.

srikartati (approved):
Please make changes in Antrea repo accordingly.

PR #33, opened 2020-09-23
Revert klog from v2 to v1

zyiou:
to keep klog version consistent with Antrea


srikartati (approved):
Please open the closed issue and add a comment that we will move to klog v2 when Antrea moves to klog v2.

PR #39, opened 2020-09-25
Github CI: Create go.yml

srikartati:
Added CI job through github actions. It checks if unit tests pass or not.


zyiou (approved):
I have no experience in CI, so I only checked the commands to run and the yaml format. LGTM

srikartati:
> I have no experience in CI, so I only checked the commands to run and the yaml format. LGTM

Thanks for the review. This is my first time too.. I am also learning :)

PR #46, opened 2020-09-25
Add RWMutex lock for clients map to prevent concurrent writes

### LINKED ISSUE #42: Collector concurrent map writes
Unit tests failed here https://github.com/vmware/go-ipfix/runs/1163485465 due to concurrent map writes on client handlers.

zyiou:
fixes #42

### DIFF
--- a/pkg/collector/process.go
+++ b/pkg/collector/process.go
@@ -46,6 +46,8 @@ type collectingProcess struct {
 	messages []*entities.Message
 	// maps each client to its client handler (required channels)
 	clients map[string]*clientHandler
+	// clientsLock allows multiple readers or one writer to access clients map at the same time
+	clientsLock sync.RWMutex
 }
 
 type clientHandler struct {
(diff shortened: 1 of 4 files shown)


pkg/collector/process.go
```diff
@@ -46,6 +46,8 @@ type collectingProcess struct {
 	messages []*entities.Message
 	// maps each client to its client handler (required channels)
 	clients map[string]*clientHandler
+	// clientsLock allows multiple readers or one writer to clients map at the same time
```
srikartati, line 49: to access  clients map is missing

srikartati (dismissed):
LGTM except for nit

pkg/collector/process.go
```diff
@@ -46,6 +46,8 @@ type collectingProcess struct {
 	messages []*entities.Message
 	// maps each client to its client handler (required channels)
```
srikartati, line 47: Not related to this PR. I feel it is better to have uniformity/convention when adding comments to the fields, i.e., starting the comment with the field name. what do you say?

srikartati (commented):
A nit about comments. It's better to follow a convention.

pkg/collector/process.go
zyiou: Makes sense. I will create a PR to uniform all the comments. Thanks!

srikartati (approved):

PR #49, opened 2020-10-05
Refactor and extend Set and Record structure

### LINKED ISSUE #32: Add multiple data records to data set when encoding IPFIX message
This will be a new method in exporting process. 
Currently, only one data record per data set is supported.

Please make sure following issue is also solved:
When adding multiple data records in IPFIX message, change the sequence number accordingly.
https://github.com/vmware/go-ipfix/projects/1#card-41458115

zyiou:
Fixes #32 

This PR involves merging data records into data set, support of multiple data records per data set, refactor set and record interfaces and corresponding changes in testing.


pkg/entities/record.go
```diff
@@ -33,12 +35,13 @@ import (
...
 type Record interface {
 	PrepareRecord() (uint16, error)
-	AddInfoElement(element *InfoElement, val interface{}) (uint16, error)
+	AddInfoElement(element *InfoElement, val interface{}, isDecoding bool) (uint16, error)
 	// TODO: Functions for multiple elements as well.
 	GetBuffer() *bytes.Buffer
 	GetTemplateID() uint16
 	GetFieldCount() uint16
 	GetTemplateElements() []*InfoElement
+	GetDataElements() []*InfoElementValue
```
srikartati, line 44: If value is added to InfoElement, we can have one interface GetInfoElements that works for both data record and template record?

pkg/entities/record.go
```diff
@@ -33,12 +35,13 @@ import (
...
 	GetMinDataRecordLen() uint16
 }
```
srikartati, line 47: This below TODO can be removed as baseRecord is already defined.

pkg/entities/record.go
```diff
@@ -99,200 +104,320 @@ func (d *dataRecord) PrepareRecord() (uint16, error) {
...
-			ipv4Int.SetBytes(ipv4Add)
-			binary.BigEndian.PutUint32(bytesToAppend, uint32(ipv4Int.Uint64()))
-		} else {
-			for i, b := range v {
-				bytesToAppend[i] = b
+	if isDecoding {
+		ieValue := NewInfoElementValue(element, value)
+		d.elements = append(d.elements, ieValue)
+	} else {
+		if element.DataType == String {
```
srikartati, line 118: This dataType seems to be handled in convertToType already. Why is there a check again?

pkg/entities/record.go
```diff
@@ -99,200 +104,320 @@ func (d *dataRecord) PrepareRecord() (uint16, error) {
...
-	return t.templateElements
+	return t.elements
 }
 
 func (t *templateRecord) GetMinDataRecordLen() uint16 {
 	return t.minDataRecLength
 }
+
+// convertToType is to decode or encode data to specific type
+func convertToType(dataType IEDataType, val interface{}, isDecoding bool) (interface{}, error) {
```
srikartati, line 192: convertToIEDataType is more appropriate name?

pkg/entities/record.go
```diff
@@ -99,200 +104,320 @@ func (d *dataRecord) PrepareRecord() (uint16, error) {
...
+	if isDecoding {
```
srikartati, line 193: can this if else block be removed if we use a variable for function pointer that is initialized to either util.Decode or util.Encode?

pkg/exporter/process.go
```diff
@@ -101,16 +101,22 @@ func InitExportingProcess(collectorAddr net.Addr, obsID uint32, tempRefTimeout u
...
 }
 
-func (ep *ExportingProcess) AddRecordAndSendMsg(recType entities.ContentType, rec entities.Record) (int, error) {
-	if recType == entities.Template {
-		ep.updateTemplate(rec.GetTemplateID(), rec.GetTemplateElements(), rec.GetMinDataRecordLen())
-	} else if recType == entities.Data {
-		err := ep.dataRecSanityCheck(rec)
-		if err != nil {
-			return 0, fmt.Errorf("AddRecordAndSendMsg: error when doing sanity check:%v", err)
+func (ep *ExportingProcess) AddRecordAndSendMsg(recType entities.ContentType, records ...entities.Record) (int, error) {
```
srikartati, line 104: AddRecordsAndSendMsg. Or this API can be modified a bit to make it AddSetsAndSendMsg as there is object for Set now. I prefer the latter option. It is better to follow the abstraction: msg->sets->records. What do you think?

srikartati (commented):
Thanks for the PR. I feel overall the change looks like to be achieving the intended abstraction/design except at few places. Let me know what you think.
Title of PR could be modified as this PR is adding more functionalities.

pkg/entities/set.go
```diff
@@ -123,114 +126,34 @@ func (s *Set) FinishSet() {
...
-		d[element.EnterpriseId][element.ElementId] = val.Bytes()
-	case String:
-		d[element.EnterpriseId][element.ElementId] = val.String()
-	default:
-		return fmt.Errorf("API supports only valid information elements with datatypes given in RFC7011")
+	d.records = append(d.records, record)
+}
+
+func (d *DataSet) GetRecords() []Record {
+	var recs []Record
```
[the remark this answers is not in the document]
zyiou: Reason being that []Record is a slice type, but not an interface type, so []*dataRecord is not assignable to it. A similar example here: https://stackoverflow.com/questions/43366452/slice-of-a-specific-interface-in-go

pkg/entities/message.go
```diff
@@ -33,21 +33,31 @@ type Message struct {
 // Does it need an interface?
 
 type MsgBuffer struct {
-	buffer      bytes.Buffer
-	dataRecFlag bool
+	buffer       bytes.Buffer
+	dataRecFlag  bool
+	numOfRecords uint32
```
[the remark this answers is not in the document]
zyiou: That's correct. I use numOfRecords here to make it easier to calculate sequence number when sending message. If we use numOfSets and numOfRecords, we will need a way to get Sets from message and then get number of records, which is a little complicated since the only use here is to calculate sequence number. What do you think?

pkg/entities/ie.go
```diff
@@ -87,6 +87,12 @@ type InfoElement struct {
 	// Add description and dataType semantics if required
 }
 
+// InfoElementValue represents mapping from element to value for data records
+type InfoElementValue struct {
+	Element *InfoElement
+	Value   interface{}
```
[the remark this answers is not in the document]
zyiou: Makes sense. we will need to keep `Value` field empty except data records then.

pkg/entities/record.go
zyiou: Sounds good. Will do that. thanks!

pkg/entities/record.go
zyiou: The only concern is that for encoding, firstly we need to check whether the provided value is the format required data type. For decoding, we don't have to do the checking, that makes the code different for handling encoding/decoding.

zyiou: Also for decoding, an empty interface of corresponding type for storing the result needs to be passed to function.

pkg/exporter/process.go
zyiou: Agreed. Will do the corresponding changes. Thanks!

pkg/entities/record.go
srikartati: We can probably move the type check with switch functionality for encoding to a different function. 
Or if refactoring is difficult, maybe we should define separate functions for encoding and decoding to cover the functionality of convertToIEDataType. Functionality seems to be different as well.. for encoding we are writing data to msg buffer and for decoding we are writing to info elements.

pkg/entities/set.go
```diff
@@ -43,28 +41,36 @@ const (
 // Do not expose set to IPFIX library user
 // Not creating any interface. Plan to use same struct for Template and Data sets
```
srikartati, line 42: Please remove the comments if not applicable.

pkg/entities/set.go
```diff
@@ -43,28 +41,36 @@ const (
...
 
-// enterpriseID -> elementID -> val
-type DataSet map[uint32]map[uint16]interface{}
+type DataSet struct {
+	records []*dataRecord
+}
 
-func NewSet(buffer *bytes.Buffer) *Set {
-	return &Set{
+func NewSet(buffer *bytes.Buffer) *BaseSet {
```
srikartati, line 65: is this used anywhere?
Set interface and other structures BaseSet, TemplateSet, DataSet seems confusing. Which structures are implementing Set interface? I do not see BaseSet embedded in TemplateSet or DataSet?

pkg/exporter/process.go
```diff
@@ -44,7 +44,7 @@ type ExportingProcess struct {
 	obsDomainID     uint32
 	seqNumber       uint32
 	templateID      uint16
-	set             *entities.Set
+	set             *entities.BaseSet
```
srikartati, line 47: This should be interface, so that it can be applied for both DataSet and TemplateSet. In addition, BaseSet should be embedded in DataSet and TemplateSet--they both should implement the inerface Set. Do you think that works?

pkg/test/exporter_collector_test.go
```diff
@@ -33,18 +33,34 @@ func TestUDPTransport(t *testing.T) {
 	if err != nil {
 		t.Error(err)
 	}
-	testExporterToCollector(address, t)
+	testExporterToCollector(address, false, t)
 }
 
 func TestTCPTransport(t *testing.T) {
```
srikartati, line 39: Change the name of the test to TestSingleRecord*?

pkg/entities/set.go
```diff
@@ -43,28 +41,36 @@ const (
 // Do not expose set to IPFIX library user
 // Not creating any interface. Plan to use same struct for Template and Data sets
 
-type Set struct {
+type Set interface {
```
srikartati, line 44: Does this interface need mock for unit testing of collector and exporter?

srikartati (commented):
Thanks for addressing earlier comments. 
Follow up comments and new comments.

pkg/entities/ie.go
```diff
@@ -94,6 +96,18 @@ func NewInfoElement(name string, ieID uint16, ieType IEDataType, entID uint32, l
 		DataType:     ieType,
 		EnterpriseId: entID,
 		Len:          len,
+		Value:        nil,
+	}
+}
+
+func NewInfoElementWithValue(ie *InfoElement, val interface{}) *InfoElement {
+	return &InfoElement{
```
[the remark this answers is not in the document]
zyiou: Yes. will change it. Thanks!

pkg/entities/record.go
```diff
@@ -35,21 +35,21 @@ import (
...
-	GetTemplateElements() []*InfoElement
+	GetInfoElements() []*InfoElement
 	GetMinDataRecordLen() uint16
 }
 
-// TODO: Create base record struct. Some functions like GetBuffer will be applicable to base record.
 type baseRecord struct {
 	buff       bytes.Buffer
 	len        uint16
 	fieldCount uint16
```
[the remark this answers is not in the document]
zyiou: `fieldCount` can work as initial declaration of number of fields in a template like [this.](https://github.com/vmware/go-ipfix/blob/82c8b12cc160655070e9d48caa7a33d0cf79fa28/pkg/entities/record.go#L125) We can also use it to check whether number of added InfoElements is the same as declared value. So I suggest we keep this value here.

pkg/entities/set.go
zyiou: Yes, will add mock here.

pkg/exporter/process.go
zyiou: Agreed that it should be interface and I will need to refactor set structure. I kept the original idea from you to use one set struct to templateSet and dataSet to make things more tidy and reusable. Thanks!

pkg/entities/record.go
```diff
@@ -101,89 +106,281 @@ func (d *dataRecord) PrepareRecord() (uint16, error) {
...
+func (t *templateRecord) GetInfoElements() []*InfoElementValue {
+	return t.elements
+}
+
+func (t *templateRecord) GetMinDataRecordLen() uint16 {
+	return t.minDataRecLength
+}
+
+// decodeToIEDataType is to decode to specific type
+func (d *dataRecord) decodeToIEDataType(dataType IEDataType, val interface{}) (interface{}, error) {
```
srikartati, line 183: Does it need to be a method of dataRecord? You could define it as utility function.

pkg/entities/record.go
```diff
@@ -101,89 +106,281 @@ func (d *dataRecord) PrepareRecord() (uint16, error) {
...
+			default:
+				return nil, fmt.Errorf("API supports only valid information elements with datatypes given in RFC7011")
+			}
+		}
+	}
+	return nil, fmt.Errorf("API supports only valid information elements with datatypes given in RFC7011")
+}
+
+// encodeToIEDataType is to encode to specific type to the buff
+func (d *dataRecord) encodeToIEDataType(dataType IEDataType, val interface{}) (interface{}, error) {
```
srikartati, line 291: Same here; it can be utility function... d.buff can be sent as argument. Seems more like a generic function rather than being part of dataRecord structure.

pkg/entities/record.go
```diff
@@ -101,89 +106,281 @@ func (d *dataRecord) PrepareRecord() (uint16, error) {
...
+				}
+				return v, nil
+			case DateTimeMicroseconds, DateTimeNanoseconds:
+				return nil, fmt.Errorf("This API does not support micro and nano seconds types yet")
+			case MacAddress, Ipv4Address, Ipv6Address:
+				return value.Bytes(), nil
+			case String:
+				return value.String(), nil
+			default:
+				return nil, fmt.Errorf("API supports only valid information elements with datatypes given in RFC7011")
```
srikartati, line 283: do we need this as L287 does the same return?

pkg/entities/record.go
```diff
@@ -101,89 +106,281 @@ func (d *dataRecord) PrepareRecord() (uint16, error) {
...
+}
+
+func (t *templateRecord) GetMinDataRecordLen() uint16 {
+	return t.minDataRecLength
+}
+
+// decodeToIEDataType is to decode to specific type
+func (d *dataRecord) decodeToIEDataType(dataType IEDataType, val interface{}) (interface{}, error) {
+	switch value := val.(type) {
+	case *bytes.Buffer:
```
srikartati, line 185: what is the purpose of this switch-case? Simple if-check is good enough I think.

pkg/exporter/process.go
```diff
@@ -101,23 +102,28 @@ func InitExportingProcess(collectorAddr net.Addr, obsID uint32, tempRefTimeout u
...
-			return 0, fmt.Errorf("AddRecordAndSendMsg: error when doing sanity check:%v", err)
+func (ep *ExportingProcess) AddSetAndSendMsg(setType entities.ContentType, set entities.Set) (int, error) {
+	recBytes := make([]byte, 0)
+	for _, record := range set.GetRecords() {
+		if setType == entities.Template {
+			ep.updateTemplate(record.GetTemplateID(), record.GetInfoElements(), record.GetMinDataRecordLen())
+		} else if setType == entities.Data {
+			err := ep.dataRecSanityCheck(record)
+			if err != nil {
+				return 0, fmt.Errorf("AddRecordAndSendMsg: error when doing sanity check:%v", err)
```
srikartati, line 113: AddSetAndSendMsg in error.

pkg/entities/set.go
```diff
@@ -40,31 +40,35 @@ const (
...
 
-// Do not expose set to IPFIX library user
-// Not creating any interface. Plan to use same struct for Template and Data sets
+type Set interface {
+	CreateNewSet(setType ContentType, templateID uint16) error
+	GetBuffLen() uint16
+	GetSetType() ContentType
+	WriteRecordToSet(recBuffer *[]byte) error
+	FinishSet()
+	AddRecord(elements []*InfoElementValue, templateID uint16, isDecoding bool)
```
srikartati, line 49: Can we merge both WriteRecordToSet and AddRecord methods? If isDecoding set, we could skip writing to the buffer as it is not needed, otherwise we should write to buffer.

srikartati (commented):
Thanks @zyiou for addressing earlier comments. Posted some more comments.
Glad how the code is shaping now. We are making Msg->Set->Record->InfoElements abstractness more robust.

pkg/entities/set.go
```diff
@@ -38,39 +40,35 @@ const (
 	Undefined = 255
 )
 
-// Do not expose set to IPFIX library user
-// Not creating any interface. Plan to use same struct for Template and Data sets
-
 type Set interface {
-	AddRecord(elements []*InfoElement, templateID uint16, isDecoding bool)
+	CreateNewSet(setType ContentType, templateID uint16) error
```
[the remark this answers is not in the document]
zyiou: If combining `CreateNewSet` and `NewSet`, we will need to pass setType and templateID into constructor. For `set` in `ExportingProcess` struct, the set is initialized without templateID or setType. Maybe this method is more like `initNewSet`?

pkg/entities/record.go
zyiou: Makes sense. I moved the functions to `ie.go` since they are info element related. Thanks!

pkg/entities/record.go
zyiou: Removed. Thanks!

pkg/entities/record.go
zyiou: I was trying to convert interface to bytes.Buffer. Just realized that if-check can work too. Thanks!

pkg/entities/set.go
zyiou: `WriteRecordToSet` is used by exporting process to write records (bytes) to `ep.set`, `AddRecord` is used by users to add `InfoElementWithValue`s as record to set. If combining them, for every set passed into `AddSetAndSendMsg`, we need to copy and add all records to `ep.set`. I would suggest to keep `WriteRecordToSet` and add code in `AddRecord` to write record to set. It will be convenient when sending message. What do you think?

srikartati: As exporting process calls `AddRecord` with false `isDecoding` flag, could skip adding elements to record based on set type and the 'isDecoding' flag? There are 4 scenarios with `setType` and `isDecoding` flag. All these scenarios could be handled in one interface. What about this approach?

pkg/exporter/process.go
srikartati: L148 also has AddRecordAndSendMsg.

pkg/entities/set.go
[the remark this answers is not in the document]
srikartati: Do we need set in ExportingProcess now, as we are sending entities.Set to AddSetAndSendMsg? ExportingProcess should have message buffer. We keep on adding sets to the msg buffer and send the msg. As `AddSetAndSendMsg` already takes in a set, we can use that directly. We do not have copy that into the set field of exportingProcess struct.  I do not see a requirement to have Set interface field in ExportingProcess. What do you think? Is removing the field a complex change?

Every set has to be associated with templateID and setType. We can definitely change the name of the Constructor as InitializeNewSet but prefer simpler name and handle the initialization internally.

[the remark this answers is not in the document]
zyiou: Sounds good. Thanks for the suggestion. I will remove set from exporting process and keep only one constructor.

pkg/entities/set.go
```diff
@@ -59,16 +58,7 @@ type set struct {
...
-		buffer:  buffer,
-		currLen: 0,
-		setType: Undefined,
-		records: make([]Record, 0),
-	}
-}
-
-func (s *set) CreateNewSet(setType ContentType, templateID uint16) error {
+func NewSet(setType ContentType, templateID uint16) Set {
 	// Create the set header and append it
```
srikartati, line 62: Do we need set header for collector/decode messages too?

pkg/exporter/process.go
```diff
@@ -120,39 +116,24 @@ func (ep *ExportingProcess) AddSetAndSendMsg(setType entities.ContentType, set e
...
 	msgBuffer := ep.msg.GetMsgBuffer()
 	var bytesSent int
-	// Check if message is exceeding the limit with new record
-	if uint16(msgBuffer.Len()+ int(ep.set.GetBuffLen())) > entities.MaxTcpSocketMsgSize {
-		ep.set.FinishSet()
-		b, err := ep.sendMsg(set.GetNumberOfRecords())
-		if err != nil {
-			return b, err
-		}
-		bytesSent = bytesSent + b
```
srikartati: We still need to send the message, right? FinishSet would have been called with the previous set.

pkg/exporter/process.go
```diff
@@ -101,55 +98,44 @@ func InitExportingProcess(collectorAddr net.Addr, obsID uint32, tempRefTimeout u
...
-	if err != nil {
-		return bytesSent, fmt.Errorf("AddRecordAndSendMsg: %v", err)
-	}
-	if recType == entities.Data && !ep.msg.GetDataRecFlag() {
+
+	if setType == entities.Data && !ep.msg.GetDataRecFlag() {
 		ep.msg.SetDataRecFlag(true)
 	}
 
 	// Send the message right after attaching the record
```
srikartati, line 134: Remove the comment. May be adding a comment why finish set is needed is better here.

pkg/entities/set.go
zyiou: we don't need set header or set buffer when decoding message. I will add a `isDecoding` field in set to handle that then. Thanks!

pkg/exporter/process.go
zyiou: Since we don't keep a `set` field in `exportingProcess` struct and currently don't support multiple set per message, there will be no previous set. For every set passed in the function, if the size is less than `MaxTcpSocketMsgSize`, we will send it because we cannot separate current set into two different sets now. Do you think we should support splitting set or multiple sets per message here?

srikartati: Got it. 
I think split sets case might be important because we do not add any gating in set.AddRecords and add any number of records. I feel it is better to add the gating there and define in the interface that the length of the records should not exceed this `MaxTcpSocketMsgSize`. What do you think?

Let us handle this in a different PRs. Please create new issue for split sets scenario with probable gating in sets. A TODO or an issue for multiple sets would be good.

srikartati (approved):
LGTM. Thanks for making the changes.
Let's handle the split sets and multiple sets cases separately.

pkg/exporter/process.go
zyiou: Sounds good. Will create issue then. Thanks for reviewing.

PR #51, opened 2020-10-19
Implementation of Intermediate Process

### LINKED ISSUE #3: Implement IPFIX mediator as per RFC7119

zyiou:
Fixes #3 

This PR is based on PR #49 refactor of set and record struct (will need rebase after #49 merged). Changes made:
1. Modify message list in collector to channel as buffer to support multi producers -> multiple consumers
2. Support adding exporter related info `originalObservationDomainId` and `originalExporterIPv4Address` in collecting process
3. Add aggregation related methods to aggregate messages by 5-tuple and correlate records to fill fields.
4. Support multiple workers to process message.
5. Add unit tests and integration test from collector to intermediate process.

### DIFF
--- a/cmd/collector/collector.go
+++ b/cmd/collector/collector.go
@@ -77,7 +77,7 @@ func printIPFIXMessage(msg *entities.Message) {
 		fmt.Fprint(&buf, "TEMPLATE SET:\n")
 		for i, record := range msg.Set.GetRecords() {
 			fmt.Fprintf(&buf, "  TEMPLATE RECORD-%d:\n", i)
-			for _, ie := range record.GetInfoElements() {
+			for _, ie := range record.GetOrderedElementList() {
 				fmt.Fprintf(&buf, "    %s: len=%d (enterprise ID = %d) \n", ie.Element.Name, ie.Element.Len, ie.Element.EnterpriseId)
 			}
 		}
@@ -85,7 +85,7 @@ func printIPFIXMessage(msg *entities.Message) {
 		fmt.Fprint(&buf, "DATA SET:\n")
 		for i, record := range msg.Set.GetRecords() {
 			fmt.Fprintf(&buf, "  DATA RECORD-%d:\n", i)
-			for _, ie := range record.GetInfoElements() {
+			for _, ie := range record.GetOrderedElementList() {
 				fmt.Fprintf(&buf, "    %s: %v \n", ie.Element.Name, ie.Value)
 			}
 		}
@@ -130,7 +130,7 @@ func run() error {
 		return fmt.Errorf("input given ipfix.transport flag is not supported or valid")
 	}
 	// Initialize collecting process
-	cp, err := collector.InitCollectingProcess(netAddr, 65535, 0, false)
+	cp, err := collector.InitCollectingProcess(netAddr, 65535, 0)
 	if err != nil {
 		return err
 	}
--- a/cmd/collector/collector.go
+++ b/cmd/collector/collector.go
@@ -130,7 +130,7 @@ func run() error {
 		return fmt.Errorf("input given ipfix.transport flag is not supported or valid")
 	}
 	// Initialize collecting process
-	cp, err := collector.InitCollectingProcess(netAddr, 65535, 0)
+	cp, err := collector.InitCollectingProcess(netAddr, 65535, 0, false)
 	if err != nil {
 		return err
 	}
@@ -138,20 +138,10 @@ func run() error {
 	messageReceived := make(chan *entities.Message)
 	go func() {
 		go cp.Start()
-		msgCount := 0
-		// This can be simplified if collector uses a message channel.
-		// Are there any use cases for having list of messages in collector process.
-		for {
-			ipfixMessages := cp.GetMessages()
-			currentLen := len(ipfixMessages)
-			if currentLen > msgCount {
-				for i := msgCount; i < currentLen; i++ {
-					klog.Info("Processing IPFIX message")
-					messageReceived <- ipfixMessages[i]
-					msgCount++
-				}
-			}
-			time.Sleep(5 * time.Millisecond)
+		msgChan := cp.GetMsgChan()
+		for message := range msgChan {
+			klog.Info("Processing IPFIX message")
+			messageReceived <- message
 		}
 	}()
 
--- a/pkg/collector/process.go
+++ b/pkg/collector/process.go
@@ -47,8 +47,6 @@ type collectingProcess struct {
 	useMsgChan bool
 	// messageChan is the channel to output message
 	messageChan chan *entities.Message
-	// messageList is the list to store all messages
-	messageList []*entities.Message
 	// maps each client to its client handler (required channels)
 	clients map[string]*clientHandler
 	// clientsLock allows multiple readers or one writer to access clients map at the same time
@@ -70,7 +68,6 @@ func InitCollectingProcess(address net.Addr, maxBufferSize uint16, templateTTL u
 		stopChan:      make(chan bool),
 		useMsgChan:    useMsgChan,
 		messageChan:   make(chan *entities.Message),
-		messageList:   make([]*entities.Message, 0),
 		clients:       make(map[string]*clientHandler),
 	}
 	return collectProc, nil
@@ -96,13 +93,6 @@ func (cp *collectingProcess) GetMsgChan() chan *entities.Message {
 		return nil
 	}
 	return cp.messageChan
-}
-
-func (cp *collectingProcess) GetMsgList() []*entities.Message {
-	if cp.useMsgChan {
-		return nil
-	}
-	return cp.messageList
 }
 
 func (cp *collectingProcess) createClient() *clientHandler {
@@ -158,8 +148,6 @@ func (cp *collectingProcess) decodePacket(packetBuffer *bytes.Buffer, exportAddr
 	if cp.useMsgChan {
 		// the thread(s)/client(s) executing the code will get blocked until the message is consumed/read in other goroutines.
 		cp.messageChan <- &message
-	} else {
-		cp.messageList = append(cp.messageList, &message)
 	}
 	return &message, nil
 }
--- a/pkg/collector/process.go
+++ b/pkg/collector/process.go
@@ -43,8 +43,6 @@ type collectingProcess struct {
 	maxBufferSize uint16
 	// chanel to receive stop information
 	stopChan chan bool
-	// useMsgChan indicates whether to use message channel to output message
-	useMsgChan bool
 	// messageChan is the channel to output message
 	messageChan chan *entities.Message
 	// maps each client to its client handler (required channels)
@@ -58,7 +56,7 @@ type clientHandler struct {
 	errChan    chan bool
 }
 
-func InitCollectingProcess(address net.Addr, maxBufferSize uint16, templateTTL uint32, useMsgChan bool) (*collectingProcess, error) {
+func InitCollectingProcess(address net.Addr, maxBufferSize uint16, templateTTL uint32) (*collectingProcess, error) {
 	collectProc := &collectingProcess{
 		templatesMap:  make(map[uint32]map[uint16][]*entities.InfoElement),
 		templatesLock: sync.RWMutex{},
@@ -66,7 +64,6 @@ func InitCollectingProcess(address net.Addr, maxBufferSize uint16, templateTTL u
 		address:       address,
 		maxBufferSize: maxBufferSize,
 		stopChan:      make(chan bool),
-		useMsgChan:    useMsgChan,
 		messageChan:   make(chan *entities.Message),
 		clients:       make(map[string]*clientHandler),
 	}
@@ -89,9 +86,6 @@ func (cp *collectingProcess) Stop() {
 }
 
 func (cp *collectingProcess) GetMsgChan() chan *entities.Message {
-	if !cp.useMsgChan {
-		return nil
-	}
 	return cp.messageChan
 }
 
@@ -145,10 +139,8 @@ func (cp *collectingProcess) decodePacket(packetBuffer *bytes.Buffer, exportAddr
 		}
 		message.Set = set
 	}
-	if cp.useMsgChan {
-		// the thread(s)/client(s) executing the code will get blocked until the message is consumed/read in other goroutines.
-		cp.messageChan <- &message
-	}
+	// the thread(s)/client(s) executing the code will get blocked until the message is consumed/read in other goroutines.
+	cp.messageChan <- &message
 	return &message, nil
 }
 
--- a/pkg/collector/process.go
+++ b/pkg/collector/process.go
@@ -47,6 +47,8 @@ type collectingProcess struct {
 	useMsgChan bool
 	// messageChan is the channel to output message
 	messageChan chan *entities.Message
+	// messageList is the list to store all messages
+	messageList []*entities.Message
 	// maps each client to its client handler (required channels)
 	clients map[string]*clientHandler
 	// clientsLock allows multiple readers or one writer to access clients map at the same time
@@ -68,6 +70,7 @@ func InitCollectingProcess(address net.Addr, maxBufferSize uint16, templateTTL u
 		stopChan:      make(chan bool),
 		useMsgChan:    useMsgChan,
 		messageChan:   make(chan *entities.Message),
+		messageList:   make([]*entities.Message, 0),
 		clients:       make(map[string]*clientHandler),
 	}
 	return collectProc, nil
@@ -93,6 +96,13 @@ func (cp *collectingProcess) GetMsgChan() chan *entities.Message {
 		return nil
 	}
 	return cp.messageChan
+}
+
+func (cp *collectingProcess) GetMsgList() []*entities.Message {
+	if cp.useMsgChan {
+		return nil
+	}
+	return cp.messageList
 }
 
 func (cp *collectingProcess) createClient() *clientHandler {
@@ -148,6 +158,8 @@ func (cp *collectingProcess) decodePacket(packetBuffer *bytes.Buffer, exportAddr
 	if cp.useMsgChan {
 		// the thread(s)/client(s) executing the code will get blocked until the message is consumed/read in other goroutines.
 		cp.messageChan <- &message
+	} else {
+		cp.messageList = append(cp.messageList, &message)
 	}
 	return &message, nil
 }
--- a/pkg/collector/process.go
+++ b/pkg/collector/process.go
@@ -19,6 +19,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"net"
+	"strings"
 	"sync"
 	"time"
 
@@ -42,8 +43,8 @@ type collectingProcess struct {
 	maxBufferSize uint16
 	// chanel to receive stop information
 	stopChan chan bool
-	// packet list
-	messages []*entities.Message
+	// messageChan is the channel to output message
+	messageChan chan *entities.Message
 	// maps each client to its client handler (required channels)
 	clients map[string]*clientHandler
 	// clientsLock allows multiple readers or one writer to access clients map at the same time
@@ -55,7 +56,7 @@ type clientHandler struct {
 	errChan    chan bool
 }
 
-func InitCollectingProcess(address net.Addr, maxBufferSize uint16, templateTTL uint32) (*collectingProcess, error) {
+func InitCollectingProcess(address net.Addr, maxBufferSize uint16, templateTTL uint32, messageChan chan *entities.Message) (*collectingProcess, error) {
 	collectProc := &collectingProcess{
 		templatesMap:  make(map[uint32]map[uint16][]*entities.InfoElement),
 		templatesLock: sync.RWMutex{},
@@ -63,7 +64,7 @@ func InitCollectingProcess(address net.Addr, maxBufferSize uint16, templateTTL u
 		address:       address,
 		maxBufferSize: maxBufferSize,
 		stopChan:      make(chan bool),
-		messages:      make([]*entities.Message, 0),
+		messageChan:   messageChan,
 		clients:       make(map[string]*clientHandler),
 	}
 	return collectProc, nil
@@ -79,10 +80,9 @@ func (cp *collectingProcess) Start() {
 
 func (cp *collectingProcess) Stop() {
 	cp.stopChan <- true
-}
-
-func (cp *collectingProcess) GetMessages() []*entities.Message {
-	return cp.messages
+	if cp.messageChan != nil {
+		close(cp.messageChan)
+	}
 }
 
 func (cp *collectingProcess) createClient() *clientHandler {
@@ -110,8 +110,10 @@ func (cp *collectingProcess) getClientCount() int {
 	return len(cp.clients)
 }
 
-func (cp *collectingProcess) decodePacket(packetBuffer *bytes.Buffer) (*entities.Message, error) {
+func (cp *collectingProcess) decodePacket(packetBuffer *bytes.Buffer, exportAddress string) (*entities.Message, error) {
 	message := entities.Message{}
+	exportAddr := strings.Split(exportAddress, ":")[0]
+	message.ExportAddress = exportAddr
 	var setID, length uint16
 	err := util.Decode(packetBuffer, binary.BigEndian, &message.Version, &message.BufferLength, &message.ExportTime, &message.SeqNumber, &message.ObsDomainID, &setID, &length)
 	if err != nil {
@@ -133,7 +135,10 @@ func (cp *collectingProcess) decodePacket(packetBuffer *bytes.Buffer) (*entities
 		}
 		message.Set = set
 	}
-	cp.messages = append(cp.messages, &message)
+	addOriginalExporterInfo(&message)
+	if cp.messageChan != nil {
+		cp.messageChan <- &message
+	}
 	return &message, nil
 }
 
@@ -299,4 +304,48 @@ func getFieldLength(dataBuffer *bytes.Buffer) int {
 	lengthBuff = dataBuffer.Next(2)
 	util.Decode(bytes.NewBuffer(lengthBuff), binary.BigEndian, &lengthTwoBytes)
 	return int(lengthTwoBytes)
+}
+
+// addOriginalExporterInfo adds originalExporterIPv4Address and originalObservationDomainId to records in message set
+func addOriginalExporterInfo(message *entities.Message) error {
+	set := message.Set
+	records := set.GetRecords()
+	for _, record := range records {
+		var originalExporterIPv4Address, originalObservationDomainId *entities.InfoElementWithValue
+
+		// Add originalExporterIPv4Address
+		ie, err := registry.GetInfoElement("originalExporterIPv4Address", registry.IANAEnterpriseID)
+		if err != nil {
+			return fmt.Errorf("IANA Registry is not loaded correctly with originalExporterIPv4Address.")
+		}
+		if set.GetSetType() == entities.Template {
+			originalExporterIPv4Address = entities.NewInfoElementWithValue(ie, nil)
+		} else if set.GetSetType() == entities.Data {
+			originalExporterIPv4Address = entities.NewInfoElementWithValue(ie, net.ParseIP(message.ExportAddress))
+		} else {
+			return fmt.Errorf("Set type %d is not supported.", set.GetSetType())
+		}
+		_, err = record.AddInfoElement(originalExporterIPv4Address, false)
+		if err != nil {
+			return err
+		}
+
+		// Add originalObservationDomainId
+		ie, err = registry.GetInfoElement("originalObservationDomainId", registry.IANAEnterpriseID)
+		if err != nil {
+			return fmt.Errorf("IANA Registry is not loaded correctly with originalObservationDomainId.")
+		}
+		if set.GetSetType() == entities.Template {
+			originalObservationDomainId = entities.NewInfoElementWithValue(ie, nil)
+		} else if set.GetSetType() == entities.Data {
+			originalObservationDomainId = entities.NewInfoElementWithValue(ie, message.ObsDomainID)
+		} else {
+			return fmt.Errorf("Set type %d is not supported.", set.GetSetType())
+		}
+		_, err = record.AddInfoElement(originalObservationDomainId, false)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
--- a/pkg/collector/process.go
+++ b/pkg/collector/process.go
@@ -43,6 +43,8 @@ type collectingProcess struct {
 	maxBufferSize uint16
 	// chanel to receive stop information
 	stopChan chan bool
+	// useMsgChan indicates whether to use message channel to output message
+	useMsgChan bool
 	// messageChan is the channel to output message
 	messageChan chan *entities.Message
 	// maps each client to its client handler (required channels)
@@ -56,7 +58,7 @@ type clientHandler struct {
 	errChan    chan bool
 }
 
-func InitCollectingProcess(address net.Addr, maxBufferSize uint16, templateTTL uint32, messageChan chan *entities.Message) (*collectingProcess, error) {
+func InitCollectingProcess(address net.Addr, maxBufferSize uint16, templateTTL uint32, useMsgChan bool) (*collectingProcess, error) {
 	collectProc := &collectingProcess{
 		templatesMap:  make(map[uint32]map[uint16][]*entities.InfoElement),
 		templatesLock: sync.RWMutex{},
@@ -64,7 +66,8 @@ func InitCollectingProcess(address net.Addr, maxBufferSize uint16, templateTTL u
 		address:       address,
 		maxBufferSize: maxBufferSize,
 		stopChan:      make(chan bool),
-		messageChan:   messageChan,
+		useMsgChan:    useMsgChan,
+		messageChan:   make(chan *entities.Message),
 		clients:       make(map[string]*clientHandler),
 	}
 	return collectProc, nil
@@ -83,6 +86,13 @@ func (cp *collectingProcess) Stop() {
 	if cp.messageChan != nil {
 		close(cp.messageChan)
 	}
+}
+
+func (cp *collectingProcess) GetMsgChan() chan *entities.Message {
+	if !cp.useMsgChan {
+		return nil
+	}
+	return cp.messageChan
 }
 
 func (cp *collectingProcess) createClient() *clientHandler {
@@ -135,8 +145,8 @@ func (cp *collectingProcess) decodePacket(packetBuffer *bytes.Buffer, exportAddr
 		}
 		message.Set = set
 	}
-	addOriginalExporterInfo(&message)
-	if cp.messageChan != nil {
+	if cp.useMsgChan {
+		// the thread(s)/client(s) executing the code will get blocked until the message is consumed/read in other goroutines.
 		cp.messageChan <- &message
 	}
 	return &message, nil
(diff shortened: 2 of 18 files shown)


srikartati (commented):
Thanks for the PR @zyiou.
I did not review pkg/intermediate code yet completely.

Overall, messageChan usage in pkg/collector looks ok but I am wondering if it covers all the cases or not. I have some related comments on that.

pkg/collector/process.go
```diff
@@ -55,15 +56,15 @@ type clientHandler struct {
 	errChan    chan bool
 }
 
-func InitCollectingProcess(address net.Addr, maxBufferSize uint16, templateTTL uint32) (*collectingProcess, error) {
+func InitCollectingProcess(address net.Addr, maxBufferSize uint16, templateTTL uint32, messageChan chan *entities.Message) (*collectingProcess, error) {
```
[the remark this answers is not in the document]
zyiou: Makes sense. I should put message chan internally and let user get message as needed.

pkg/collector/process.go
```diff
@@ -42,8 +43,8 @@ type collectingProcess struct {
 	maxBufferSize uint16
 	// chanel to receive stop information
 	stopChan chan bool
-	// packet list
-	messages []*entities.Message
+	// messageChan is the channel to output message
+	messageChan chan *entities.Message
```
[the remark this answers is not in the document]
zyiou: I changed `messageChan` to be created internally, then let user pass a `bool` in parameter to decide whether to use `messageChan` or not.

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+		}
+	}
+	for _, ieWithValue := range elements {
+		if ieWithValue.Element.Name == "sourcePodName" && ieWithValue.Value != "" {
+			// add record to tupleRecordMap when the record is from source node
+			a.tupleRecordMap[tuple] = append(a.tupleRecordMap[tuple], record)
+			return
+		}
+	}
```
srikartati, line 122: I did not follow this logic. What happens if we get the record from source node first and destination node later?

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,218 @@
...
+	records := message.Set.GetRecords()
+	a.tupleRecordLock.Lock()
+	defer a.tupleRecordLock.Unlock()
+	for _, record := range records {
+		tuple, err := getTupleFromRecord(record)
+		if err != nil {
+			return err
+		}
+		a.correlateRecords(tuple, record)
```
srikartati, line 86: What happens if we get multiple records for the same flow from the same source or destination node?

pkg/collector/process.go
```diff
@@ -55,15 +58,16 @@ type clientHandler struct {
...
 	collectProc := &collectingProcess{
 		templatesMap:  make(map[uint32]map[uint16][]*entities.InfoElement),
 		templatesLock: sync.RWMutex{},
 		templateTTL:   templateTTL,
 		address:       address,
 		maxBufferSize: maxBufferSize,
 		stopChan:      make(chan bool),
-		messages:      make([]*entities.Message, 0),
+		useMsgChan:    useMsgChan,
+		messageChan:   make(chan *entities.Message),
```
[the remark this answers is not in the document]
zyiou: Makes sense. Updated. Thanks!

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// getTupleFromRecord returns 5-tuple from data record
+func getTupleFromRecord(record entities.Record) (Tuple, error) {
+	var srcIP, dstIP uint32
+	var srcPort, dstPort uint16
+	var proto uint8
+	// count is for checking whether 5-tuple is fully collected
+	count := 0
+	for _, infoElementWithValue := range record.GetInfoElements() {
```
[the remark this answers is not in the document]
zyiou: Sure. That makes sense.

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,218 @@
...
+		a.tupleRecordMap[tuple] = make([]entities.Record, 0)
+		a.tupleRecordMap[tuple] = append(a.tupleRecordMap[tuple], record)
+		return
+	}
+	existingRecords := a.tupleRecordMap[tuple]
+	elements := record.GetInfoElements()
+	for _, existingRecord := range existingRecords {
+		existingElements := existingRecord.GetInfoElements()
+		for i, ieWithValue := range existingElements {
+			if ieWithValue.Value == "" {
```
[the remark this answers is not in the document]
zyiou: Agree. I have changed the correlation code using following logic:
1. in `correlateRecords` function, make sure to fill all missing fields when the record is from source node.
    1.1. if current record is from source, fill destination related fields from existing records.
    1.2 if current record is from destination, fill destination related fields to existing records. 
2. add current record to map
3. clean the map `removeDuplicates`
    3.1 if there is record from source node, discard all records from destination node (considered as duplicates)
    3.2 if there is no record from source node, keep all records from destination node (pending to be filled)
Overall we only want to keep one record when same record is sent from source and destination node. I filled and kept record from source node in my updated code. Comment if you have any concerns or there are cases that are not covered. Thanks!

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,218 @@
...
+	messageChan chan *entities.Message
+	// workerNum is the number of workers to process message
+	workerNum int
+	// workerList is the list of workers
+	workerList []*worker
+}
+
+type Tuple struct {
+	// SourceAddress is the encoded format of sourceIPv4Address
+	SourceAddress uint32
```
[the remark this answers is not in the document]
zyiou: `net.IP` is a slice type, which cannot be used as struct field as map key because keyType of map should be [comparable](https://golang.org/ref/spec#Comparison_operators). reference: https://blog.golang.org/maps. 
Changed to array type which will work for both IPv4 and IPv6

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,218 @@
...
+func (a *aggregation) GetTupleRecordMap() map[Tuple][]entities.Record {
+	a.tupleRecordLock.RLock()
+	defer a.tupleRecordLock.RUnlock()
+	return a.tupleRecordMap
+}
+
+// correlateRecords fills records info by correlating incoming and current records
+// only records from source node will be added to tupleRecordMap
+func (a *aggregation) correlateRecords(tuple Tuple, record entities.Record) {
+	if _, exist := a.tupleRecordMap[tuple]; !exist {
```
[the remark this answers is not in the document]
zyiou: Maybe in the future. what is the concern here?

pkg/intermediate/aggregate.go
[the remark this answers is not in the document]
srikartati: Let's use string type for IPs?

pkg/intermediate/aggregate.go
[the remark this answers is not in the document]
srikartati: As flows become inactive, we do not have to keep their records in the map. If we do not delete, memory can blow up.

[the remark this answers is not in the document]
zyiou: Got it. I added a delete function to support.

pkg/intermediate/aggregate.go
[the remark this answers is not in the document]
zyiou: Great idea. Thanks!

srikartati (commented):
The explanation of correlation logic sounds good. Just keeping one record is good and makes the exporter job easy in IPFIX mediator.
However, don't we need to take care of the following stats fields if we are discarding records? With current logic one of the flow records statistics will be ignored. If you want to take care of this in separate PR, that is also fine.

packetTotalCountFromSourceNode, octetTotalCountFromSourceNode, packetDeltaCountFromSourceNode... reverse_OctetDeltaCountFromSourceNode (8 fields in total); packetTotalCountFromDestinationNode, octetTotalCountFromDestinationNode, packetDeltaCountFromDestinationNode... reverse_OctetDeltaCountFromDestinationNode (8 fields in total)

pkg/entities/record.go
```diff
@@ -38,15 +38,18 @@ type Record interface {
...
-	buff       bytes.Buffer
-	len        uint16
-	fieldCount uint16
-	templateID uint16
-	elements   []*InfoElementWithValue
+	buff        bytes.Buffer
+	len         uint16
+	fieldCount  uint16
+	templateID  uint16
+	elements    []*InfoElementWithValue
```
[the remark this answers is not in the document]
zyiou: The reason to use element list is that in `sendRefreshedTemplates`, we need to get the template elements in sequence to re-send it. Using map will not keep the order. https://github.com/vmware/go-ipfix/blob/ceccafac0e51e43e8aa896b22e3769c2f7eeb27a/pkg/exporter/process.go#L227

pkg/entities/record.go
```diff
@@ -38,15 +38,18 @@ type Record interface {
 	GetTemplateID() uint16
 	GetFieldCount() uint16
 	GetInfoElements() []*InfoElementWithValue
+	GetInfoElement(name string) *InfoElementWithValue
+	ContainsInfoElement(name string) bool
```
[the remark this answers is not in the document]
zyiou: There are pros and cons. Providing a `ContainsInfoElement` method will make it easier when we check whether many info elements are in the record (e.g. for [getting 5-tuple info](https://github.com/vmware/go-ipfix/blob/b51831a0ba8ea36c72c24cbff14a65610a2ba060/pkg/intermediate/aggregate.go#L178)) because checking whether the key is in the map will return two values:
```
if val, exist := map["key"]; ok{
   dosomething()
}
```
But it only works well specially for this case, if users want to read the map in other ways, two functions are not that good. Which way do you prefer?

zyiou:
> The explanation of correlation logic sounds good. Just keeping one record is good and makes the exporter job easy in IPFIX mediator.
> However, don't we need to take care of the following stats fields if we are discarding records? With current logic one of the flow records statistics will be ignored. If you want to take care of this in separate PR, that is also fine.
> 
> packetTotalCountFromSourceNode, octetTotalCountFromSourceNode, packetDeltaCountFromSourceNode... reverse_OctetDeltaCountFromSourceNode (8 fields in total); packetTotalCountFromDestinationNode, octetTotalCountFromDestinationNode, packetDeltaCountFromDestinationNode... reverse_OctetDeltaCountFromDestinationNode (8 fields in total)

If two duplicate flows are received from source and destination node, will the reverse packet/bytes from source node and  packet/bytes from destination node be the same? e.g. is it possible to use `reverseOctetDeltaCount` from source node to calculate `octetDeltaCountFromDestinationNode`?

pkg/entities/record.go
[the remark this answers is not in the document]
srikartati: How about having one function to get the map and have couple of simple utilities to get the element and check whether element exists or not?

pkg/entities/record.go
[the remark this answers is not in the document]
srikartati: Oh ok. Let's change the field name of the element list to represent that.

srikartati:
> If two duplicate flows are received from source and destination node, will the reverse packet/bytes from source node and packet/bytes from destination node be the same? e.g. is it possible to use `reverseOctetDeltaCount` from source node to calculate `octetDeltaCountFromDestinationNode`?

Source and destination node won't be sending the flow records at the same instant (different export time). Therefore, we cannot rely on reverse stats to deduce the stats for flow record on destination node. It is probably better the copy of the stats coming from both source and destination nodes.

pkg/entities/record.go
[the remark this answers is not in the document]
zyiou: Sure. Thanks!

zyiou:
> > If two duplicate flows are received from source and destination node, will the reverse packet/bytes from source node and packet/bytes from destination node be the same? e.g. is it possible to use `reverseOctetDeltaCount` from source node to calculate `octetDeltaCountFromDestinationNode`?
> 
> Source and destination node won't be sending the flow records at the same instant (different export time). Therefore, we cannot rely on reverse stats to deduce the stats for flow record on destination node. It is probably better the copy of the stats coming from both source and destination nodes.

Got it. Maybe we can take care of these stats in another PR. What do you think?

cmd/collector/collector.go
```diff
@@ -130,28 +130,18 @@ func run() error {
 		return fmt.Errorf("input given ipfix.transport flag is not supported or valid")
 	}
 	// Initialize collecting process
-	cp, err := collector.InitCollectingProcess(netAddr, 65535, 0)
+	cp, err := collector.InitCollectingProcess(netAddr, 65535, 0, false)
```
srikartati, line 133: does useMsgChan has to be true here?

pkg/collector/process.go
```diff
@@ -42,8 +43,10 @@ type collectingProcess struct {
 	maxBufferSize uint16
 	// chanel to receive stop information
 	stopChan chan bool
-	// packet list
-	messages []*entities.Message
+	// useMsgChan indicates whether to use message channel to output message
+	useMsgChan bool
```
srikartati, line 47: does useMsgChan have any other use case than usage in unit tests?

pkg/entities/record.go
```diff
@@ -76,11 +79,12 @@ type templateRecord struct {
...
-			buff:       bytes.Buffer{},
-			len:        0,
-			fieldCount: count,
-			templateID: id,
-			elements:   make([]*InfoElementWithValue, 0),
+			buff:        bytes.Buffer{},
+			len:         0,
+			fieldCount:  count,
+			templateID:  id,
+			elementList: make([]*InfoElementWithValue, 0),
```
srikartati, line 86: could you change the name of elementList like orderedElementList or something that signifies why you need a list along with the map?

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/vmware/go-ipfix/pkg/entities"
+	"github.com/vmware/go-ipfix/pkg/registry"
+)
+
+type aggregation struct {
+	// tupleRecordMap maps each connection with its records
+	tupleRecordMap map[Tuple][]entities.Record
+	// tupleRecordLock allows multiple readers or one writer at the same time
+	tupleRecordLock sync.RWMutex
+	// workerPool is for storing worker channels to process message
```
srikartati, line 17: s/process message/process the messages.

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,284 @@
...
+	workerPool chan chan *entities.Message
+	// messageChan is the channel to receive message
```
srikartati, line 19: ... receive the message

srikartati (commented):
Overall LGTM. Few comments. Let's resolve them as soon as we can and include this v0.3.0 release. Thanks.

pkg/collector/process.go
zyiou: I cannot figure out any other. I suppose we can remove it.

pkg/entities/record.go
zyiou: Sure. Updated. Thanks!

srikartati (approved):
Thanks for making the changes. 
LGTM.

srikartati:
> > > If two duplicate flows are received from source and destination node, will the reverse packet/bytes from source node and packet/bytes from destination node be the same? e.g. is it possible to use `reverseOctetDeltaCount` from source node to calculate `octetDeltaCountFromDestinationNode`?
> > 
> > 
> > Source and destination node won't be sending the flow records at the same instant (different export time). Therefore, we cannot rely on reverse stats to deduce the stats for flow record on destination node. It is probably better the copy of the stats coming from both source and destination nodes.
> 
> Got it. Maybe we can take care of these stats in another PR. What do you think?

Sure, this can be addressed in a separate PR.

cmd/collector/collector.go
```diff
@@ -138,20 +138,10 @@ func run() error {
...
-					klog.Info("Processing IPFIX message")
-					messageReceived <- ipfixMessages[i]
-					msgCount++
-				}
-			}
-			time.Sleep(5 * time.Millisecond)
+		msgChan := cp.GetMsgChan()
+		for message := range msgChan {
+			klog.Info("Processing IPFIX message")
+			messageReceived <- message
```
shihhaoli, line 144: Just curious, why not using the channel from cp.GetMsgChan() directly inside signalHandler instead of piping one channel to another channel?

For example, can we extend signalHandler (or rename it to eventHandler) as

func eventHandler(stopCh chan struct{}, cp *collectingProcess) {
	cp.Start()
	msgChan := cp.GetMsgChan()
	signalCh := make(chan os.Signal, 1)
	signal.Notify(signalCh, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)

        for {
                select {
                case msg := <-msgChan:
                        printIPFIXMessage(msg)
                case <-signalCh:
                        close(stopCh)
                        return
                }
        }
}

Then there is no need for the go func on lines 139-146?
Thanks.

srikartati: Hi Shihao,
Yes, I noticed that too. We do not need an extra message channel. Thought of taking care of this in follow-up patch as we want to move on with v0.3.0 release to have this available for changing Antrea e2e tests as part of Antrea 0.11.0 release.

However, I think it is best to have a separate signal handler that sends out signal on stopCh. This way we can have multiple independent go routines that start their functionalities (using `Run` as in convention of Antrea or other well known go applications) and they can be stopped with the stopCh. Extending this collector application will be easier.

shihhaoli: Yes, a dedicated signal handler sounds good too. Then we can have the original go func (i.e. lines 139-146) calling printIPFIXMessage on the messages from the message channel directly. Is that right? Thanks.

srikartati: Yes that's correct. It taken in stopCh as argument.

srikartati: Btw, please feel free to raise a PR against release-0.3.0 branch. We can add this to patch release.

pkg/collector/process.go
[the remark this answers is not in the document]
shihhaoli: I do not see the the boolean useMsgChan here. So it is updated in another branch?

pkg/collector/process.go
```diff
@@ -133,7 +139,8 @@ func (cp *collectingProcess) decodePacket(packetBuffer *bytes.Buffer) (*entities
 		}
 		message.Set = set
 	}
-	cp.messages = append(cp.messages, &message)
+	// the thread(s)/client(s) executing the code will get blocked until the message is consumed/read in other goroutines.
```
shihhaoli, line 142: I assume the boolean useMsgChan will be checked here?

zyiou: We removed the `useMsgChan` because we will create a message channel to output our messages anyways. `useMsgChan` was used for cases that users want to get message list instead of from channel. But we decide to remove it since users can get message list directly from the message channel.

shihhaoli: Got it. Another question, who will be the consumer of this messageChan, intermediate process?
And I assume there should be only one consumer, is that right? Thanks.

pkg/exporter/process.go
```diff
@@ -101,7 +101,7 @@ func InitExportingProcess(collectorAddr net.Addr, obsID uint32, tempRefTimeout u
 func (ep *ExportingProcess) AddSetAndSendMsg(setType entities.ContentType, set entities.Set) (int, error) {
```
shihhaoli, line 101: Is the ExportingProcess in this file a module running on Antrea Agent or Flow Aggregator?
Or both modules share the same code?

pkg/collector/process.go
srikartati: Yes, there can be only one consumer. It can be either a standalone collector or the intermediate process but not both.

pkg/exporter/process.go
srikartati: Yes, they share the same code. Essentially exporting process is considered a library that can be consumed by users in different places.

shihhaoli (commented):
I have a question not directly-related to this commit.
In set.go

func (s *set) FinishSet() {
	// TODO:Add padding when multiple sets are sent in single IPFIX message                                     
        if !s.isDecoding {
                // Add length to the message                                                                        
                byteSlice := s.buffer.Bytes()
		setOffset := s.buffer.Len() - int(s.currLen)
                binary.BigEndian.PutUint16(byteSlice[setOffset+2:setOffset+4], s.currLen)
                // Reset the length                                                                                 
                s.currLen = 0
	}
}

By looking at the way setOffset is calculated, it looks like there could be multiple Sets in a set object's buffer?
Also why do we reset s.currLen to 0? Do we re-use the set object somewhere?
Thanks.

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,284 @@
...
+	messageChan chan *entities.Message
+	// workerNum is the number of workers to process the messages
+	workerNum int
+	// workerList is the list of workers
+	workerList []*worker
+	// correlateFields are the fields to be filled in correlating process
+	correlateFields []string
+}
+
+type Tuple struct {
```
shihhaoli, line 29: Should we name this struct something like ConnectionKey or FlowKey?
Tuple may sound too general for a public type?

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/vmware/go-ipfix/pkg/registry"
+)
+
+type aggregation struct {
+	// tupleRecordMap maps each connection with its records
+	tupleRecordMap map[Tuple][]entities.Record
+	// tupleRecordLock allows multiple readers or one writer at the same time
+	tupleRecordLock sync.RWMutex
+	// workerPool is for storing worker channels to process the messages
+	workerPool chan chan *entities.Message
```
shihhaoli, line 18: Is that possible to define a single channel shared by all workers,
i.e. all workers consume messages from the same channel and every worker can get a message from the channel when it is available?

zyiou: The reason of passing channels over channels is that it is sending message(job) to available worker and make sure each worker is processing one message(job) at a time. The worker pool contains the channels that are available to receive message.

shihhaoli: Thanks for the explanation. I did a simple test. It looks like a single channel shared by multiple workers can achieve the same goal. For example,

```
package main

import (
        "fmt"
        "time"
)

const NUM_WORKERS int = 3

func produce(jobs chan<- int) {
	for i := 1; i <= 10; i++ {
                jobs <- i
	}
	close(jobs)
}

func consume(worker int, jobs <-chan int, done chan<- bool) {
        for job := range jobs {
                fmt.Printf("[%v] worker %d got %d\n", time.Now().Format(time.Stamp), worker, job)
                time.Sleep(time.Duration(worker) * time.Second)
        }
        done <- true
}

func main() {
        jobs := make(chan int)
	done := make(chan bool)

        go produce(jobs)

	for i := 1; i <= NUM_WORKERS; i++ {
		go consume(i, jobs, done)
	}

        for i := 1; i <= NUM_WORKERS; i++ {
		<-done
	}
}
```

Then the output is like:

$ go run  test.go
[Nov 11 20:57:03] worker 3 got 1
[Nov 11 20:57:03] worker 1 got 2
[Nov 11 20:57:03] worker 2 got 3
[Nov 11 20:57:04] worker 1 got 4
[Nov 11 20:57:05] worker 2 got 5
[Nov 11 20:57:05] worker 1 got 6
[Nov 11 20:57:06] worker 3 got 7
[Nov 11 20:57:06] worker 1 got 8
[Nov 11 20:57:07] worker 2 got 9
[Nov 11 20:57:07] worker 1 got 10

pkg/intermediate/aggregate.go
srikartati: There is no tuples type in golang, so it might be ok.
I do agree the name `FlowKey` is more informative. 
We use `ConnectionKey` in Antrea, however, there is no context of connection here, so `FlowKey` is good.

pkg/intermediate/aggregate.go
zyiou: Yes, it can also achieve that. The way we did for intermediate process is that we create several workers (each is a goroutine with a channel). then we put the channel into the worker pool for producers to receive message. The channel (worker) which gets the message will start processing it and then notify the worker pool by trying to sending the channel back to worker pool. It mimics the way of sending jobs to go routine and this model is used by similar project [here](https://github.com/cloudflare/goflow/blob/efa7272e2cc3aa91c1d904e982b4f8a8eba28460/decoders/decoder.go) and some [examples](https://medium.com/better-programming/hands-on-go-concurrency-the-producer-consumer-pattern-c42aab4e3bd2) for comparison.
Actually I don't see one is much better than the other. The advantage I can see of using passing channel over channel is that only available worker will be put into the worker pool to process the message and the finish of each job can be notified.

shihhaoli: Thanks for the details. My example is actually copied from the Single Producer Multiple Consumer case in the same [examples](https://medium.com/better-programming/hands-on-go-concurrency-the-producer-consumer-pattern-c42aab4e3bd2) link. So I was curious why we require channel over channel in our scenario.
Sorry one more question. When you said "the finish of each job can be notified" is an advantage of using channel over channel, did you just mean this can be done with channel over channel, but it is not implemented this way in our code, Is that right? because I do not see we ack the finish of job in the channel over channel. Thanks again!

srikartati:
> I have a question not directly-related to this commit.
> In set.go
> 
> ```
> func (s *set) FinishSet() {
> 	// TODO:Add padding when multiple sets are sent in single IPFIX message                                     
>         if !s.isDecoding {
>                 // Add length to the message                                                                        
>                 byteSlice := s.buffer.Bytes()
> 		setOffset := s.buffer.Len() - int(s.currLen)
>                 binary.BigEndian.PutUint16(byteSlice[setOffset+2:setOffset+4], s.currLen)
>                 // Reset the length                                                                                 
>                 s.currLen = 0
> 	}
> }
> ```
> 
> By looking at the way setOffset is calculated, it looks like there could be multiple Sets in a set object's buffer?
> Also why do we reset s.currLen to 0? Do we re-use the set object somewhere?
> Thanks.

Originally FinishSet was part of Msg struct and that code continued. Yes, setOffset was calculated by considering the fact that msgBuf can have multiple sets. This can be refactored as Set object corresponds to only one set. However, we do need the padding in the set when supporting multiple sets in a message. We currently do not support this scenario.
Same with s.currLen = 0.. this is not needed. As far as I know we do not reuse the set object.

All in all there is scope for refactoring and simplifying the code. Thanks for pointing.

pkg/intermediate/aggregate.go
zyiou: The suggestion makes sense to me. And it is easier to understand if not using channel over channel. will try to modify the code in a separate PR. Thanks!

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,284 @@
...
+
+func (a *aggregation) Stop() {
+	for _, worker := range a.workerList {
+		worker.stop()
+	}
+}
+
+// AggregateMsgBy5Tuple gets 5-tuple info from records in message and stores in cache
+func (a *aggregation) AggregateMsgBy5Tuple(message *entities.Message) error {
+	addOriginalExporterInfo(message)
```
shihhaoli, line 74: Should we move addOriginalExporterInfo() after line 77 since we skip Template Sets?

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,284 @@
...
+		proto = protoNum
+	} else {
+		return Tuple{}, fmt.Errorf("missing 5-tuple value(s) in the record.")
+	}
+	return Tuple{srcIP, dstIP, proto, srcPort, dstPort}, nil
+}
+
+// addOriginalExporterInfo adds originalExporterIPv4Address and originalObservationDomainId to records in message set
+func addOriginalExporterInfo(message *entities.Message) error {
+	set := message.Set
```
shihhaoli, line 237: Can a message has multiple Sets?

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+	}
+	return nil
+}
+
+func containsInfoElement(recordMap map[string]*entities.InfoElementWithValue, name string) bool {
+	if _, exist := recordMap[name]; exist {
+		return true
+	}
+	return false
```
shihhaoli, line 283: or just two lines?

```
_, exist := recordMap[name]
return exist
```

pkg/intermediate/aggregate.go
zyiou: Yes currently we only support one set per message. We should support multiple sets per message in the future. There is an issue to track it #55

pkg/intermediate/aggregate.go
zyiou: For template records, currently we keep the `originalExporterIPv4Address` and `originalObservationDomainId` in case we want to re-use the elements from template record when exporting message from flow aggregator.

pkg/intermediate/aggregate.go
zyiou: That's much simpler. Thanks!

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,284 @@
...
+		protoNum, ok := infoElementMap["protocolIdentifier"].Value.(uint8)
+		if !ok {
+			return Tuple{}, fmt.Errorf("protocolIdentifier is not in correct format.")
+		}
+		proto = protoNum
+	} else {
+		return Tuple{}, fmt.Errorf("missing 5-tuple value(s) in the record.")
+	}
+	return Tuple{srcIP, dstIP, proto, srcPort, dstPort}, nil
+}
```
shihhaoli, line 233: We may consider simplifying getTupleFromRecord() as the following, although the error report for the address part may not be precious.

```
func (b *baseRecord) GetInfoElementValue(name string, dataType entities.IEDataType) (interface{}, error) {
        if element, exist := b.elementsMap[name]; exist {
                return entities.DecodeToIEDataType(dataType, element.Value)
                /*                                                                                             
                 * If element.Element.DataType is populated with correct type,
                 * then we don't need to pass dataType as an input. We can just do                                                                              
                 *    return entities.DecodeToIEDataType(element.Element.DataType, element.Value)              
                 */
        } else {
                return nil, fmt.Errorf("missing %s in the record.", name)
        }
}

func getTupleFromRecord(record entities.Record) (Tuple, error) {
        srcPort, err := record.GetInfoElementValue("sourceTransportPort", entities.Unsigned16)
        if err {
                return Tuple{}, err
        }

        dstPort, err := record.GetInfoElementValue("destinationTransportPort", entities.Unsigned16)
        if err {
                return Tuple{}, err
        }

        proto, err := record.GetInfoElementValue("destinationTransportPort", entities.Unsigned8)
        if err {
                return Tuple{}, err
        }

        srcIPAddr, err := record.GetInfoElementValue("sourceIPv4Address", entities.Ipv4Address)
        if err {
                srcIPAddr, err = record.GetInfoElementValue("sourceIPv6Address", entities.Ipv6Address)
        }
        if err {
                return Tuple{}, err
        }
        srcIP := srcIPAddr.String()

        dstIPAddr, err := record.GetInfoElementValue("destinationIPv4Address", entities.Ipv4Address)
        if err {
                dstIPAddr, err = record.GetInfoElementValue("destinationIPv6Address", entities.Ipv6Address)
        }
        if err {
                return Tuple{}, err
        }
        dstIP := dstIPAddr.String()

        return Tuple{srcIP, dstIP, proto, srcPort, dstPort}, nil
}
```

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+		a.correlateRecords(tuple, record)
+	}
+	return nil
+}
+
+func (a *aggregation) GetTupleRecordMap() map[Tuple][]entities.Record {
+	a.tupleRecordLock.RLock()
+	defer a.tupleRecordLock.RUnlock()
+	return a.tupleRecordMap
```
shihhaoli, line 92: I am not sure a lock here is useful or not. My understanding is once the function returns the map, then it is unlocked. Maybe what we really want is to put the lock in the caller?

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,284 @@
...
+func (a *aggregation) removeDuplicates(tuple Tuple) {
+	a.tupleRecordLock.Lock()
+	defer a.tupleRecordLock.Unlock()
+	records := a.tupleRecordMap[tuple]
+	srcRecords := make([]entities.Record, 0)
+	dstRecords := make([]entities.Record, 0)
+	for _, record := range records {
+		if isRecordFromSrc(record) {
+			srcRecords = append(srcRecords, record)
+		} else {
```
shihhaoli, line 142: Do we need to check destinationPodName existing or not?

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,284 @@
...
+			dstRecords = append(dstRecords, record)
+		}
+	}
+	if len(srcRecords) != 0 {
```
shihhaoli, line 146: Is that possible both srcRecords and dstRecords are not empty?

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,284 @@
...
+
+// correlateRecords fills records info by correlating incoming and current records
+func (a *aggregation) correlateRecords(tuple Tuple, record entities.Record) {
+	existingRecords := a.GetTupleRecordMap()[tuple]
+	// only fill the information for record from source node
+	if isRecordFromSrc(record) {
+		var isFilled bool
+		for _, existingRec := range existingRecords {
+			for _, field := range a.correlateFields {
+				if containsInfoElement(record.GetInfoElementMap(), field) {
```
shihhaoli, line 109: Should this be the reverse instead, i.e.

`if containsInfoElement(existingRec.GetInfoElementMap(), field)` ?

pkg/intermediate/aggregate.go
```diff
@@ -0,0 +1,284 @@
...
+					record.GetInfoElementMap()[field].Value = existingRec.GetInfoElementMap()[field].Value
+					isFilled = true
+				}
+			}
+			if isFilled {
+				break
```
shihhaoli, line 115: Why break here?
Is that possible the incoming record (fromSrc) gets field_A from existingRec_1 (fromDst) and field_B from existingRec_2 (fromDst)?
Or all fromSrc records always have the same set of fields populated? Similar for all fromDst records?

pkg/intermediate/aggregate.go
zyiou: yes it makes more sense to put the lock in the caller since map is a reference type.

pkg/intermediate/aggregate.go
zyiou: the correlate the assumption for using `isRecordFromSrc` is that the record will have both sourcePodName and destinationPodName (specific to deal with records exported from Antrea flow exporter)

pkg/intermediate/aggregate.go
zyiou: reminds me that we actually need to check both since they may not use the same template. Thanks!

pkg/intermediate/aggregate.go
zyiou: Currently we don't have use cases that incoming record(fromSrc) will need fill the fields from different dst records. One fromSrc and one fromDst will complete each other. But it is a good point to pay attention in the future.

pkg/intermediate/aggregate.go
srikartati: @zyiou My main point was not having two interfaces and just to have one. You had both `GetInfoElementValue` and `ContainsInfoElement`.

Above modification to`GetInfoElementValue` looks good. However, I suggest to return `exists` bool rather than error like (interface{}, bool, error). Caller can consume `exists`.

 A note in above code: It's better to return `nil, err` rather than `Tuple{}, err`. We can save on malloc.

zyiou: Got it. will update that. Thanks!

pkg/intermediate/aggregate.go
srikartati: Hi Shihao,
Went through this thread in a more detailed way. The approach of using one channel sounds good. However, I have a question related to the code snippet provided here: https://github.com/vmware/go-ipfix/pull/51#discussion_r521841193

Will all consume go routines exit properly? A bool signal on unbuffered channel `done` sent by say consume-1, and it is received by main, then main exits and consume-1 exits too. What about consume-2 and consume-3 which are in blocked state after sending on `done`. As there are no receivers for these signals, they will not exit properly. Do you agree?

shihhaoli: Yes, the complete code should be doing
```
        for i := 1; i <= NUM_WORKERS; i++ {
		<-done
	}
```
I was too lazy :-) Thanks.

PR #53, opened 2020-10-21
Upgrade Go to 1.15 and add golangci-lint

### LINKED ISSUE #17: Add golinters in CI
For PRs, we need to run golinters as part of CI.

zyiou:
Fixes #17 and bump go to 1.15


.github/workflows/go.yml
```diff
@@ -43,3 +43,19 @@ jobs:
...
+
+
+  golangci-lint:
+    name: Golangci-lint
+    runs-on: [ubuntu-18.04]
+    steps:
+      - name: Set up Go 1.13
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13
```
srikartati, line 55: May be we can bump up to higher go version? Go 1.15?

.golangci.yml
```diff
@@ -0,0 +1,22 @@
+# golangci-lint configuration used for CI
+run:
+  tests: true
+  timeout: 10m
+  skip-files:
+    - ".*\\.pb\\.go"
```
srikartati, line 6: pb not required I guess?

Do we need to add .csv for antrea registry?

srikartati (commented):
You should add this ci work flow as required github action. You need to get admin rights for the repo.

.golangci.yml
zyiou: Got it. Will remove this line(for skipping `.pb.go` files). it will skip code that is not in Golang so we don't need to add `.csv` here. Thanks!

pkg/exporter/process_test.go
```diff
@@ -31,7 +31,7 @@ func init() {
 
 func TestExportingProcess_SendingTemplateRecordToLocalTCPServer(t *testing.T) {
 	// Create local server for testing
-	listener, err := net.Listen("tcp", ":0")
+	listener, err := net.Listen("tcp", "127.0.0.1:4738")
```
srikartati, line 34: what if the port is already in use? `:0` opens a random available port on localhost. What is the issue with this?

codecov.yaml
```diff
@@ -1,5 +1,5 @@
 codecov:
-  token: ${{secrets.CODECOV_TOKEN}}
+  token: 61ef0651-a3c3-4b71-8ca6-f82630f39579
```
srikartati, line 2: A clarification: does Antrea use the hardcoded token as well?

srikartati:
> You should add this ci work flow as required github action. You need to get admin rights for the repo.

I still see ci checks (integration tests and and golint) are not yet made mandatory. Let me know if you need some info regarding this. We can discuss offline.

zyiou:
> > You should add this ci work flow as required github action. You need to get admin rights for the repo.
> 
> I still see ci checks (integration tests and and golint) are not yet made mandatory. Let me know if you need some info regarding this. We can discuss offline.

I have set integration test and Golangci as required before merging to master or release* branches. Thanks!

codecov.yaml
zyiou: Yes, this is what Antrea does: https://github.com/vmware-tanzu/antrea/blob/561d179f2e497fa834608c28705ac1e870cf3fc7/codecov.yaml#L2.

pkg/exporter/process_test.go
zyiou: It will have [G102](https://securego.io/docs/rules/g102.html) error for `gosec` in Golangci-lint. Used another workaround to solve it.

srikartati (approved):

PR #58, opened 2020-10-27
Change InfoElementLength from list to map

shihhaoli:
This will be less error-prone when adding new data types and
simplify the checking for out-of-range InvalidDataType(255).

This patch also includes some auto golang style changes.


pkg/registry/build_registry/build_registry.go
```diff
@@ -167,10 +167,7 @@ func generateIEString(name string, elementid string, datatype string, enterprise
 	elementID, _ := strconv.ParseUint(elementid, 10, 16)
 	enterpriseID, _ := strconv.ParseUint(enterpriseid, 10, 16)
 	dataType := entities.IENameToType(datatype)
-	length := uint16(0)
-	if entities.IsValidDataType(dataType) {
-		length = entities.InfoElementLength[dataType]
-	}
+	length := entities.InfoElementLength[dataType]
```
srikartati, line 170: If some registry provides an info element with invalidDataType, we will still be adding the infoElement, right? This is not because of this fix. 
@zyiou  Should we allow registries to add info elements with invalidDataType? I feel we should throw an error when building the registry. What do you think?

pkg/registry/build_registry/build_registry.go
```diff
@@ -167,10 +167,7 @@ func generateIEString(name string, elementid string, datatype string, enterprise
 	elementID, _ := strconv.ParseUint(elementid, 10, 16)
 	enterpriseID, _ := strconv.ParseUint(enterpriseid, 10, 16)
 	dataType := entities.IENameToType(datatype)
-	length := uint16(0)
-	if entities.IsValidDataType(dataType) {
```
srikartati, line 171: This "if" check can be removed even if `InfoElementLength` is a list too right as there is an index for `invalidDataType`.

shihhaoli: If InfoElementLength is a list, removing the checking will cause list out-of-range error because InvalidDataType is 255 while list length is only 24.

srikartati: Sure. Misled by the last element in the list which is 0 that has index 23--I think it was intended for InvalidDataType but was not used at all.

Another alternative is to continue with a list and avoid hardcoding InvalidDataType to be 255 and instead have the next iterated value in enum dataTypes. What do you think of this option?

shihhaoli: If InvalidDataType is only used internally, then it is OK.
Personally I prefer using a map because it is easier to insert/delete/update types.
Maintaining the indices and values in two different structures are more error-prone in my experiences.
But this is just my opinion. Thanks.

pkg/registry/build_registry/build_registry.go
zyiou: yes. we should not allow adding info elements with invalidDataType. Will fix it in another patch.

pkg/registry/build_registry/build_registry.go
srikartati: Thanks Shihhao. We add and access the elements in the map/list only once when registering InfoElements from registries and do not modify it at all. The rest of the code uses the length in InfoElement struct directly. In addition, when we restrict adding the elements with invalidDataType to the registry, the usage comes further down.

Either way, it is ok, but because of its minimal usage, we may go with a list instead of a map.

shihhaoli: My concert is more about the readability of the code.
I think keeping keys and values in the same data structure is easier to maintain.
Whoever needs to add/delete/update an entry doesn't need to count the offset in one place and change the value in another place. Does it make sense? Thanks.

srikartati: Sure from a code readability point of view, it is beneficial to have a map. Since it is not very large, it is ok from a memory overhead point of view. We can use a map.

srikartati (approved):
LGTM.

srikartati:
Hi @shihhaoli ,
You probably explored this already. 
I stumbled across the `IEDataType` when working on standalone collector and was thinking of the reason why uint8 used for `IEDataType`. I cannot remember why I chose that route -- maybe because of enthusiasm to explore iota/enumerations as a noob to go programming :) 

I checked the RFC..there is no type suggested and this was not necessary.
We can stick with string type. This way we can simplify the code and delete the unnecessary function `IENameToType`. What do you think?

shihhaoli:
> Hi @shihhaoli ,
> You probably explored this already.
> I stumbled across the `IEDataType` when working on standalone collector and was thinking of the reason why uint8 used for `IEDataType`. I cannot remember why I chose that route -- maybe because of enthusiasm to explore iota/enumerations as a noob to go programming :)
> 
> I checked the RFC..there is no type suggested and this was not necessary.
> We can stick with string type. This way we can simplify the code and delete the unnecessary function `IENameToType`. What do you think?

Yes that's why I asked in a previous comment "BTW, are the data types defined as integers or strings in IPFIX RFCs? "
I thought you wanted to save spaces by passing them as integers in the messages sent to other components?

srikartati:
> Yes that's why I asked in a previous comment "BTW, are the data types defined as integers or strings in IPFIX RFCs? "
> I thought you wanted to save spaces by passing them as integers in the messages sent to other components?

Sorry missed the comment. 
We pass the pointer to `InfoElement` struct whenever possible. However, the size of the allocated memory increase with string type. By considering an increase of 10B (string size) on average for every element and there are ~500 elements. That is around 5KB memory. Smaller struct size also helps from cacheline optimization point of view.

Thanks for raising this. Not sure of my intentions when I wrote this but seems like it does help :) If you have no issues, we can keep this as it is.

shihhaoli:
> > Yes that's why I asked in a previous comment "BTW, are the data types defined as integers or strings in IPFIX RFCs? "
> > I thought you wanted to save spaces by passing them as integers in the messages sent to other components?
> 
> Sorry missed the comment.
> We pass the pointer to `InfoElement` struct whenever possible. However, the size of the allocated memory increase with string type. By considering an increase of 10B (string size) on average for every element and there are ~500 elements. That is around 5KB memory. Smaller struct size also helps from cacheline optimization point of view.
> 
> Thanks for raising this. Not sure of my intentions when I wrote this but seems like it does help :) If you have no issues, we can keep this as it is.

No problem. Yes I am good with current design. Thanks.

PR #59, opened 2020-10-27
Change DateTimeSeconds type to uint32

zyiou:
Patch release for changing `DateTimeSeconds` type to `uint32` according to RFC7011. 
Issue reported in Antrea: https://github.com/vmware-tanzu/antrea/issues/1417


pkg/entities/record.go
```diff
@@ -179,7 +179,14 @@ func (d *dataRecord) AddInfoElement(element *InfoElement, val interface{}) (uint
 		} else {
 			bytesToAppend[0] = 2
 		}
-	case DateTimeSeconds, DateTimeMilliseconds:
+	case DateTimeSeconds:
+		// We expect time to be given in int64 as unix time type in go
+		v, ok := val.(int64)
```
srikartati, line 184: Let's make the user give the seconds in uint32 rather than doing the conversion. Let's make ourselves independent go time package. What do you think?

pkg/entities/record.go
```diff
@@ -179,7 +179,14 @@ func (d *dataRecord) AddInfoElement(element *InfoElement, val interface{}) (uint
...
+		if !ok {
+			return 0, fmt.Errorf("val argument is not of type int64")
+		}
+		binary.BigEndian.PutUint32(bytesToAppend, uint32(v))
+	case DateTimeMilliseconds:
 		// We expect time to be given in int64 as unix time type in go
 		v, ok := val.(int64)
```
srikartati, line 191: Same here. Expect the user to give in uint64.

srikartati (commented):
Let's keep the patch releasechanges separate.

pkg/entities/record.go
zyiou: Makes sense to me. Thanks!

srikartati (approved):
LGTM.

PR #61, opened 2020-10-30
Add testing for IPv6 support

zyiou:
To better support IPv6 for Antrea project, this PR adds some tests to cover possible scenarios flow exporter in Antrea Agent may use.


zyiou:
Close this PR. Will add corresponding tests about IPv6 in next release after #64 is merged.

Issue #15, opened 2020-08-13
Support IPv6 addresses


srikartati:
Please make sure the following encoding function works for IPv6 below:
https://github.com/vmware/go-ipfix/blob/master/pkg/entities/ie.go#L390

srikartati:
More context for this: https://github.com/vmware-tanzu/antrea/pull/1444#discussion_r515045437

Need to fix this in the master branch by v0.3.0 release.

PR #64, opened 2020-10-30
Add encoding support for IPv6Address datatype

srikartati:
In addition, change the type to explicit net.IP when decoding.

Also, made some minor changes in accordance to fmt.Error guidelines.

Fixes: #15


pkg/entities/ie.go
```diff
@@ -399,7 +401,19 @@ func EncodeToIEDataType(dataType IEDataType, val interface{}, buff *bytes.Buffer
 			err := util.Encode(buff, binary.BigEndian, uint32(ipv4Int.Uint64()))
 			return uint32(ipv4Int.Uint64()), err
 		} else {
-			return v, nil
+			return 0, fmt.Errorf("provided IP does not belong to IPv4 address family")
```
zyiou, line 404: Do we need to change the way of decoding IPv4 similar to IPv6: directly encode the result from `To4()` method?

srikartati: That is an extra check if someone gives the wrong IPv6 address with incorrect length. I created separate case blocks for easy code readability.

zyiou: Got that. I mean for line 399-401, maybe we can change to `err := util.Encode(buff, binary.BigEndian, v)` as well. (similar to the way you did for IPv6 to make code more tidy)

srikartati: Yes that can be done. I was under the impression `To4()` may return IPv4 address with a length of 16bytes. Looking at the function closely, it only returns 4 bytes. Thanks for pointing. Updated the patch.

zyiou (approved):
Thanks for updating.

zyiou (approved):

PR #67, opened 2020-11-03
Remove unnecessary testing log in release v0.2.3

zyiou:


srikartati (approved):
LGTM. I saw some fmt.Errof logs starting with caps which is not the convention. We need to address those in a future PR.

PR #68, opened 2020-11-03
Change reverse information element naming

zyiou:
Change reverse ie naming back to the one with `reverse` prefix.
Reason:
Elastic stack collector in Antrea will take reverse ie and ie with the same name (but different enterprise ID) as duplicates can cannot decode the message correctly. 
Use prefix `reverse` but not `reverse_` is also to keep consistent with elastic definition.

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -16,6 +16,8 @@ package registry
 
 import (
 	"fmt"
+	"strings"
+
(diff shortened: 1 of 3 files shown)


srikartati (approved):
LGTM.
Better to change the element name. Felt it was confusing when using reverse elements in Antrea. To avoid that we can have prefix.

PR #71, opened 2020-11-05
Add standalone ipfix collector

srikartati:
This is for Antrea e2e tests.

Binary can be generated through the "collector" target that is defined as part of Makefile.
Tested manually with flow exporter in Antrea. 

Thinking of ways to add tests. All individual components used in go-ipfix/cmd/collector/collector.go are covered through individual unit tests. Need to explore ways to add tests against generated executable.


srikartati:
IPFIX message with data record is printed like following:

I1105 19:30:24.844051   17598 collector.go:79] 
IPFIX-HDR:
  version: 10,  Message Length: 221
  Exported Time: 1604604624 (2020-11-05 19:30:24 +0000 UTC)
  Sequence No.: 6,  Observation Domain ID: 2294339213
DATA SET:
  DATA RECORD-0:
    flowStartSeconds: 1604604619 
    flowEndSeconds: 2288912640 
    sourceIPv4Address: 10.10.2.7 
    destinationIPv4Address: 10.10.2.8 
    sourceTransportPort: 44076 
    destinationTransportPort: 8080 
    protocolIdentifier: 6 
    packetTotalCount: 24 
    octetTotalCount: 128280 
    packetDeltaCount: 0 
    octetDeltaCount: 0 
    reversePacketTotalCount: 24 
    reverseOctetTotalCount: 128280 
    reversePacketDeltaCount: 0 
    reverseOctetDeltaCount: 0 
    sourcePodName: web-client-9f9c7f647-phf46 
    sourcePodNamespace: default 
    sourceNodeName: k8s-node-worker-1 
    destinationPodName: web-server-5c69958856-lqqp9 
    destinationPodNamespace: default 
    destinationNodeName: k8s-node-worker-1 
    destinationClusterIPv4: 0.0.0.0 
    destinationServicePortName:  
    ingressNetworkPolicyName:  
    ingressNetworkPolicyNamespace:  
    egressNetworkPolicyName:  
    egressNetworkPolicyNamespace:

cmd/collector/collector.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func newCollectorCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:  "ipfix-collector",
+		Long: "IPFIX collector to decode the exported flow records",
+		Run: func(cmd *cobra.Command, args []string) {
+			initLoggingToFile(cmd.Flags())
+			if err := run(); err != nil {
+				klog.Fatalf("Error running agent: %v", err)
```
zyiou, line 158: Message should be `Error running collector: %v` here

cmd/collector/collector.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return err
+	}
+	// Start listening to connections and receiving messages.
+	messageReceived := make(chan *entities.Message)
+	go func() {
+		go cp.Start()
+		msgCount := 0
+		// This can be simplified if collector uses a message channel.
+		// Are there any use cases for having list of messages in collector process.
```
zyiou, line 125: I'm a little confused about this comment. Do you mean there are still use cases of having a list?

cmd/collector/collector.go
```diff
@@ -0,0 +1,178 @@
...
+	klog.Infof(buf.String())
+}
+
+func signalHandler(signalCh chan os.Signal, messageReceived chan *entities.Message) {
+	for {
+		select {
+		case msg := <-messageReceived:
+			printIPFIXMessage(msg)
+		case <-signalCh:
+			break
```
zyiou, line 88: Wondering whether it is better to use `return` here since `break` will only break the select statement.

zyiou (commented):
Thanks for the PR. Tested with Antrea and it works well for me too. Only have some trivial comments.

cmd/collector/collector.go
srikartati: Yes :)

cmd/collector/collector.go
srikartati: Yeah. As I was using signalCh directly in main.. it was working fine. `return` should be the proper way. Added a stop channel to use it in main.

cmd/collector/collector.go
zyiou: Got it. I will modify it in intermediate process PR.

zyiou (approved):

cmd/collector/collector.go
srikartati: Thanks.

PR #78, opened 2020-11-13
Add docker image for ipfix-collector

srikartati:
Also add github workflow to push to Antrea org of docker registry.
Image is pushed as antrea/ipfix-collector with latest (for master branch) or release tag (for releases).

### DIFF
--- a/.github/workflows/build_tag.yml
+++ b/.github/workflows/build_tag.yml
@@ -1,0 +1,22 @@
+name: Build and push a release image
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  build:
+    runs-on: [ubuntu-18.04]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build go-ipfix Docker images and push to Antrea Docker registry
+        env:
+          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+          TAG: ${{ github.ref }}
+        run: |
+          make docker-collector
+          echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+          docker tag antrea/ipfix-collector antrea/ipfix-collector:"${TAG:10}"
+          docker push antrea/ipfix-collector:"${TAG:10}"
--- a/build/images/Dockerfile.build.collector
+++ b/build/images/Dockerfile.build.collector
@@ -1,0 +1,19 @@
+FROM golang:1.15 as go-ipfix-build
+
+WORKDIR /go-ipfix
+
+COPY go.mod /go-ipfix/go.mod
+
+RUN go mod download
+
+COPY . /go-ipfix
+
+RUN make collector
+
+FROM ubuntu:18.04
+
+LABEL maintainer="go-ipfix"
+LABEL description="A Docker image based on Ubuntu 18.04 which contains a IPFIX collector"
+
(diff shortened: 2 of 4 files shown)


zyiou:
Thanks for the PR. 
I saw that the image was pushed automatically to the Antrea docker hub. But the image does not work well for antrea e2e tests. I tried to use the dockerfile from this PR to build image and push to my personal docker hub it works with no problem. The reason may be the [image](https://hub.docker.com/layers/antrea/ipfix-collector/v0.1.4/images/sha256-0604fab2fdb55e4427e0734b304ee90e4315c8893db6a93cfb381c6ef33827b4?context=explore) was not loaded correctly: `ENTRYPOINT` is not loaded in image layers. Do you know why it is happening?

srikartati:
> Thanks for the PR.
> I saw that the image was pushed automatically to the Antrea docker hub. But the image does not work well for antrea e2e tests. I tried to use the dockerfile from this PR to build image and push to my personal docker hub it works with no problem. The reason may be the [image](https://hub.docker.com/layers/antrea/ipfix-collector/v0.1.4/images/sha256-0604fab2fdb55e4427e0734b304ee90e4315c8893db6a93cfb381c6ef33827b4?context=explore) was not loaded correctly: `ENTRYPOINT` is not loaded in image layers. Do you know why it is happening?

Yes, that was a test image. I forgot the `ENTRYPOINT` point before. Now added. Once I push this PR, it will push antrea/ipfix-collector:latest. You could use that for testing. For Antrea code, I think we need the image antrea/ipfix-collector:v0.3.1. We will get this after doing the corresponding patch release on branch release-0.3.0

build/images/Dockerfile.build.collector
```diff
@@ -0,0 +1,22 @@
...
+RUN go mod download
+
+COPY . /go-ipfix
+
+RUN make collector
+
+
+FROM antrea/base-ubuntu:2.14.0
+
+FROM ubuntu:18.04
```
zyiou, line 16: Just curious about why we want to use both `antrea/base-ubuntu` and `ubuntu`

.github/workflows/build_tag.yml
```diff
@@ -0,0 +1,22 @@
...
+      - name: Build go-ipfix Docker images and push to Antrea Docker registry
+        env:
+          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+          TAG: ${{ github.ref }}
+        run: |
+          make docker-collector
+          echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+          docker tag antrea/ipfix-collector antrea/ipfix-collector:"${TAG:10}"
+          docker push antrea/ipfix-collector:"${TAG:10}"
```
zyiou, line 22: where do we define this ${TAG:10}?

srikartati: TAG is an environmental variable that gets ref tags in L17. Ref tags look like .git/refs/v0.*. TAG:10 ignores the first 10 characters.

build/images/Dockerfile.build.collector
srikartati: Thank for the question. We do not need antrea base-ubuntu image.

.github/workflows/build_tag.yml
zyiou: Got it. Thanks for explaining.

zyiou (commented):
The PR looks good to me. Please remove the line using the `antrea/base-ubuntu` image then it can be merged.

zyiou (approved):

PR #81, opened 2020-11-18
Improve intermediate process

zyiou:
based on after-merge comments on PR #51

### DIFF
--- a/pkg/intermediate/aggregate.go
+++ b/pkg/intermediate/aggregate.go
@@ -3,19 +3,20 @@ package intermediate
 import (
 	"fmt"
 	"net"
+	"strings"
 	"sync"
+
+	"k8s.io/klog"
 
 	"github.com/vmware/go-ipfix/pkg/entities"
 	"github.com/vmware/go-ipfix/pkg/registry"
 )
 
 type AggregationProcess struct {
-	// tupleRecordMap maps each connection with its records
-	tupleRecordMap map[Tuple][]entities.Record
-	// tupleRecordLock allows multiple readers or one writer at the same time
-	tupleRecordLock sync.RWMutex
-	// workerPool is for storing worker channels to process the messages
-	workerPool chan chan *entities.Message
+	// flowKeyRecordMap maps each connection (5-tuple) with its records
+	flowKeyRecordMap map[FlowKey][]entities.Record
+	// flowKeyRecordLock allows multiple readers or one writer at the same time
+	flowKeyRecordLock sync.RWMutex
 	// messageChan is the channel to receive the message
 	messageChan chan *entities.Message
 	// workerNum is the number of workers to process the messages
@@ -24,9 +25,11 @@ type AggregationProcess struct {
 	workerList []*worker
 	// correlateFields are the fields to be filled in correlating process
 	correlateFields []string
+	// stopChan is the channel to receive stop message
+	stopChan chan bool
 }
 
-type Tuple struct {
+type FlowKey struct {
 	SourceAddress      string
 	DestinationAddress string
 	Protocol           uint8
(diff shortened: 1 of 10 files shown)


pkg/intermediate/aggregate.go
```diff
@@ -144,92 +144,83 @@ func (a *aggregation) removeDuplicates(tuple Tuple) {
...
+
+	dstPortElement, exist := record.GetInfoElementWithValue("destinationTransportPort")
+	if !exist {
+		return FlowKey{}, fmt.Errorf("destinationTransportPort does not exist")
+	}
+	dstPort, ok := dstPortElement.Value.(uint16)
+	if !ok {
+		return FlowKey{}, fmt.Errorf("destinationTransportPort is not in correct format")
+	}
+
```
srikartati, line 189: Can these lines be simplified for code (including till L221) readability purposes? 
Define the flow key elements in string list, and then iterate over them using switch-case?
There is common code for (srcPort, dstPort) and (srcIP, dstIP). This decreases the lines of code.
What do you think?

pkg/intermediate/aggregate_test.go
```diff
@@ -211,10 +211,10 @@ func TestAggregationProcess(t *testing.T) {
 		aggregationProcess.Stop()
 	}()
 	aggregationProcess.Start()
```
srikartati, line 213: Maybe a comment here saying that `Start` of `aggregationProcess` is blocked until the above go routine with `Stop` finishes is better to have. This will give an idea of their usage.
Proper usage would be having `Start` in a go-routine with some external channel. And until the external channel is closed, `Stop` won't be called.

pkg/intermediate/aggregate.go
```diff
@@ -144,92 +144,83 @@ func (a *aggregation) removeDuplicates(tuple Tuple) {
...
-		srcPortNum, ok := infoElementMap["sourceTransportPort"].Value.(uint16)
-		if !ok {
-			return Tuple{}, fmt.Errorf("sourceTransportPort is not in correct format.")
-		}
-		srcPort = srcPortNum
-		dstPortNum, ok := infoElementMap["destinationTransportPort"].Value.(uint16)
-		if !ok {
-			return Tuple{}, fmt.Errorf("destinationTransportPort is not in correct format.")
+// getFlowKeyFromRecord returns 5-tuple from data record
+func getFlowKeyFromRecord(record entities.Record) (FlowKey, error) {
```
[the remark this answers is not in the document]
zyiou: Sure. Updated.

pkg/intermediate/aggregate.go
zyiou: Agree. Changed to switch-case. Thanks!

pkg/intermediate/aggregate.go
```diff
@@ -124,16 +130,14 @@ func (a *aggregation) correlateRecords(flowKey FlowKey, record entities.Record)
...
-	a.FlowKeyRecordLock.Unlock()
 	a.addRecordToMap(flowKey, record)
 	a.removeDuplicates(flowKey)
 }
 
 func (a *aggregation) removeDuplicates(flowKey FlowKey) {
-	a.FlowKeyRecordLock.Lock()
-	defer a.FlowKeyRecordLock.Unlock()
-	records := a.flowKeyRecordMap[flowKey]
+	records := a.GetFlowKeyRecordMap()[flowKey]
```
shihhaoli, line 140: This is still inside a lib function (i.e. not an external caller).
Do we need to make a copy of the map?
Can we just lock/unlock it?

pkg/intermediate/aggregate.go
```diff
@@ -79,24 +80,29 @@ func (a *aggregation) AggregateMsgByFlowKey(message *entities.Message) error {
...
+		a.correlateRecords(*flowKey, record)
 	}
 	return nil
 }
 
 func (a *aggregation) GetFlowKeyRecordMap() map[FlowKey][]entities.Record {
-	return a.flowKeyRecordMap
+	a.flowKeyRecordLock.RLock()
+	defer a.flowKeyRecordLock.RUnlock()
+	resMap := make(map[FlowKey][]entities.Record)
```
shihhaoli, line 91: We may consider returning a pointer of the new map to avoid making a second copy during returned-by-value?

pkg/intermediate/aggregate.go
zyiou: Sure. Updated. Thanks!

pkg/intermediate/aggregate.go
zyiou: [Map](https://blog.golang.org/maps) in Golang is reference type like pointer, so it should be returned by reference.

shihhaoli: Cool, nice to learn that. Thanks.

srikartati: The main objective of this function is to iterate over the records and perform a common operation.
Could we do something like below in Antrea to avoid the step of creating a copy of the map? I think this map will be common to the cluster w.r.t flow aggregator implementation and could be very large; making a copy can involve high overhead.

https://github.com/vmware-tanzu/antrea/blob/master/pkg/agent/flowexporter/connections/connections.go#L209

srikartati: I see the above function is currently utilized only in tests, and all the others are internal functions that access the map directly. However, exposing this function may lead to inefficient usage.

zyiou: Thanks for the suggestion. For flow aggregator, we will need to get the map, obtain the records and send them, that is the use case for exposing the function.

pkg/intermediate/aggregate.go
```diff
@@ -168,59 +171,65 @@ func isRecordFromSrc(record entities.Record) bool {
...
+			}
+			if name == "sourceTransportPort" {
+				srcPort = port
+			} else {
+				dstPort = port
+			}
+		case "sourceIPv4Address", "destinationIPv4Address":
+			element, exist := record.GetInfoElementWithValue(name)
+			if !exist {
+				// if sourceIPv4Address/destinationIPv4Address does not exist, check sourceIPv6Address/destinationIPv6Address
```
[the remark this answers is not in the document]
zyiou: I was thinking separate sourceIPv6Address and destinationIPv6Address would cause duplications. But it is clearer. Thanks!

pkg/intermediate/aggregate.go
zyiou: Checked with @dreamtalen that passing a callback function will be enough for flow aggregator to process the records.

srikartati (commented):
LGTM. 
Please add some comments for methods in the Aggregation data structure. Especially where lock needs to be acquired by the caller. 
Also, changing name from aggregation->aggregationProcess is better? We can be consistent with exportingProcess and collectionProcess.

zyiou:
> Please add some comments for methods in the Aggregation data structure. Especially where lock needs to be acquired by the caller.

Currently the lock will not be acquired publicly. Which part are you referring to?
> Also, changing name from aggregation->aggregationProcess is better? We can be consistent with exportingProcess and collectionProcess.

Done. Thanks!

srikartati:
> > Please add some comments for methods in the Aggregation data structure. Especially where lock needs to be acquired by the caller.
> 
> Currently the lock will not be acquired publicly. Which part are you referring to?

I was referring to `addRecordToMap`. Yes, it is an internal method, but the add function of the map is usually protected with a lock. If someone wants to extend the aggregateProcess structure in future, comments will be useful.

zyiou:
> I was referring to `addRecordToMap`. Yes, it is an internal method, but the add function of the map is usually protected with a lock. If someone wants to extend the aggregateProcess structure in future, comments will be useful.

Got it. Thanks!

srikartati (dismissed):
Thanks for working on this.
LGTM other than comments for methods. 
Please merge once you get approval from Shihao.

pkg/intermediate/aggregate.go
```diff
@@ -222,20 +223,22 @@ func getFlowKeyFromRecord(record entities.Record) (*FlowKey, error) {
...
+			if isSrcFilled && isDstFilled {
+				break checkFlowKey
+			}
 		case "sourceIPv6Address", "destinationIPv6Address":
-			element, exist := record.GetInfoElementWithValue(name)
-			if isIPv4 {
-				if exist {
-					klog.Warning("two ip version (ipv4 & ipv6) is not supported at the same time in Flow Key.")
-				}
+			if (isSrcFilled && strings.Contains(name, "source")) || (isDstFilled && strings.Contains(name, "destination")) {
```
shihhaoli, line 238: Do we want to print warning if both (src ipv4 and src ipv6) or (dst ipv4 and dst ipv6) are set?

zyiou: Sure. Updated

shihhaoli (approved):
Thanks for the changes. LGTM.

PR #82, opened 2020-11-18
Keep go.mod pkg version same as Antrea

zyiou:
Upgrading Antrea to go-ipfix v0.3.0 causes conflicts when run makefile. Changed corresponding pkg to keep consistent with Antrea.


srikartati (approved):
LGTM. 
Curious what happens in Antrea if we upgrade cobra pkg?
Understand the issue with K8s..

zyiou:
> Curious what happens in Antrea if we upgrade cobra pkg?

It will not cause any problems. I changed the version just to keep consistent with Antrea one. Do we need to use the upgraded version?

srikartati:
> > Curious what happens in Antrea if we upgrade cobra pkg?
> 
> It will not cause any problems. I changed the version just to keep consistent with Antrea one. Do we need to use the upgraded version?

If there is no requirement to move to an upgraded version as in no new features are required, we can carry on with the old version.

PR #83, opened 2020-11-18
Simplify standalone collector code

srikartati:
Modify standalone collector code to make it extendable in the future.

### DIFF
--- a/cmd/collector/collector.go
+++ b/cmd/collector/collector.go
@@ -60,6 +60,25 @@ func initLoggingToFile(fs *pflag.FlagSet) {
 	}
 }
 
+func signalHandler(stopCh chan struct{}) {
+	// We define a channel with buffer of two signals. The first signal that is
+	// received will cause the stopCh channel to be closed, giving the opportunity
+	// to the program to exit gracefully. If a second signal is received before
+	// then, we will force exit with code 1.
+	signalCh := make(chan os.Signal, 2)
+
+	go func() {
+		<-signalCh
+		close(stopCh)
+		<-signalCh
+		klog.Warning("Received second signal, will force exit")
+		logs.FlushLogs()
+		os.Exit(1)
+	}()
+
+	signal.Notify(signalCh, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+}
+
 func addIPFIXFlags(fs *pflag.FlagSet) {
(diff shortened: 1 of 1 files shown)


zyiou (dismissed):
LGTM

cmd/collector/collector.go
```diff
@@ -60,6 +60,19 @@ func initLoggingToFile(fs *pflag.FlagSet) {
...
+	signal.Notify(signalCh, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+
+	for {
+		select {
+		case <-signalCh:
+			close(stopCh)
+			return
+		}
+	}
+}
```
shihhaoli, line 74: Just curious, if we have only one case, do we still need to use select?
For example, does the following suit our need?

```
func signalHandler(stopCh chan struct{}) {
	signalCh := make(chan os.Signal, 1)
	signal.Notify(signalCh, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
	<-signalCh
	close(stopCh)
}
```
Thanks.

srikartati: Yes, you are right. We can simplify the code.
Taking inspiration from Antrea applications, I defined signalCh with capacity 2 to handle the case if a second interrupt/termination comes through. In that case, program exits forcefully.

srikartati:
> So if the second signal was not received, this function will not exit normally?

I think you meant the go routine; the function returns immediately after starting the go routine. Yes that go routine abruptly stops when the main exits gracefully. Main does not need to wait for this signal handler go routine to finish.

cmd/collector/collector.go
shihhaoli: Yes, I meant the go routine. Thanks for the explanation.

cmd/collector/collector.go
```diff
@@ -60,6 +60,26 @@ func initLoggingToFile(fs *pflag.FlagSet) {
...
+		<-signalCh
+		close(stopCh)
+		<-signalCh
+		klog.Warning("Received second signal, will force exit")
+		logs.FlushLogs()
+		os.Exit(1)
+	}()
+
+	signal.Notify(signalCh, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+	return
```
shihhaoli, line 80: Looks like the return is not needed.

shihhaoli (approved):
LGTM.

shihhaoli (approved):

PR #84, opened 2020-11-20
Expose collecting process and aggregation process as public struct

zyiou:
so that it can be packed as fields in IPFIXCollectingProcess/IPFIXAggregationProcess in flow aggregator.


srikartati (approved):
looks good

Issue #54, opened 2020-10-22
Documentation for go-ipfix library


srikartati:
- [] Documentation for abstraction of pkg/entities or in general code structure.
- [] Documentation for exporter
- [] Documentation for collector
- [] Documentation for intermediate process

PR #88, opened 2020-11-24
Add Kafka producer

### LINKED ISSUE #66: Implement Kafka Exporter (Producer)
Given Kafka broker system, Kafka exporter/producer can export IPFIX flow records in the format of Kafka topics. 

- Given there is Kafka Broker and Consumers, we can create a Kafka Producer to convert IPFIX flow records into Kafka topics and send them to the consumer.
- Communication in the Kafka broker system is through the pub-sub mechanism. The underlying communication protocol is binary protocol.
- Encoding Kafka messages through protobuf flows seems to be the best practice.
- Convert IPFIX flow records to records in protobuf format and convert them to Kafka topics. 
- Get the protobuf format/schema from protobuf schema registry (Example: https://github.com/riferrei/srclient). Other option is have a abstraction for protobuf schema to plug in any desired schema for the coversion of flow records in to kafka topics.

srikartati:
Convert IPFIX Message from the collector process or intermediate process
to flow messages (proto schema); each flow record corresponds to one
flow message. Eventually, encode them into kafka topic and pass them
to the input channel of asyncProducer.
AsyncProducer is created given the kafka brokers and other config paramters.
For this purpose, we utilized go kafka client library from sarama.

Fixes #66


pkg/producer/kafka.go
```diff
@@ -0,0 +1,182 @@
...
+				flowMsg.SrcNodeName = ie.Value.(string)
+			case "destinationPodNamespace":
+				flowMsg.DstPodNamespace = ie.Value.(string)
+			case "destinationPodName":
+				flowMsg.DstPodName = ie.Value.(string)
+			case "destinationNodeName":
+				flowMsg.DstNodeName = ie.Value.(string)
+			case "destinationClusterIPv4", "destinationClusterIPv6":
+				if flowMsg.DstClusterIP != "" {
+					klog.Warningf("Do not expect destination IP: %v to be filled already", flowMsg.DstClusterIP)
```
shihhaoli, line 102: destination IP -> destination cluster IP, to distinguish from line 63?

pkg/producer/kafka.go
```diff
@@ -0,0 +1,182 @@
...
+
+	return producer, nil
+}
+
+// SendFlowMessage takes in the flow message in proto schema, encodes it and sends
+// it to on the producer channel.
+func (kp *KafkaProducer) SendFlowMessage(msg *protobuf.FlowMessage) {
+	buf := proto.NewBuffer([]byte{})
+	if err := buf.EncodeMessage(msg); err != nil {
+		klog.Errorf("Error when encoding flow message: %v", err)
```
shihhaoli, line 163: Should we return here whe error happened?

pkg/producer/kafka_test.go
```diff
@@ -0,0 +1,248 @@
...
+		ExportTime:    0,
+		ExportAddress: "127.0.0.1",
+		Set:           set,
+	}
+}
+
+func TestKafkaProducer_Publish(t *testing.T) {
+	kafkaConfig := sarama.NewConfig()
+	kafkaConfig.Version = KafkaConfigVersion
+	kafkaConfig.Producer.Return.Successes = true
```
zyiou, line 220: What is this variable for? passing the message when producer succeeded?

pkg/producer/protobuf/flow.proto
```diff
@@ -0,0 +1,65 @@
...
+
+syntax = "proto3";
+
+package github.com.vmware.goipfix.producer.protobuf;
+
+option go_package = "pkg/producer/protobuf";
+
+message FlowMessage {
+  // Header of IPFIX Message.
+  uint32 TimeReceived = 1;
```
zyiou, line 23: This file is for defining the Proto format, right? noticed that the number assigned is not in sequence order. how are the numbers defined?

pkg/test/collector_producer_test.go
```diff
@@ -0,0 +1,99 @@
...
+	kafkaConfig := sarama.NewConfig()
+	kafkaConfig.Version = producer.KafkaConfigVersion
+	kafkaConfig.Producer.Return.Successes = true
+	kafkaConfig.Producer.Return.Errors = true
+
+	mockProducer := saramamock.NewAsyncProducer(t, kafkaConfig)
+	kafkaProducer := producer.NewKafkaProducer(mockProducer, "test-flow-msgs")
+
+	go func() {
+		time.Sleep(time.Second)
```
zyiou, line 64: We can use similar to latest code without sleep for this.

pkg/producer/kafka.go
```diff
@@ -0,0 +1,182 @@
...
+	producer sarama.AsyncProducer
+	topic    string
+}
+
+// convertIPFIXMsgToFlowMsgs converts data records in IPFIX message to flow messages
+// in given proto schema.
+func convertIPFIXMsgToFlowMsgs(msg *entities.Message) []*protobuf.FlowMessage {
+	if msg.Set.GetSetType() == entities.Template {
+		return nil
+	}
```
zyiou, line 41: we will not need to send flow message to Kafka consumer for it to decode right?

zyiou (commented):
The PR looks good. Only have some questions about the concepts. Thanks!

pkg/producer/protobuf/flow.proto
srikartati: Yes these numbers signify the order of encoding of fields in Kafka message. Numbers did not follow a sequence because I added some fields later.

pkg/test/collector_producer_test.go
srikartati: done

pkg/producer/kafka_test.go
srikartati: Yes whenever message sending is successful, we get a return success message. This is used in the mock in the test.

pkg/producer/kafka.go
srikartati: Yes, need to return. Thanks for catching. Done.

pkg/producer/kafka.go
srikartati: Yes template functionality is already achieved through the defined protobuf schema.

shihhaoli (dismissed):
LGTM

zyiou (dismissed):
LGTM

srikartati:
Squashed the commits. Requesting review again.

zyiou (approved):
LGTM. we can consider to add doc for kafka later.

srikartati:
> LGTM. we can consider to add doc for kafka later.

We need some documentation in IPFIX too :) Tracking an Issue #54 here. Extending that for Kafka producer too.

Issue #90, opened 2020-11-26
Failures when the unit tests are run with race flag


srikartati:
Changed the test-unit target in Makefile to run with race flag: "$(GO) test -race ./..."

There are quite a few data race failures!! Logs are below.

I1126 06:25:16.101098   63379 udp.go:80] Collecting process from 127.0.0.1:52940 has stopped.
I1126 06:25:18.099195   63379 process.go:262] Template with id 256, and obsDomainID 1 is expired.

WARNING: DATA RACE
Write at 0x00c000115c20 by goroutine 63:
  runtime.mapdelete()
      /usr/local/go/src/runtime/map.go:685 +0x0
  github.com/vmware/go-ipfix/pkg/collector.(*CollectingProcess).deleteTemplate()
      /Users/stati/work/antrea_all/go-ipfix/pkg/collector/process.go:282 +0x138
  github.com/vmware/go-ipfix/pkg/collector.(*CollectingProcess).addTemplate.func1()
      /Users/stati/work/antrea_all/go-ipfix/pkg/collector/process.go:263 +0x214

Previous read at 0x00c000115c20 by goroutine 57:
  runtime.mapaccess1()
      /usr/local/go/src/runtime/map.go:394 +0x0
  github.com/vmware/go-ipfix/pkg/collector.TestUDPCollectingProcess_TemplateExpire()
      /Users/stati/work/antrea_all/go-ipfix/pkg/collector/process_test.go:367 +0x2d0
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1123 +0x202

Goroutine 63 (running) created at:
  github.com/vmware/go-ipfix/pkg/collector.(*CollectingProcess).addTemplate()
      /Users/stati/work/antrea_all/go-ipfix/pkg/collector/process.go:257 +0x44e
  github.com/vmware/go-ipfix/pkg/collector.(*CollectingProcess).decodeTemplateSet()
      /Users/stati/work/antrea_all/go-ipfix/pkg/collector/process.go:207 +0xc11
  github.com/vmware/go-ipfix/pkg/collector.(*CollectingProcess).decodePacket()
      /Users/stati/work/antrea_all/go-ipfix/pkg/collector/process.go:130 +0x404
  github.com/vmware/go-ipfix/pkg/collector.(*CollectingProcess).handleUDPClient.func1()
      /Users/stati/work/antrea_all/go-ipfix/pkg/collector/udp.go:88 +0x2c7

[...]

PR #91, opened 2020-11-28
Fix the unit tests of pkg/exporter and its cleanup

srikartati:
- Merge MsgBuffer and Message structs to use common struct
  for encoding and decoding. In addition, make fields in Message
  struct private.
- Fix the unit tests in pkg/exporter to resolve the data race
  errors (race flag)
- Cleanup FinishSet.
- Add IPv6 exporter support for addOriginalExporterInfo function
  in pkg/intermediate.
- Other miscellaneous stuff.

Partially fixes: #90


pkg/entities/message.go
```diff
@@ -14,45 +14,175 @@
...
+	buffer        *bytes.Buffer
+	version       uint16
+	length        uint16
+	seqNumber     uint32
+	obsDomainID   uint32
+	exportTime    uint32
+	exportAddress string
+	isDecoding    bool
+	set           Set
+	mutex         sync.Mutex
```
zyiou, line 37: nit: Will it be better to use RWMutex to unblocking multiple reads?

zyiou (commented):
The PR overall LGTM. I can see that the race condition happens mainly due to multiple access to message struct. Is it possible to also happen in set struct?

pkg/entities/message.go
srikartati: There are not that many routines doing reads in parallel with the message object. The main reason for data race is because of goroutine that does sendRefreshedTemplates in the case of UDP. In specific, message object in exporting process is shared.

srikartati:
> The PR overall LGTM. I can see that the race condition happens mainly due to multiple access to message struct. Is it possible to also happen in set struct?

Set in most usages looks like a local object in a given routine. In many methods, we create the set and use it locally and do not share it with other routines. If you see any scenario where set object can be shared, let me know.

pkg/entities/message.go
zyiou: Got it. Thanks for clarifying

zyiou (approved):

srikartati:
Thanks for the review @zyiou 

After discussing with you, I thought about the message field in the exporting process struct and reusing the message object. 
I did not feel that is necessary, so I removed it and let the method AddSetAndSendMsg create the message object for us. 
This made sense to me. Even with multiple sets, we can do something like AddSetsAndSendMsg and create a separate message object in the method. 

As a consequence, we can remove the lock in Message struct. In general, one message object should only belong to one go routine, it cannot be shared by multiple threads/go routines. Let me know what you think.

Sorry this may lead to some more work in the followup PRs that are dependent on this.

pkg/exporter/process.go
```diff
@@ -164,26 +165,22 @@ func (ep *ExportingProcess) NewTemplateID() uint16 {
 	return ep.templateID
 }
 
-func (ep *ExportingProcess) sendMsg(set entities.Set) (int, error) {
+func (ep *ExportingProcess) sendMsg(msg *entities.Message, set entities.Set) (int, error) {
```
zyiou, line 168: `sendMsg` and `sendSet` is kinda dependent/overlapping on each other. In `sendSet` there is creation of new message but not filling it completely. `sendMsg` will take the incomplete message and fill it. Will it be better to only checking the set in `sendSet` and call function like `createMsgAndSend(set)`. in `createMsgAndSend(set)` it will create new message, add set(s) into it and send it. What do you think?

zyiou (commented):
Current solution makes sense to me. It is much clearer and can avoid multiple access to message. Thanks!

pkg/exporter/process.go
srikartati: Good idea. Initially, I thought about that but felt the logic in`SendSet` will become very lean and almost like sanity check of the set. I agree that it is a cleaner way to do that and can be extended easily for multiple sets. Modified it.
In addition, I optimized the function a bit more to avoid writing the set buffer to message buffer. I append the slices and send it on the exporter connection.

zyiou (approved):
Thanks for the change. LGTM

PR #92, opened 2020-11-29
Message size check for UDP transport

srikartati:
Added a path MTU config parameter for InitExportingProcess function.
For UDP transport, based on path MTU parameter, we make a decision on
the message size. This is optional parameter for TCP transport; TCP
supports max socket buffer size (65535).

Cleaned up AddSetAndSendMsg function a bit as well.

This PR is on top of PR #91. It will be checked in after that PR. Please look for separate commit when reviewing.

### DIFF
--- a/pkg/exporter/process.go
+++ b/pkg/exporter/process.go
@@ -43,16 +43,22 @@ type ExportingProcess struct {
 	obsDomainID     uint32
 	seqNumber       uint32
 	templateID      uint16
+	pathMTU         int
 	templatesMap    map[uint16]templateValue
 	templateRefCh   chan struct{}
 	mutex           sync.Mutex
 }
 
-// InitExportingProcess takes in collector address(net.Addr format), obsID(observation ID) and tempRefTimeout
-// (template refresh timeout). tempRefTimeout is applicable only for collectors listening over UDP; unit is seconds. For TCP, you can
-// pass any value. For UDP, if 0 is passed, consider 1800s as default.
-// TODO: Get obsID, tempRefTimeout as args which can be of dynamic size supporting both TCP and UDP.
-func InitExportingProcess(collectorAddr net.Addr, obsID uint32, tempRefTimeout uint32) (*ExportingProcess, error) {
+// InitExportingProcess takes in collector address(net.Addr format), obsID(observation ID)
+// and tempRefTimeout(template refresh timeout). tempRefTimeout is applicable only
+// for collectors listening over UDP; unit is seconds. For TCP, you can pass any
+// value. For UDP, if 0 is passed, consider 1800s as default.
+//
+// PathMTU is recommended for UDP transport. If not given a valid value, i.e., either
+// 0 or a value more than 1500, we consider a default value of 512B as per RFC7011.
+// PathMTU is optional for TCP as we use max socket buffer size of 65535. It can
+// be provided as 0.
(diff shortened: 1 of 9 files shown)


pkg/exporter/process.go
```diff
@@ -149,11 +160,17 @@ func (ep *ExportingProcess) createAndSendMsg(set entities.Set) (int, error) {
...
-	// Check if message is exceeding the limit with new set
+	// Check if message is exceeding the limit after adding the set. Include message
+	// header length too.
 	msgLen := uint16(msg.GetMsgBufferLen()) + set.GetBuffLen()
-	// TODO: Change the limit for UDP transport. This is only valid for TCP transport.
-	if msgLen > entities.MaxTcpSocketMsgSize {
-		return 0, fmt.Errorf("set size exceeds max socket size")
+	if ep.connToCollector.LocalAddr().Network() == "tcp" {
+		if msgLen > entities.MaxTcpSocketMsgSize {
+			return 0, fmt.Errorf("TCP transport: set size exceeds max socket buffer size")
```
zyiou, line 168: nit: s/set/message

pkg/exporter/process.go
```diff
@@ -149,11 +160,17 @@ func (ep *ExportingProcess) createAndSendMsg(set entities.Set) (int, error) {
...
+		}
+	} else {
+		if msgLen > ep.pathMTU {
+			return 0, fmt.Errorf("UDP transport: set size exceeds max pathMTU (set as %v)", ep.pathMTU)
```
zyiou, line 172: same as above.

zyiou (dismissed):
LGTM

pkg/exporter/process_test.go
```diff
@@ -247,13 +247,21 @@ func TestExportingProcess_SendingDataRecordToLocalTCPServer(t *testing.T) {
...
 	assert.Equal(t, uint32(1), exporter.seqNumber)
+
+	// Create data set with multiple data records to test invalid message length
+	// logic for TCP transport.
+	dataSet = entities.NewSet(entities.Data, templateID, false)
+	for i := 0; i < 10000; i++ {
+		dataSet.AddRecord(elements, templateID)
+	}
+	bytesSent, err = exporter.SendSet(dataSet)
+	assert.Error(t, err)
```
zyiou, line 263: I tried to log the error here. the error message is `could not send the complete message on the connection` instead of exceeding maximum size. Not sure whether we can test it here since 65535 is rather large and current msgLen for 10000 records is only 14484

pkg/exporter/process_test.go
```diff
@@ -328,12 +336,20 @@ func TestExportingProcess_SendingDataRecordToLocalUDPServer(t *testing.T) {
...
-		t.Fatalf("Got error when sending record: %v", err)
-	}
+	assert.NoError(t, err)
 	// 28 is the size of the IPFIX message including all headers (20 bytes)
 	assert.Equal(t, 28, bytesSent)
 	assert.Equal(t, dataRecBytes, <-buffCh)
 	assert.Equal(t, uint32(1), exporter.seqNumber)
+
+	// Create data set with multiple data records to test invalid message length
+	// logic for
```
zyiou, line 346: incomplete comment here.

pkg/exporter/process_test.go
srikartati: Good catch with the error. Each data record is around ~8 bytes, so the number of data records added is fine but uint16 casting in `set.GetBufferLen` is causing the issue. Fixed it.

zyiou (approved):
LGTM

PR #99, opened 2020-12-03
Modify correlation and add stats support in aggregation process

srikartati:
- Create AggregationFlowRecord structure to store required metadata
- Simplify the correlation process, so that we just store one record
- Add stats support in Aggregation process.
- There are issues with the present code for intra-node flows that do not need correlation. This PR fixes that.
- Add tests for intra-node flows and stats support.


srikartati:
@zyiou
I modified the correlation part of the code to make it slightly simpler and to maintain some metadata. 
I added some comments where stats can updated and maintained. PTAL.
Since you started working on the stats part, want to make sure the approach is as expected.
I can fill up the stats and timestamp code too, if you haven't made too much progress on that front.

srikartati:
Please consider this as a priority in reviewing. It is an important PR for the flow aggregator process.

pkg/intermediate/aggregate.go
```diff
@@ -118,74 +130,150 @@ func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
...
+				if ieWithValue.Value == 0 {
+					existingIeWithValue, _ := existingRecord.GetInfoElementWithValue(field)
+					if existingIeWithValue.Value != 0 {
+						klog.Warningf("This field with name %v should not have been filled with value %v in existing record.", field, existingIeWithValue.Value)
+					}
+					existingIeWithValue.Value = ieWithValue.Value
+				}
+			case entities.Ipv4Address:
+				ipInString := ieWithValue.Value.(net.IP).To4().String()
+				if ipInString == "0. 0. 0. 0" {
```
zyiou, line 196: And `==` is supposed to be `!=`, right?

pkg/intermediate/aggregate_test.go
```diff
@@ -112,8 +117,8 @@ func createMsgwithDataSet2() *entities.Message {
 	ie5 := entities.NewInfoElementWithValue(entities.NewInfoElement("protocolIdentifier", 4, 1, 0, 1), proto)
 	ie6 := entities.NewInfoElementWithValue(entities.NewInfoElement("sourcePodName", 101, 13, 55829, 65535), srcPod)
 	ie7 := entities.NewInfoElementWithValue(entities.NewInfoElement("destinationPodName", 103, 13, 55829, 65535), dstPod)
-	ie8 := entities.NewInfoElementWithValue(entities.NewInfoElement("destinationClusterIP", 106, 18, 55829, 4), nil)
-	ie9 := entities.NewInfoElementWithValue(entities.NewInfoElement("destinationServicePort", 107, 2, 55829, 2), nil)
+	ie8 := entities.NewInfoElementWithValue(entities.NewInfoElement("destinationClusterIPv4", 106, 18, 55829, 4), net.IP{0, 0, 0, 0})
```
zyiou, line 120: `net.IP{0,0,0,0}` => `bytes.NewBuffer(net.IP{0, 0, 0, 0})`

zyiou (commented):
Thanks for the PR. Current changes makes sense to me. The only concern is that for inter-node case,  if two consecutive records are from source node without receiving record from destination node, it will discard the second record from source node (only update the stats).

> I can fill up the stats and timestamp code too, if you haven't made too much progress on that front.

Since the structure is changed quite a bit, I may need to start over for stats. It will be helpful if you have time to do that. I can also pick it up next week. Thanks!

pkg/intermediate/aggregate_test.go
```diff
@@ -266,18 +271,17 @@ func TestCorrelateRecords(t *testing.T) {
 	flowKey1, _ := getFlowKeyFromRecord(record1)
 	record2 := createMsgwithDataSet2().GetSet().GetRecords()[0]
 	flowKey2, _ := getFlowKeyFromRecord(record2)
-	aggregationProcess.correlateRecords(*flowKey1, record1)
-	aggregationProcess.correlateRecords(*flowKey2, record2)
+	aggregationProcess.aggregateRecord(*flowKey1, record1)
+	aggregationProcess.aggregateRecord(*flowKey2, record2)
```
zyiou, line 275: The logic problem when correlating ipv4 fields is not reflected in the unit tests. I think we should add a case that reverse the sequence of sending these two records, which will cover the case that destination record is received first.

srikartati: Yes, there was a typo for the case of uint and ints in L187, it carried over. 
We are missing that scenario. In addition, we are also missing tests for intra-node flows. Was planning to add them.

pkg/intermediate/aggregate.go
```diff
@@ -118,74 +130,150 @@ func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
...
+						klog.Warningf("This field with name %v should not have been filled with value %v in existing record.", field, existingIeWithValue.Value)
+					}
+					existingIeWithValue.Value = ieWithValue.Value
+				}
+			case entities.Ipv6Address:
+				ipInString := ieWithValue.Value.(net.IP).To16().String()
+				dummyIP := net.IP{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
+				if ipInString == dummyIP.To16().String() {
+					existingIeWithValue, _ := existingRecord.GetInfoElementWithValue(field)
+					ipInString := existingIeWithValue.Value.(net.IP).To4().String()
```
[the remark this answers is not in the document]
srikartati: yes

pkg/intermediate/aggregate.go
```diff
@@ -118,74 +130,150 @@ func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
...
+	if exist {
+		if !isRecordIntraNode(record) {
+			// Do correlation of records if record belongs to inter-node flow and
+			// records from source and destination node are not received.
+			if !aggregationRecord.ReadyToSend && !areRecordsFromSameNode(record, aggregationRecord.Record) {
+				a.correlateRecords(record, aggregationRecord.Record)
+				aggregationRecord.ReadyToSend = true
+			} else {
+				// If the record from the node is already present, update the stats
+				// and timestamps.
```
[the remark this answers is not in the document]
srikartati: We are not discarding the record as the previous record from the same node (either source/destination) will have the same flow key fields and metadata fields except for new stats and timestamps.

pkg/intermediate/aggregate.go
srikartati: yes

pkg/intermediate/aggregate.go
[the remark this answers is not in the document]
zyiou: Got it. we will only update existing record based on upcoming record. what if the existing record has not been exported and the upcoming record comes? will flow aggregator export one record instead of two?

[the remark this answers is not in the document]
srikartati: Yes, exporter in flow aggregator will export one record instead of two. The most updated version of it.

zyiou (commented):
Thanks for adding the stats part. 
I have a question about the concept of the aggregated stats, e.g. for `octetDeltaCountFromSourceNode`, we are summing up all the `octetDeltaCount` if the record is from source node, no matter what the source node is, right? What would be the use cases for that? I thought we would aggregate based on current source node name. Just want to make sure about this. Thanks!

pkg/intermediate/aggregate.go
```diff
@@ -23,28 +40,24 @@ type AggregationProcess struct {
...
-	Protocol           uint8
-	SourcePort         uint16
-	DestinationPort    uint16
-}
-
 type AggregationInput struct {
 	MessageChan     chan *entities.Message
 	WorkerNum       int
 	CorrelateFields []string
+	aggregateElements *AggregationElements
```
shihhaoli, line 58: This is a private member, so callers can not pass it.
Do we need it in the public input struct?

pkg/intermediate/aggregate.go
```diff
@@ -124,74 +140,272 @@ func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
...
-		for _, existingRec := range existingRecords {
-			if isRecordFromSrc(existingRec) {
-				for _, field := range a.correlateFields {
-					if ieWithValue, exist := record.GetInfoElementWithValue(field); exist {
-						existingIeWithValue, _ := existingRec.GetInfoElementWithValue(field)
-						existingIeWithValue.Value = ieWithValue.Value
+		// Add all the new stat fields and initialize them.
+		if !isRecordIntraNode(record) {
+			if isRecordFromSrc(record) {
+				if err := a.addFieldsForStatsAggregation(record, true, false); err != nil {
```
shihhaoli, line 175: Just curious, why only checking error in the exist=false cases?

pkg/intermediate/aggregate.go
```diff
@@ -124,74 +140,272 @@ func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
...
+			if err := a.addFieldsForStatsAggregation(record, true, true); err != nil {
+				return err
+			}
+		}
+		aggregationRecord = AggregationFlowRecord{
+			record,
+			false,
+			true,
+		}
+		if isRecordIntraNode(record) {
```
shihhaoli, line 193: This seems not consistent with line 151, where ReadyToSend is set to true when !isRecordIntraNode(record) ?

pkg/intermediate/aggregate.go
```diff
@@ -124,74 +140,272 @@ func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
...
+			case entities.String:
+				if ieWithValue.Value != "" {
+					existingIeWithValue, _ := existingRecord.GetInfoElementWithValue(field)
+					if existingIeWithValue.Value != "" {
+						klog.Warningf("This field with name %v should not have been filled with value %v in existing record.", field, existingIeWithValue.Value)
 					}
+					existingIeWithValue.Value = ieWithValue.Value
 				}
+			case entities.Unsigned16:
+				if ieWithValue.Value != uint16(0) {
```
shihhaoli, line 217: Could 0 be a valid value??

pkg/intermediate/aggregate.go
```diff
@@ -124,74 +140,272 @@ func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
...
+func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord entities.Record, fillSrcStats, fillDstStats bool) {
+	if a.aggregateElements == nil {
+		return
+	}
+
+	for _, field := range a.aggregateElements.nonStatsElements {
+		if ieWithValue, exist := incomingRecord.GetInfoElementWithValue(field); exist {
+			switch ieWithValue.Element.Name {
+			case "flowEndSeconds":
+				existingIeWithValue, _ := existingRecord.GetInfoElementWithValue(field)
```
shihhaoli, line 262: So no need to check if the field exists in existing record?

pkg/intermediate/aggregate.go
```diff
@@ -124,74 +140,272 @@ func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
...
+				// reset after sending the record from flowKeyMap in aggregation process.
+				// Delta stats from source and destination nodes are added, so we will have
+				// two times the stats approximately.
+				// For delta stats, it is better to use source and destination specific
+				// stats.
+				existingIeWithValue.Value = existingIeWithValue.Value.(uint64) + ieWithValue.Value.(uint64)
+			}
+			// Update the corresponding source element in antreaStatsElement list.
+			if fillSrcStats {
+				existingIeWithValue, _ := existingRecord.GetInfoElementWithValue(antreaSourceStatsElements[i])
```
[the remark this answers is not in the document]
shihhaoli: Yes, I would suggest to make statsElementList, antreaSourceStatsElements, and antreaDestinationStatsElements all independent string lists. Also it looks like there should be no duplicated elements in these three lists, otherwise the stats could be double or triple counted?

pkg/intermediate/aggregate.go
```diff
@@ -124,74 +140,272 @@ func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
...
+		if err != nil {
+			return err
+		}
+	}
+	// Initialize the values of newly added stats info elements.
+	for i, element := range statsElementList {
+		if ieWithValue, exist := record.GetInfoElementWithValue(element); exist {
+			// Initialize the corresponding source element in antreaStatsElement list.
+			if fillSrcStats {
+				existingIeWithValue, _ := record.GetInfoElementWithValue(antreaSourceStatsElements[i])
```
shihhaoli, line 348: Same question as the previous one for here.

pkg/intermediate/aggregate.go
```diff
@@ -124,74 +140,272 @@ func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
...
 	}
-	a.flowKeyRecordMap[flowKey] = append(a.flowKeyRecordMap[flowKey], record)
+	return false
 }
 
+// isRecordFromSrc returns true if record belongs to inter-node flow and from source node.
 func isRecordFromSrc(record entities.Record) bool {
+	if isRecordIntraNode(record) {
+		return false
+	}
```
shihhaoli, line 378: It looks to me we maybe doing a checking on dst twice in some cases, such as src != "" && dst == ""?
How about just doing something like the following?

```
func isRecordFromSrc(record entities.Record) bool {
	srcIEWithValue, exist := record.GetInfoElementWithValue("sourcePodName")
	if !exist || srcIEWithValue.Value == "" {
                return false
        }
        dstIEWithValue, exist := record.GetInfoElementWithValue("destinationPodName")
	if exist && dstIEWithValue.Value != "" {
		return false
	}
	return true
}
```

pkg/intermediate/aggregate.go
```diff
@@ -124,74 +140,272 @@ func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
...
+	ieWithValue, exist := record.GetInfoElementWithValue("destinationPodName")
+	if exist && ieWithValue.Value == "" {
```
shihhaoli, line 380: Should this be
    `if !exist || ieWithValue.Value == "" {`
?

pkg/intermediate/aggregate.go
```diff
@@ -124,74 +140,272 @@ func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
...
+		return true
+	}
+	return false
+}
+
+// isRecordFromDst returns true if record belongs to inter-node flow and from destination node.
+func isRecordFromDst(record entities.Record) bool {
+	if isRecordIntraNode(record) {
+		return false
+	}
```
shihhaoli, line 390: Similarly here.

pkg/intermediate/aggregate.go
```diff
@@ -124,74 +140,272 @@ func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
...
+
+	for _, element := range antreaElements {
+		// Get the new info element from Antrea registry.
+		// TODO: Take antrea registry enterpriseID as input to make this generic.
+		ie, err := registry.GetInfoElement(element, registry.AntreaEnterpriseID)
+		if err != nil {
+			return err
+		}
+		value := new(bytes.Buffer)
+		if err = util.Encode(value, binary.BigEndian, uint64(0)); err != nil {
```
shihhaoli, line 334: Just curious, does the 64-bit integer here cover InfoElement ID, field length, and enterprise number?

srikartati:
> Thanks for adding the stats part. 
> I have a question about the concept of the aggregated stats, e.g. for `octetDeltaCountFromSourceNode`, we are summing up all the `octetDeltaCount` if the record is from source node, no matter what the source node is, right? What would be the use cases for that? I thought we would aggregate based on current source node name. Just want to make sure about this. Thanks!

Let me understand when you say different source node. At any point of time there is only one source node for a given flow in the cluster.
 Presumption is 5-tuple key unique to the flow. If the pod destroys and comes again on a different node.. there will be a new flow. Source port varies and IP can be different. 
As we are aggregating records for same flow key, source node stays the same. 
Ultimately when the record from aggregation process is exported, user need to reset the stats.

zyiou:
> Let me understand when you say different source node. At any point of time there is only one source node for a given flow in the cluster.
> Presumption is 5-tuple key unique to the flow. If the pod destroys and comes again on a different node.. there will be a new flow. Source port varies and IP can be different.
> As we are aggregating records for same flow key, source node stays the same.
> Ultimately when the record from aggregation process is exported, user need to reset the stats.

Sorry I forgot the context we are aggregating under the same flow key. Then it makes sense to me. So the `octetTotalCountFromSourceNode` is actually total octet count from source node for this specific flow, but not total octet count for all flows whose source is current node, right?

srikartati:
>  > Let me understand when you say different source node. At any point of time there is only one source node for a given flow in the cluster.
> > Presumption is 5-tuple key unique to the flow. If the pod destroys and comes again on a different node.. there will be a new flow. Source port varies and IP can be different.
> > As we are aggregating records for same flow key, source node stays the same.
> > Ultimately when the record from aggregation process is exported, user need to reset the stats.
> 
> Sorry I forgot the context we are aggregating under the same flow key. Then it makes sense to me. So the `octetTotalCountFromSourceNode` is actually total octet count from source node for this specific flow, but not total octet count for all flows whose source is current node, right?
> 
> 

Yup that's correct.

pkg/intermediate/aggregate.go
srikartati: Yes. For example, for pod-to-pod flow the destinationServicePort (k8s service port) is not set.

shihhaoli: If 0 could be a valid value in some cases, do we still skip the storing when the incoming record has 0 value ? Or the 0 value actually has no useful info?

pkg/intermediate/aggregate.go
srikartati: Done.

pkg/intermediate/aggregate.go
srikartati: No we do not skip storing it, we store everything sent by exporter at collector. If there is a field in IPFIX template, the exporter has to send some value, otherwise collector cannot validate the corresponding IPFIX data record. 0 value of course do not have useful info, but also needed in some case where it has to be correlated with its corresponding record and filled up. 

We have to use and manage multiple templates if the field has no use for specific use case for example destinationServicePortName/destinationService are not applicable for Pod-To-Pod flows. Currently, we use only one template in Antrea, the user of go-ipfix library.

pkg/intermediate/aggregate.go
srikartati: Yes not needed as we do the check, we already check for this in addFieldsForStatsAggregation. The assumption is that all the flow records for a given flow follow the same template.

pkg/intermediate/aggregate.go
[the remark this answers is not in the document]
srikartati: Yes this part need to have more checks that validate the assumptions taken in the code. As Antrea is the only user and we control the input. 
I want to add a TODO and move on, considering the time factor. What do you say?

pkg/intermediate/aggregate.go
srikartati: It just initializes the value for new stat fields added for source and destination.

pkg/intermediate/aggregate.go
srikartati: Here as well we check in the add new fields function and presume that records come with same template.

pkg/intermediate/aggregate.go
srikartati: Good suggestion. Done.

pkg/intermediate/aggregate.go
srikartati: We expect destinationPodName to be present. Again we presume same template use case on both source and destination nodes. 
In future, we have to extend where there are different templates for source and destination node, but we probably have to consider different cases. We need to have one template intra-node flow record, one template for inter-node source node flow record, one template for inter-node destination flow record etc. This case is not supported.

pkg/intermediate/aggregate.go
srikartati: done

srikartati (commented):
Thanks for the review, @shihhaoli 

Addressed comments.

pkg/intermediate/aggregate.go
```diff
@@ -54,12 +69,13 @@ func InitAggregationProcess(input AggregationInput) (*AggregationProcess, error)
...
-		make(map[FlowKey][]entities.Record),
+		make(map[FlowKey]AggregationFlowRecord),
 		sync.RWMutex{},
 		input.MessageChan,
 		input.WorkerNum,
 		make([]*worker, 0),
 		input.CorrelateFields,
+		input.AggregateElements,
 		make(chan bool),
 	}, nil
```
zyiou, line 80: nit: it will be better to specify corresponding field names here for future maintenance.

pkg/intermediate/aggregate.go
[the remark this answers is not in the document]
zyiou: I'm ok with adding TODO comment.

zyiou (approved):
LGTM. just ensure the TODO comment is added. Thanks!

pkg/intermediate/aggregate.go
shihhaoli: I guess here the default value for unint16 type is 0, so only a non-zero value from incoming record needs to be stored.

pkg/intermediate/aggregate.go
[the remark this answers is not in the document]
shihhaoli: OK too

shihhaoli (approved):

pkg/intermediate/aggregate.go
srikartati: Yes, we only need to correlate non-zero uint16 value and fill up the existing record in flowKeyRecord map.

srikartati:
> LGTM. just ensure the TODO comment is added. Thanks!

It was added in L45.

PR #101, opened 2020-12-08
Support client authentication for TLS

zyiou:
Add support for root CA and client authentication, which are essential parts for mTLS support in Antrea.


pkg/collector/process.go
```diff
@@ -49,7 +49,8 @@ type CollectingProcess struct {
 	clients map[string]*clientHandler
 	// isEncrypted indicates whether to use TLS/DTLS for communication
 	isEncrypted bool
-	// serverCert and serverKey are for storing encryption info when using TLS/DTLS
+	// caCert, serverCert and serverKey are for storing encryption info when using TLS/DTLS
+	caCert     []byte
```
srikartati, line 53: So, we need CACert as a public key to verify or authenticate the owner of ServerCert/ClientCert + ServerKey/ClientKey. Is my understanding correct?

pkg/collector/process.go
```diff
@@ -59,6 +60,7 @@ type CollectorInput struct {
 	MaxBufferSize uint16
 	TemplateTTL   uint32
 	IsEncrypted   bool
+	CACert        []byte
 	ServerCert    []byte
 	ServerKey     []byte
```
srikartati, line 65: should we group them in a separate structure for this and use it for both exporter and collector by renaming ServerCert/ClientCert -> Cert and other input to Key? 
Not sure if it changes any meaning for the respective processes.

Ok to move on with a TODO as this is a minor comment for cleaner code.

srikartati:
A request @zyiou. As part of your rebase, could you fix the import pkg grouping: https://github.com/vmware/go-ipfix/blob/master/pkg/intermediate/aggregate.go#L21

Seems like go import do not seem to be working that run in linters workflow.

pkg/collector/process.go
zyiou: Yes.

pkg/collector/process.go
zyiou: I added a TODO comment here for changes later.

srikartati (dismissed):

srikartati (approved):

PR #108, opened 2020-12-09
Kafka producer supporting multiple proto schemas

srikartati:
Demonstrate the ability to support different proto schemas
    
As part of this commit, we demonstrate that Kafka producer can
support encoding IPFIX flow records in different protobuf schemas.

We consider two different protobuf schema with same fields for
simplicity.


pkg/producer/kafka.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/vmware/go-ipfix/pkg/producer/protobuf"
+)
+
+var (
+	KafkaConfigVersion sarama.KafkaVersion
+	// ProtoSchemaTypes lists all proto schema supported in Kafka producer.
+	ProtoSchemaTypes = map[string]string{
+		"FlowType1": "FlowType1",
+		"FlowType2": "FlowType2",
+	}
```
shihhaoli, line 34: Just curious, do you really need the value part since it is the same as the key?
If not, maybe we can store the mapping as map[string]convertRecordToFlowType?
For example,

```
	ProtoSchemaTypeConverterTable = map[string]sconvertRecordToFlowType{
		"FlowType1": convertRecordToFlowType1,
		"FlowType2": convertRecordToFlowType2,
	}
```
	
Thus lines 52-57 can be simplified?

srikartati: Good idea.

jianjuns: +1

pkg/producer/kafka.go
```diff
@@ -0,0 +1,129 @@
...
+func convertIPFIXMsgToFlowMsgs(msg *entities.Message, protoSchemaType string) []*protobuf.FlowMessage {
+	set := msg.GetSet()
+	if set.GetSetType() == entities.Template {
+		return nil
+	}
+
+	var convertRecordFunc convertRecordToFlowType
+	switch protoSchemaType {
+	case "FlowType1":
+		convertRecordFunc = convertRecordToFlowType1
```
jianjuns, line 54: Can we save convertRecordToFlowType1 to KafkaProducer struct?

pkg/producer/util.go
```diff
@@ -0,0 +1,194 @@
...
+import (
+	"net"
+
+	"k8s.io/klog"
+
+	"github.com/vmware/go-ipfix/pkg/entities"
+	"github.com/vmware/go-ipfix/pkg/producer/protobuf"
+)
+
+type convertRecordToFlowType func(msg *entities.Message, record entities.Record) *protobuf.FlowMessage
```
jianjuns, line 26: Do you think better to define a Go interface for converter, which includes a convert func?

jianjuns (dismissed):

pkg/producer/kafka.go
srikartati: sure.

pkg/producer/util.go
srikartati: We can. Do you envision that more functions may be required in addition to converter function for each protobuf schema type?

jianjuns: Just feel that is more OO design.

pkg/producer/util.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+	}
+	return &protobuf.FlowMessage{
+		FlowType: &protobuf.FlowMessage_Flow2{Flow2: flowType2},
+	}
+}
+
+// convertIPFIXMsgToFlowMsgs converts data records in IPFIX message to flow messages
+// in a given proto schema.
+func convertIPFIXMsgToFlowMsgs(msg *entities.Message, convertRecordFunc convertRecordToFlowType) []*protobuf.FlowMessage {
```
shihhaoli, line 198: Should we return a pointer to []*protobuf.FlowMessage, assuming there could be a huge number of messages being converted?

pkg/producer/util.go
```diff
@@ -0,0 +1,210 @@
...
+	set := msg.GetSet()
+	if set.GetSetType() == entities.Template {
+		return nil
+	}
+
+	flowMsgs := make([]*protobuf.FlowMessage, 0)
+	for _, record := range set.GetRecords() {
+		flowMsg := convertRecordFunc(msg, record)
+		flowMsgs = append(flowMsgs, flowMsg)
+	}
```
shihhaoli, line 208: Trivial comment, since you use make on line 204, the following code could be a little more efficient.

        records := set.GetRecords() 
        flowMsgs := make([]*protobuf.FlowMessage, len(records))
   	for i, record := range records {
		flowMsgs[i] = convertRecordFunc(msg, record)
	}

srikartati: Agreed. Initializing fixed size slice makes it efficient.

pkg/producer/util.go
srikartati: Internally slice is implemented as a pointer, so returning pointer to the slice is redundant?

shihhaoli: Yes, if that's case, then we don't need to return the pointer to it. Thanks.

pkg/producer/util.go
srikartati: Thanks for the suggestion. 
Sorry I could not get to this earlier. Addressed this comment. Could you take a look at this again?

zyiou (commented):
Overall LGTM. 
I can see that `pkg/producer/convertor.go` and folder `pkg/producer/protobuf/` are basically the examples of convertors users need to customize if they want to use Kafka producer. Without considering demo multiple convertors, in our current use case, only one convertor is enough, right?

srikartati:
> Overall LGTM.
> I can see that `pkg/producer/convertor.go` and folder `pkg/producer/protobuf/` are basically the examples of convertors users need to customize if they want to use Kafka producer. Without considering demo multiple convertors, in our current use case, only one convertor is enough, right?

Not sure what use case you have in your mind. In this PR, we tried to build some key pieces to support more proto schemas that will help to support any kafka broker system. This makes the library more useful.

zyiou:
> > Overall LGTM.
> > I can see that `pkg/producer/convertor.go` and folder `pkg/producer/protobuf/` are basically the examples of convertors users need to customize if they want to use Kafka producer. Without considering demo multiple convertors, in our current use case, only one convertor is enough, right?
> 
> Not sure what use case you have in your mind. In this PR, we tried to build some key pieces to support more proto schemas that will help to support any kafka broker system. This makes the library more useful.

Got it. I was thinking that we only need one proto schema which matches two templates (IPv4 and IPv6) in our case. Yes I can see that current PR is to make producer more extensible. Thanks!

zyiou (approved):

pkg/producer/convertor.go
```diff
@@ -0,0 +1,246 @@
...
+
+	"k8s.io/klog/v2"
+
+	"github.com/vmware/go-ipfix/pkg/entities"
+	"github.com/vmware/go-ipfix/pkg/producer/protobuf"
+)
+
+// ConvertIPFIXToKafka is an interface to support multiple proto schema for Kafka
+// producer.
+type ConvertIPFIXToKafka interface {
```
jianjuns, line 28: IPFIXToKafkaConverter?

pkg/producer/convertor.go
```diff
@@ -0,0 +1,246 @@
...
+			schemaType: "FlowType2",
+		}
+	default:
+		return nil
+	}
+}
+
+// convertRecordToFlowType1 is to support the FlowType1 proto schema.
+type convertRecordToFlowType1 struct {
+	schemaType string
```
jianjuns, line 51: But why you need this field if convertRecordToFlowType1 support only flowType1?

pkg/producer/convertor.go
```diff
@@ -0,0 +1,246 @@
...
+// producer.
+type ConvertIPFIXToKafka interface {
+	// convertIPFIXMsgToFlowMsgs converts multiple data records in the IPFIX message
+	// to flow messages for any given proto schema.
+	convertIPFIXMsgToFlowMsgs(msg *entities.Message) []*protobuf.FlowMessage
+}
+
+func NewConvertorForProtoSchema(schemaType string) ConvertIPFIXToKafka {
+	switch schemaType {
+	case "FlowType1":
```
jianjuns, line 36: Define two constants?

jianjuns, line 36: I feel cleaner to define a map and add schemaType -> converter mapping there, or you can define a registration func for each schema to register itself.

pkg/producer/convertor.go
```diff
@@ -0,0 +1,246 @@
...
+		return nil
+	}
+}
+
+// convertRecordToFlowType1 is to support the FlowType1 proto schema.
+type convertRecordToFlowType1 struct {
+	schemaType string
+}
+
+func (c *convertRecordToFlowType1) convertIPFIXMsgToFlowMsgs(msg *entities.Message) []*protobuf.FlowMessage {
```
jianjuns, line 54: Are Type1 and 2 added for testing only? If so, we should put the code in some xyz_test.go files?

pkg/producer/convertor.go
srikartati: More appropriate. Done.

pkg/producer/convertor.go
srikartati: done

srikartati: Addressed. Please take a look.

pkg/producer/convertor.go
srikartati: Not really needed. Did not want to keep the struct empty. Removed it.

pkg/producer/convertor.go
srikartati: Yes, that is correct. Makes sense. I created a couple of test files that will serve as a template for actual proto schemas of Kafka broker systems.

srikartati (commented):
Thanks for the review @jianjuns

pkg/producer/convertor.go
srikartati: I had to change the package structure to include test files in integration tests (collector_producer_test.go)

pkg/producer/convertor/convertor.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"github.com/vmware/go-ipfix/pkg/entities"
+	"github.com/vmware/go-ipfix/pkg/producer/protobuf"
+)
+
+type RegisterProtoSchema func() IPFIXToKafkaConverter
+
+const (
+	FlowType1 string = "FlowType1"
```
jianjuns, line 25: Do we need to define them here? Could they be in the test code?

srikartati: Yes, that is the appropriate place.

jianjuns (dismissed):
Thanks for addressing the comments.

srikartati:
> Thanks for addressing the comments.

Thanks for making this change better.

pkg/producer/convertor/convertor.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/vmware/go-ipfix/pkg/producer/protobuf"
+)
+
+type RegisterProtoSchema func() IPFIXToKafkaConverter
+
+var ProtoSchemaConvertor = map[string]RegisterProtoSchema{}
+
+// IPFIXToKafkaConverter is an interface to support multiple proto schema for Kafka
+// producer.
+type IPFIXToKafkaConverter interface {
```
zyiou, line 28: It's better to keep spelling of convertor consistent -> either convertor or converter

pkg/producer/convertor/convertor.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2020 VMware, Inc.
```
zyiou, line 1: Maybe we should add support for checking year of newly-added file.

zyiou (commented):
Thanks for the changes, the layout with separate folder for convertor and testing cases looks clearer to me. Only one nit.

pkg/producer/convertor/convertor.go
srikartati: yes we should.

zyiou (approved):
LGTM

PR #110, opened 2020-12-14
Add new delete function in aggregation process

srikartati:
The current delete function acquires lock. In some cases, caller will already
have acquired the lock in the aggregation process. In those cases, we
will need another function that does not acquire lock and deletes the flowKey from the map.

### DIFF
--- a/pkg/intermediate/aggregate.go
+++ b/pkg/intermediate/aggregate.go
@@ -105,9 +105,18 @@ func (a *AggregationProcess) ForAllRecordsDo(callback FlowKeyRecordMapCallBack) 
 	return nil
 }
 
-func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
+func (a *AggregationProcess) DeleteFlowKeyFromMapWithLock(flowKey FlowKey) {
 	a.flowKeyRecordLock.Lock()
 	defer a.flowKeyRecordLock.Unlock()
+	delete(a.flowKeyRecordMap, flowKey)
+}
+
+// DeleteFlowKeyFromMapWithoutLock need to be used only when the caller has already
+// acquired the lock. For example, this can be used in a callback of ForAllRecordsDo
+// function.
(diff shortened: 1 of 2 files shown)


pkg/intermediate/aggregate_test.go
```diff
@@ -265,16 +265,34 @@ func TestCorrelateRecords(t *testing.T) {
 	}
 }
 
-func TestDeleteTupleFromMap(t *testing.T) {
+func TestDeleteTupleFromMapWithLock(t *testing.T) {
```
zyiou, line 268: Not in this PR, can you help change Tuple -> FlowKey in test names for this and next testing function? Thanks!

pkg/intermediate/aggregate.go
```diff
@@ -105,12 +105,16 @@ func (a *AggregationProcess) ForAllRecordsDo(callback FlowKeyRecordMapCallBack)
...
-func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) {
+func (a *AggregationProcess) DeleteFlowKeyFromMapWithLock(flowKey FlowKey) {
 	a.flowKeyRecordLock.Lock()
 	defer a.flowKeyRecordLock.Unlock()
 	delete(a.flowKeyRecordMap, flowKey)
 }
 
+func (a *AggregationProcess) DeleteFlowKeyFromMapWithoutLock(flowKey FlowKey) {
+	delete(a.flowKeyRecordMap, flowKey)
+}
```
zyiou, line 116: how would users acquire the lock if they want to use this function? Should we make lock public?

srikartati: Added a comment. Ideally, we should not expose this. This is a temporary addition until we support the concept of invalid flows in Flow Exporter and thereby in the aggregation process, which will make flowKey deletion easier in flow aggregator of Antrea.

zyiou (approved):
LGTM

PR #114, opened 2020-12-14
Fix testing problems for TestCollectorToIntermediate and golangci-lint

### LINKED ISSUE #109: TestCollectorToIntermediate is flaky
**Describe the bug**
TestCollectorToIntermediate is flaky. It is not passing consistently. Example run: https://github.com/vmware/go-ipfix/runs/1539468529

**To Reproduce**
Run the test locally with count=50. 

**Expected behavior**
Test should pass consistently.

**Actual Behavior**
I1211 19:59:29.646039    4907 tcp.go:38] Start tcp collecting process on [::]:4739
I1211 19:59:29.749036    4907 tcp.go:73] Connection from 127.0.0.1:49328 has been closed.
--- FAIL: TestCollectorToIntermediate (0.21s)
    collector_intermediate_test.go:167: 
        	Error Trace:	collector_intermediate_test.go:167
        	Error:      	Not equal: 
        	            	expected: 0x190
        	            	actual  : 0x3e8
        	Test:       	TestCollectorToIntermediate
    collector_intermediate_test.go:177: 
        	Error Trace:	collector_intermediate_test.go:177
        	Error:      	Not equal: 
        	            	expected: 0x3e8
        	            	actual  : 0x190
        	Test:       	TestCollectorToIntermediate

**Versions**
ToT master branch

### LINKED ISSUE #119: Github Actions Golangci-lint is not working correctly
**Describe the bug**
Golangci-lint is passed automatically even though there is error in `make golangci-lint`.

**To Reproduce**
run `make golangci-lint` locally with some error logs. when pushing the code, the workflow will pass.
Here is an [example](https://github.com/vmware/go-ipfix/runs/1561102825?check_suite_focus=true) run.

**Expected behavior**
workflow should not pass when there are errors in golangci-lint and marked workflow as failed.

**Actual Behavior**
Golangci-lint workflow passed.

**Versions**
master branch

**Additional context**

zyiou:
fixes #109 
fixes #119

### DIFF
--- a/pkg/collector/process_test.go
+++ b/pkg/collector/process_test.go
@@ -178,7 +178,7 @@ func init() {
 }
 
 func TestTCPCollectingProcess_ReceiveTemplateRecord(t *testing.T) {
-	address, err := net.ResolveTCPAddr("tcp", "0.0.0.0:4730")
+	address, err := net.ResolveTCPAddr("tcp", "127.0.0.1:0")
 	if err != nil {
 		t.Error(err)
 	}
@@ -196,12 +196,12 @@ func TestTCPCollectingProcess_ReceiveTemplateRecord(t *testing.T) {
 	}
 	go cp.Start()
 	// wait until collector is ready
-	waitForCollectorReady(t, address)
-
+	waitForCollectorReady(t, cp)
+	collectorAddr := cp.GetAddress()
 	go func() {
-		conn, err := net.Dial(address.Network(), address.String())
+		conn, err := net.Dial(collectorAddr.Network(), collectorAddr.String())
 		if err != nil {
-			t.Errorf("Cannot establish connection to %s", address.String())
+			t.Errorf("Cannot establish connection to %s", collectorAddr.String())
 		}
 		defer conn.Close()
 		conn.Write(validTemplatePacket)
@@ -213,7 +213,7 @@ func TestTCPCollectingProcess_ReceiveTemplateRecord(t *testing.T) {
 }
 
 func TestUDPCollectingProcess_ReceiveTemplateRecord(t *testing.T) {
-	address, err := net.ResolveUDPAddr("udp", "0.0.0.0:4731")
+	address, err := net.ResolveUDPAddr("udp", "127.0.0.1:0")
 	if err != nil {
 		t.Error(err)
 	}
@@ -231,10 +231,10 @@ func TestUDPCollectingProcess_ReceiveTemplateRecord(t *testing.T) {
 	}
 	go cp.Start()
 	// wait until collector is ready
-	waitForCollectorReady(t, address)
-
+	waitForCollectorReady(t, cp)
+	collectorAddr := cp.GetAddress()
 	go func() {
-		resolveAddr, err := net.ResolveUDPAddr(address.Network(), address.String())
+		resolveAddr, err := net.ResolveUDPAddr(collectorAddr.Network(), collectorAddr.String())
 		if err != nil {
 			t.Errorf("UDP Address cannot be resolved.")
 		}
@@ -253,7 +253,7 @@ func TestUDPCollectingProcess_ReceiveTemplateRecord(t *testing.T) {
 }
 
 func TestTCPCollectingProcess_ReceiveDataRecord(t *testing.T) {
-	address, err := net.ResolveTCPAddr("tcp", "0.0.0.0:4732")
+	address, err := net.ResolveTCPAddr("tcp", "127.0.0.1:0")
 	if err != nil {
 		t.Error(err)
 	}
@@ -275,12 +275,12 @@ func TestTCPCollectingProcess_ReceiveDataRecord(t *testing.T) {
 	go cp.Start()
 
 	// wait until collector is ready
-	waitForCollectorReady(t, address)
-
+	waitForCollectorReady(t, cp)
+	collectorAddr := cp.GetAddress()
 	go func() {
-		conn, err := net.Dial(address.Network(), address.String())
+		conn, err := net.Dial(collectorAddr.Network(), collectorAddr.String())
 		if err != nil {
-			t.Errorf("Cannot establish connection to %s", address.String())
+			t.Errorf("Cannot establish connection to %s", collectorAddr.String())
 		}
 		defer conn.Close()
 		conn.Write(validDataPacket)
@@ -290,7 +290,7 @@ func TestTCPCollectingProcess_ReceiveDataRecord(t *testing.T) {
 }
 
 func TestUDPCollectingProcess_ReceiveDataRecord(t *testing.T) {
-	address, err := net.ResolveUDPAddr("udp", "0.0.0.0:4733")
+	address, err := net.ResolveUDPAddr("udp", "127.0.0.1:0")
 	if err != nil {
 		t.Error(err)
 	}
@@ -311,10 +311,10 @@ func TestUDPCollectingProcess_ReceiveDataRecord(t *testing.T) {
 
 	go cp.Start()
 	// wait until collector is ready
-	waitForCollectorReady(t, address)
-
+	waitForCollectorReady(t, cp)
+	collectorAddr := cp.GetAddress()
 	go func() {
-		resolveAddr, err := net.ResolveUDPAddr(address.Network(), address.String())
+		resolveAddr, err := net.ResolveUDPAddr(collectorAddr.Network(), collectorAddr.String())
 		if err != nil {
 			t.Errorf("UDP Address cannot be resolved.")
 		}
@@ -330,7 +330,7 @@ func TestUDPCollectingProcess_ReceiveDataRecord(t *testing.T) {
 }
 
 func TestTCPCollectingProcess_ConcurrentClient(t *testing.T) {
-	address, err := net.ResolveTCPAddr("tcp", "0.0.0.0:4734")
+	address, err := net.ResolveTCPAddr("tcp", "127.0.0.1:0")
 	if err != nil {
 		t.Error(err)
 	}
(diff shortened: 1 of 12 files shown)


pkg/intermediate/aggregate.go
```diff
@@ -304,7 +309,13 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
 			}
 			// Update the corresponding source element in antreaStatsElement list.
 			if fillSrcStats {
-				existingIeWithValue, _ := existingRecord.GetInfoElementWithValue(antreaSourceStatsElements[i])
+				name := antreaSourceStatsElements[i]
+				var existingIeWithValue *entities.InfoElementWithValue
+				if isIntraNode {
+					existingIeWithValue, _ = existingRecord.GetInfoElementWithValue(name)
+				} else {
+					existingIeWithValue, _ = existingRecord.GetInfoElementWithValue(getReverseInfoElementName(name))
```
srikartati, line 317: Why is this needed? In the case of inter node flows, flows in conntrack module on both nodes follow the same directions.

pkg/test/collector_intermediate_test.go
```diff
@@ -143,6 +143,8 @@ func TestCollectorToIntermediate(t *testing.T) {
 		defer conn.Close()
 		conn.Write(templatePacket)
 		conn.Write(dataPacket1)
+		// sleep to ensure dataPacket1 is processed first so that the total count fields are using values from dataPacket1
+		time.Sleep(time.Millisecond)
```
srikartati, line 147: This is TCP connection right. Ord
er of packets should be followed. Did you get a chance to run the test with count=500 or some large number?

pkg/intermediate/aggregate.go
zyiou: you mean that for flow record from src and flow record from dst, if they represent the same flow, they should have same `packetDeltaCount` and `packetTotalCount`? 
I was under impression that the `packetDeltaCount` in src record should be the same as `reversePacketDeltaCount` in dst record.

srikartati: Yup it is the same packet that is used to create conntrack flow on both nodes, so its direction is considered as original direction and the other direction is reverse direction. You should dump the conntrack flows in the case of inter-node flow scenario and see this.

pkg/collector/process_test.go
```diff
@@ -628,14 +634,14 @@ func TestDTLSCollectingProcess(t *testing.T) {
...
 }
 
-func waitForCollectorReady(t *testing.T, address net.Addr) {
+func waitForCollectorReady(t *testing.T, cp *CollectingProcess) {
 	checkConn := func() (bool, error) {
-		if _, err := net.Dial(address.Network(), address.String()); err != nil {
+		if _, err := net.Dial(cp.GetAddress().Network(), cp.GetAddress().String()); err != nil {
 			return false, err
 		}
 		return true, nil
```
srikartati, line 642: Could we close the connection before returning true, so we can check for exact connections in this line:
https://github.com/vmware/go-ipfix/pull/114/files#diff-24e923a024538e6ea8a21097adf7e88b862e14a44c743290238252e0a429d1e7R362 ?

pkg/collector/tcp.go
```diff
@@ -47,18 +45,15 @@ func (cp *CollectingProcess) startTCPServer() {
...
 			}
-			wg.Add(1)
-			go cp.handleTCPClient(conn, &wg)
+			go cp.handleTCPClient(conn)
 		}
 	}()
 	<-cp.stopChan
 	// close all connections
 	cp.closeAllClients()
-	wg.Wait()
```
srikartati, line 57: As stopChan is there, is waitGroup redundant?

pkg/test/collector_intermediate_test.go
```diff
@@ -56,14 +56,14 @@ dataPacket 2:
 	"destinationTransportPort": 5678
 	"protocolIdentifier": 6
 	"flowEndSeconds": 1257895000
-	"packetTotalCount": 400
-	"packetDeltaCount": 200
+	"packetTotalCount": 1000
```
srikartati, line 59: why same? I know it can be similar but as they are not synchronized, they may not be exactly the same stats. 

Does the order issue with tcp still exists? I am just curious.

srikartati (commented):
Thanks for taking care of go lint issue.

Why checking into release-0.4.0? Isn't master good enough?

pkg/collector/process_test.go
zyiou: if we close the connection, we cannot send data to corresponding connection. Do you mean we can create another function to wait for two clients shown up and close the connection?

pkg/collector/tcp.go
zyiou: Yes

pkg/test/collector_intermediate_test.go
zyiou: I was just trying to simulate the condition. since they may not be exactly the same, I changed it to different values.
For order issue in tcp, it will not cause order problem in collector since data are received in order. Because aggregation has two workers, the sequence of processing will differ. I have changed the worker number to 1. It should be working now.

zyiou:
> Why checking into release-0.4.0? Isn't master good enough?

Sorry I was using the previous setting. Changed to use master branch as base.

pkg/collector/process_test.go
srikartati: I was talking about the connection created in L639, which we are ignoring right now.

pkg/test/collector_intermediate_test.go
srikartati: Processing with two workers is an important point. We should allow this either sending all flow records of an individual flow to one worker or update the flow records carefully if flow records can be processed by multiple workers.
Should we update only if the value is more than current value for Total stats?

zyiou: Yes it makes sense to update total stats when value is larger. Will update the code then. Thanks!

pkg/producer/kafka_test.go
```diff
@@ -31,14 +31,14 @@ import (
...
 		0x30, 0x2e, 0x32, 0x40, 0xd2, 0x9, 0x48, 0xae, 0x2c, 0x50, 0x6, 0x58, 0xe8,
 		0x7, 0x60, 0xe8, 0x7, 0x68, 0xe8, 0x7, 0x70, 0xe8, 0x7, 0x78, 0x32, 0x80,
 		0x1, 0x32, 0x88, 0x1, 0x32, 0x90, 0x1, 0x32, 0xca, 0x1, 0xb, 0x31, 0x39,
 		0x32, 0x2e, 0x31, 0x36, 0x38, 0x2e, 0x30, 0x2e, 0x31, 0x8a, 0x2, 0x9, 0x31,
 		0x32, 0x37, 0x2e, 0x30, 0x2e, 0x30, 0x2e, 0x31, 0x90, 0x2, 0x83, 0x25}
-	msg2InBytes = []byte{0x86, 0x1, 0x10, 0x1, 0x18, 0xd2, 0x9, 0x32, 0x19, 0x32,
+	msg2InBytes = []byte{0x10, 0x1, 0x18, 0xd2, 0x9, 0x32, 0x19, 0x32,
 		0x30, 0x30, 0x31, 0x3a, 0x30, 0x3a, 0x33, 0x32, 0x33, 0x38, 0x3a, 0x64,
 		0x66, 0x65, 0x31, 0x3a, 0x36, 0x33, 0x3a, 0x3a, 0x66, 0x65, 0x66, 0x62,
 		0x3a, 0x19, 0x32, 0x30, 0x30, 0x31, 0x3a, 0x30, 0x3a, 0x33, 0x32, 0x33,
```
zyiou, line 44: @srikartati when I changed to marshal function from `google.golang.org/protobuf/proto`, the marshal result and encoding result is a little bit different. Do you know the reason for that?

pkg/collector/process_test.go
shihhaoli: Just curious, should waitForCollectorReady return error if collector is not ready after 5 retries?

pkg/producer/kafka_test.go
srikartati: @zyiou Sorry for the delayed response.
The difference is because the `encodedMessage` method prefixes the message with length whereas Marshal doesn't do that. https://godoc.org/github.com/golang/protobuf/proto#Buffer.EncodeMessage

zyiou: Got it. Do I need to prepend the length to current bytes to keep things same as before? Because `google.golang.org/protobuf/proto` does not have the exact same function as `EncodeMessage` in previous version.

srikartati: Looked at this more closely. This depends on the Kafka decoder functionality--some seem to expect length-delimited messages and some do not. I did not consider both options in my previous PR as we have a mock consumer in tests, there was no issue. 
However, it is better to support both with an input bool option "KafkaDelimitMsgWithLen" by defining a default value.

zyiou: Got it. Will add it in this patch since it is part of fixing for golangci-lint. Thanks!

srikartati: Are you thinking of fixing this in separate PR?

zyiou: Yes, when we cherry-pick the fixes from v0.4.3 to master branch, we can fix that. Since now the 0.4.0 branch does not have the Kafka commit.

srikartati: sounds good.

pkg/collector/process_test.go
srikartati: This still seems to be outstanding. Any follow up comment on this?

zyiou: Sorry for the late reply. @srikartati updated the close connection code, now we can check exact number of clients. @shihhaoli yes it will log error if the collector is not ready, which will cause the tests to fail.

srikartati (approved):
LGTM

PR #115, opened 2020-12-15
Expose pathMTU in exporter

zyiou:
so that users can use it to test whether current message size exceeds limit or not


srikartati (approved):
Thanks for adding this. LGTM.

PR #116, opened 2020-12-15
update CHANGELOG for v0.4.2

zyiou:


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.  The format
 Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+## 0.4.2 12-15-2020
+### Added
+- Expose message size limit in exporter. (#115, @zyiou)
```
srikartati, line 9: also the method to delete the flow key from flowKeyMap without lock?

zyiou: Do you mean we need to cherry pick the commit too from v0.3.2?

srikartati: It is already done. #113 
We just need to add it as another "added" change.

srikartati: We just need to add the change separately here. As release-0.3.0 and release-0.4.0 are already broken, changelog can have the same change in different patch releases. There will be only one cherry-pick on to master as part of v0.5.0 we will mention that all the fixes are present in the new release.

zyiou: Got it. so here we are mentioning #113 instead of #110, right?

srikartati: yes thats correct.

srikartati (approved):

srikartati:
Looks like the option is there! Will check again on my end again..

PR #118, opened 2020-12-15
Changes from release v0.4.2

zyiou:


srikartati:
I think it is better to keep the commits separate as we have to squash the commit before we can merge it. I know it is somewhat painful but it will be easy to track back to individual commits that has a proper commit message. What do you think?

zyiou:
> I think it is better to keep the commits separate as we have to squash the commit before we can merge it. I know it is somewhat painful but it will be easy to track back to individual commits that has a proper commit message. What do you think?

I think there is an option for us to rebase and commit when merging to master branch. So the commits will be kept separately.

srikartati:
> > I think it is better to keep the commits separate as we have to squash the commit before we can merge it. I know it is somewhat painful but it will be easy to track back to individual commits that has a proper commit message. What do you think?
> 
> I think there is an option for us to rebase and commit when merging to master branch. So the commits will be kept separately.

Do we have the option? I think that is disabled. I always see "rebase and merge" disabled.

zyiou:
> Do we have the option? I think that is disabled. I always see "rebase and merge" disabled.

I'm under the impression that `rebase and merge` is enabled for master branch. We can have a try after reviewing this code. If `rebase and merge` does not show up, I will create separate PRs for us to merge these commits.

srikartati (approved):

PR #125, opened 2021-01-29
Validating data records in aggregation process

srikartati:
Before aggregating, valid data records if all the records have
proper values after decoding.

### DIFF
--- a/pkg/intermediate/aggregate.go
+++ b/pkg/intermediate/aggregate.go
@@ -110,14 +110,25 @@ func (a *AggregationProcess) AggregateMsgByFlowKey(message *entities.Message) er
 		return nil
 	}
 	records := set.GetRecords()
+	invalidRecs := 0
 	for _, record := range records {
-		flowKey, err := getFlowKeyFromRecord(record)
-		if err != nil {
-			return err
+		// Validate the data record. If invalid, we log the error and move to the next
+		// record.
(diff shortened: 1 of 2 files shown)


pkg/intermediate/aggregate.go
```diff
@@ -110,15 +110,26 @@ func (a *AggregationProcess) AggregateMsgByFlowKey(message *entities.Message) er
...
-		flowKey, err := getFlowKeyFromRecord(record)
-		if err != nil {
-			return err
-		}
-		if err = a.addOrUpdateRecordInMap(flowKey, record); err != nil {
-			return err
+		// Validate the data record. If invalid, we log the error and move to the next
+		// record.
+		if !validateDataRecord(record) {
+			klog.V(2).Infof("Invalid data record because decoded values of elements are not valid.")
```
zyiou, line 118: seems it is more an `Errorf` than `Infof`

zyiou (commented):
Overall LGTM. I will fix the tls certificate issue in the unit tests.

pkg/intermediate/aggregate.go
srikartati: Yes `Errorf` is more appropriate. The intention was to use error log but missed it.

zyiou (approved):
LGTM

PR #127, opened 2021-02-02
Update TLS certificate in testing

zyiou:
TLS certificate in tests was set to default expiration time which is 30 days. Updated the certificates to expire in 1 year.


srikartati (commented):
Could we add a comment somewhere appropriate that the certificates need to be regenerated after a year with the date?

srikartati (approved):
LGTM

PR #129, opened 2021-02-03
Modify the input of Exporting and Collecting process

### LINKED ISSUE #121: Change the address input format in exporter process and collector process
Currently we have net.Addr as the format that provides both "host:port" string and "proto" string. Because of this, users have to resolve the address and send the input.
Instead, exporter and collector process can take two separate inputs in string format: "host:port" and "proto". This makes things easier if host is DNS name instead of IP.

srikartati:
Instead of accepting resolved TCP/UDP addresses, exporting and
collecting process takes hostPort and protocol string inputs and
do the addressing resolver job internally to make it easier for the
user.

Fixes #121

### DIFF
--- a/pkg/collector/process.go
+++ b/pkg/collector/process.go
@@ -56,7 +56,11 @@ type CollectingProcess struct {
 }
 
 type CollectorInput struct {
-	Address       net.Addr
+	// Address needs to be provided in hostIP:port format.
+	Address string
+	// Protocol needs to be provided in lower case format.
+	// We support "tcp" and "udp" protocols.
+	Protocol      string
 	MaxBufferSize uint16
 	TemplateTTL   uint32
 	IsEncrypted   bool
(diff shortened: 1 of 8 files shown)


pkg/collector/process.go
```diff
@@ -72,11 +73,23 @@ type clientHandler struct {
...
 
 func InitCollectingProcess(input CollectorInput) (*CollectingProcess, error) {
+	var address net.Addr
+	var err error
+	if input.Protocol == "tcp" {
+		address, err = net.ResolveTCPAddr(input.Protocol, input.HostPort)
+	} else if input.Protocol == "udp" {
+		address, err = net.ResolveUDPAddr(input.Protocol, input.HostPort)
+	} else {
+		return nil, fmt.Errorf("collecting process with protocol %s is not syupported", input.Protocol)
```
zyiou, line 83: syupported -> supported

pkg/collector/process.go
```diff
@@ -56,7 +56,8 @@ type CollectingProcess struct {
 }
 
 type CollectorInput struct {
-	Address       net.Addr
+	HostPort      string
+	Protocol      string
```
zyiou, line 60: We can add comments here to specify the format of HostPort and Protocol. 
Also, the name `HostPort` is a little ambiguous to me, I would wonder whether there is a hostIP for the hostPort. Maybe we can keep the Address or Add as Antrea side is using `externalFlowCollectorAddr` and `collectorAddr`.

pkg/exporter/process.go
```diff
@@ -53,7 +53,8 @@ type ExportingProcess struct {
 }
 
 type ExporterInput struct {
-	CollectorAddr       net.Addr
+	CollectorHostPort   string
+	CollectorTransport  string
```
zyiou, line 57: Same as above.
out-of-curiosity, why do we change the name to Transport instead of Protocol?

pkg/collector/process.go
srikartati: Followed the format in net.JoinHostPort method. I can see how this can be confusing, so changed the name to Address.

pkg/exporter/process.go
srikartati: Changed to Protocol to be consistent.

zyiou (approved):
LGTM

PR #132, opened 2021-02-04
Add log info for debugging

zyiou:
Log the error values in Encode function to assist debugging.

### DIFF
--- a/pkg/intermediate/aggregate.go
+++ b/pkg/intermediate/aggregate.go
@@ -237,7 +237,7 @@ func (a *AggregationProcess) correlateRecords(incomingRecord, existingRecord ent
 				if ieWithValue.Value != "" {
 					existingIeWithValue, _ := existingRecord.GetInfoElementWithValue(field)
 					if existingIeWithValue.Value != "" {
-						klog.Warningf("This field with name %v should not have been filled with value %v in existing record.", field, existingIeWithValue.Value)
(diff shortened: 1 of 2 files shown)


pkg/intermediate/aggregate.go
```diff
@@ -237,15 +237,15 @@ func (a *AggregationProcess) correlateRecords(incomingRecord, existingRecord ent
 				if ieWithValue.Value != "" {
 					existingIeWithValue, _ := existingRecord.GetInfoElementWithValue(field)
 					if existingIeWithValue.Value != "" {
-						klog.Warningf("This field with name %v should not have been filled with value %v in existing record.", field, existingIeWithValue.Value)
+						klog.Warningf("This field with name %v and value %v should not have been filled with value %v in existing record.", field, ieWithValue.Value, existingIeWithValue.Value)
```
srikartati, line 240: This log is not clear. If we need current value, lets do this:
" %v field should not have been filled in the existing record; existing value: %v and current value: %v"

srikartati (commented):
A comment on the clarity of logs.

pkg/intermediate/aggregate.go
zyiou: Done. Thanks!

srikartati (approved):
LGTM.

PR #133, opened 2021-02-06
Changes for release v0.4.3

zyiou:
Cherry-picked everything from release v0.4.3. 
The only thing to pay attention is that the length-delimited way is a little different from the previous one: I changed to use a uint64 number as prefix because I cannot figure out how the Encode is doing for length-delimiter. Let me know whether current solution works for you @srikartati .


srikartati:
> Cherry-picked everything from release v0.4.3.
> The only thing to pay attention is that the length-delimited way is a little different from the previous one: I changed to use a uint64 number as prefix because I cannot figure out how the Encode is doing for length-delimiter. Let me know whether current solution works for you @srikartati .

Thanks for taking care of the length prefix in Kafka message when length delimited is enabled. I see 32bits being typically used considering the limits of protobuf messaging. 
https://medium.com/@seb.nyberg/length-delimited-protobuf-streams-a39ebc4a4565#:~:text=The%20official%20recommendation%20is%20to,are%20there%20for%20a%20reason.

zyiou:
> > Cherry-picked everything from release v0.4.3.
> > The only thing to pay attention is that the length-delimited way is a little different from the previous one: I changed to use a uint64 number as prefix because I cannot figure out how the Encode is doing for length-delimiter. Let me know whether current solution works for you @srikartati .
> 
> Thanks for taking care of the length prefix in Kafka message when length delimited is enabled. I see 32bits being typically used considering the limits of protobuf messaging.
> https://medium.com/@seb.nyberg/length-delimited-protobuf-streams-a39ebc4a4565#:~:text=The%20official%20recommendation%20is%20to,are%20there%20for%20a%20reason.

Thanks for the info. Updated with uint32 length-prefixed encoding support.

srikartati (approved):
LGTM

PR #134, opened 2021-02-11
Modify address resolution method in exporter and collector

zyiou:
Add DNS resolution in exporter and collector, remove the resolution for tls address because tls will use DNS name for validating certificates.

### DIFF
--- a/pkg/collector/process.go
+++ b/pkg/collector/process.go
@@ -68,6 +68,7 @@ type CollectorInput struct {
 	CACert     []byte
 	ServerCert []byte
 	ServerKey  []byte
+	IsIPv6     bool
 }
 
 type clientHandler struct {
@@ -78,10 +79,14 @@ type clientHandler struct {
 func InitCollectingProcess(input CollectorInput) (*CollectingProcess, error) {
 	var address net.Addr
 	var err error
+	collectorAddress, err := util.ResolveDNSAddress(input.Address, input.IsIPv6)
+	if err != nil {
+		return nil, err
+	}
 	if input.Protocol == "tcp" {
-		address, err = net.ResolveTCPAddr(input.Protocol, input.Address)
+		address, err = net.ResolveTCPAddr(input.Protocol, collectorAddress)
 	} else if input.Protocol == "udp" {
-		address, err = net.ResolveUDPAddr(input.Protocol, input.Address)
+		address, err = net.ResolveUDPAddr(input.Protocol, collectorAddress)
 	} else {
 		return nil, fmt.Errorf("collecting process with protocol %s is not supported", input.Protocol)
 	}
--- a/pkg/exporter/process.go
+++ b/pkg/exporter/process.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/klog"
 
 	"github.com/vmware/go-ipfix/pkg/entities"
+	"github.com/vmware/go-ipfix/pkg/util"
 )
 
 const startTemplateID uint16 = 255
@@ -65,6 +66,7 @@ type ExporterInput struct {
 	CACert              []byte
 	ClientCert          []byte
 	ClientKey           []byte
+	IsIPv6              bool
 }
 
 // InitExportingProcess takes in collector address(net.Addr format), obsID(observation ID)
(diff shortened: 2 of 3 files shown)


pkg/collector/process.go
```diff
@@ -78,10 +78,15 @@ type clientHandler struct {
 func InitCollectingProcess(input CollectorInput) (*CollectingProcess, error) {
 	var address net.Addr
 	var err error
+	klog.Info(input.Address)
```
srikartati, line 81: debug log?

pkg/util/util.go
```diff
@@ -50,3 +52,26 @@ func Decode(buffer io.Reader, byteOrder binary.ByteOrder, outputs ...interface{}
 	}
 	return nil
 }
+
+// ResolveDNSAddress gets ip:port or dns:port as input and resolves the dns name if exists
```
srikartati, line 56: >ResolveDNSAddress takes address string in the format of host:port as the input; "host" can be either IP or DNS name. If a DNS name is provided, it resolves the IP address. Its output is always in the format of IP:port. 

Do we need to do this IPv6 parsing here? Dont we do this in Antrea?

pkg/exporter/process.go
```diff
@@ -67,6 +68,17 @@ type ExporterInput struct {
 	ClientKey           []byte
 }
 
+type tlsAddress struct {
```
srikartati, line 71: do you need this?

pkg/exporter/process.go
```diff
@@ -83,13 +95,21 @@ func InitExportingProcess(input ExporterInput) (*ExportingProcess, error) {
...
+	collectorAddress, err := util.ResolveDNSAddress(input.CollectorAddress)
+	if err != nil {
+		return nil, err
+	}
 	if input.CollectorProtocol == "tcp" {
-		collectorAddr, err = net.ResolveTCPAddr(input.CollectorProtocol, input.CollectorAddress)
-		if err != nil {
-			return nil, err
+		if input.IsEncrypted {
+			collectorAddr = &tlsAddress{input.CollectorAddress}
```
srikartati, line 104: do you need this if block? Just use input.CollectorAddress directly in TLS dial in L124? 
If encryption is required, add a check to make sure input.CollectorAddress is not IP and if not, throw an error.

pkg/util/util.go
zyiou: We can resolve the DNS name here and remove the code https://github.com/vmware-tanzu/antrea/blob/main/pkg/agent/flowexporter/exporter/exporter.go#L178 as you mentioned.

pkg/exporter/process.go
zyiou: Actually we don't need this block and tlsAddress above. Removed and changed L124. Thanks!

pkg/collector/process.go
zyiou: Removed. Thanks!

pkg/exporter/process.go
```diff
@@ -101,7 +107,7 @@ func InitExportingProcess(input ExporterInput) (*ExportingProcess, error) {
 			if configErr != nil {
 				return nil, configErr
 			}
-			conn, err = tls.Dial(collectorAddr.Network(), collectorAddr.String(), config)
+			conn, err = tls.Dial(input.CollectorProtocol, input.CollectorAddress, config)
```
srikartati, line 110: Exporting process should always expect a DNS name for collectorAddress for TLS or is IP a possible input too?

zyiou: IP is possible, but for Antrea we support DNS name only. and the DNS name will be passed as a parameter in flow aggregator.

pkg/exporter/process.go
```diff
@@ -83,13 +85,17 @@ func InitExportingProcess(input ExporterInput) (*ExportingProcess, error) {
 	var err error
 
 	// Resolve TCP and UDP addresses
```
srikartati, line 87: Change the comment to DNS address

pkg/exporter/process.go
srikartati: Add a comment here we take the input address directly without resolving any DNS name.

srikartati (approved):
LGTM

pkg/exporter/process.go
antoninbas: > IP is possible, but for Antrea we support DNS name only. and the DNS name will be passed as a parameter in flow aggregator.

That's not accurate I think. Someone could choose to provide an IP address, in which case the certificate would be generated using the IP address as the DNS name. This is a valid use case and it should work.

zyiou: Agree. Since we cannot get the flow aggregator IP address currently in Antrea, I only use DNS name. In the future we can support custom fields (IPs or DNSs) to generate certificate.

PR #136, opened 2021-02-11
Fix standalone collector issue

zyiou:
Standalone collector takes empty string as address input, which causes the ResolveDNSAddress to throw an error.


srikartati (approved):
LGTM

PR #139, opened 2021-02-16
Add methods to Set Interface

srikartati:
Add PrepareSet and ResetSet methods to Set interface.
This is to avoid multiple allocations by the user, for example, when
exporting sets user could allocate the Set only once and re-use it
for exporting multiple records.

PS: In Antrea, the main user of go-ipfix library, this change enables to write better unit tests.

### DIFF
--- a/pkg/entities/set.go
+++ b/pkg/entities/set.go
@@ -41,7 +41,8 @@ const (
 )
 
 type Set interface {
-	GetBuffLen() int
+	PrepareSet(setType ContentType, templateID uint16) error
+	ResetSet()
 	GetBuffer() *bytes.Buffer
 	GetSetType() ContentType
 	UpdateLenInHeader()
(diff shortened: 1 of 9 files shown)


pkg/entities/set.go
```diff
@@ -58,22 +59,27 @@ type set struct {
...
+	return nil
 }
 
-func (s *set) GetBuffLen() int {
-	return s.buffer.Len()
+func (s *set) ResetSet() {
+	s.buffer.Reset()
+	s.setType = Undefined
+	s.records = s.records[:0]
 }
```
zyiou, line 83: Since we will probably encounter `Undefined` for setType, we should check whether setType is `Undefined` in AddRecords function to avoid adding record to invalid set.

pkg/entities/set.go
```diff
@@ -58,22 +59,27 @@ type set struct {
...
+	return &set{
 		buffer:     &bytes.Buffer{},
 		setType:    setType,
 		records:    make([]Record, 0),
 		isDecoding: isDecoding,
 	}
-	if !isDecoding {
+}
+
+func (s *set) PrepareSet(templateID uint16) error {
```
zyiou, line 71: Is it possible that we `PrepareSet` after `ResetSet`? I think this is a valid use case for us to re-use the set. If so, we can consider moving the `setType` parameter to `PrepareSet` otherwise the `setType` will be `Undefined`

srikartati: Agreed it is good to have these checks. Yes, we can reuse the set for both data and template types instead of allocating again. 
Added those checks.

zyiou: I think the problem is still there -- if we call `ResetSet` and then want to use `PrepareSet`, the setType will always be `Undefined`.

zyiou: And it's better to add another check [here](https://github.com/vmware/go-ipfix/blob/5993fb33828e98810e63f3aa8e1cab451783624e/pkg/exporter/process.go#L179) so that `Undefined` setType will not go through this function.

srikartati:
Originally I was thinking of using separate sets for data and template, so did not take care of set type.
It makes sense to use one set and just have separate ones for encoding and decoding. 

Made the changes accordingly and addressed comments too. 
Thanks for the comments @zyiou

zyiou (approved):
Thanks for addressing the comments. LGTM

PR #140, opened 2021-02-16
Simplify address input consumption

srikartati:
In exporting and collecting process, simplify the address input
consumption.

This is as per suggestions provided through Antrea PRs, who is the main user of go-ipfix library.


zyiou (approved):
LGTM

zyiou (approved):

PR #148, opened 2021-03-09
Add new field flowType in Antrea registry

srikartati:
Modify aggregate.go to take in the flowType from the record
and do correlation only for records with InterNode flowType.

This is to support Pod-To-External flows in Antrea: https://github.com/vmware-tanzu/antrea/issues/1925

This PR also has refactored the integration tests quite a bit. This is to cleanup the current clunky exporter_collector_test.go and also to simplify the process of generating raw bytes for required packets in
collector_intermediate_test.go


pkg/intermediate/aggregate.go
```diff
@@ -60,8 +60,9 @@ type AggregationInput struct {
 	AggregateElements *AggregationElements
 }
 
-// InitAggregationProcess takes in message channel (e.g. from collector) as input channel, workerNum(number of workers to process message)
-// and correlateFields (fields to be correlated and filled).
+// InitAggregationProcess takes in message channel (e.g. from collector) as input
+// channel, workerNum(number of workers to process message), correlateFlows(flows
+// to be correlated) bitmask and correlateFields (fields to be correlated and filled).
```
zyiou, line 65: what does `bitmask` mean here?

srikartati (commented):
I removed that input. Let me clean up the comment. Missed it.

pkg/intermediate/aggregate.go
```diff
@@ -168,9 +169,23 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
 	a.mutex.Lock()
 	defer a.mutex.Unlock()
 
+	correlationRequired := false
+	// Get Antrea flowType from the record that will help to determine if correlation
+	// is required or not.
+	if ieWithValue, exist := record.GetInfoElementWithValue("flowType"); exist {
+		if ieWithValue.Element.DataType == entities.Unsigned8 {
+			recordFlowType := ieWithValue.Value.(uint8)
```
zyiou, line 177: These two lines seem redundant. Why not use `if recordFlowType, ok := ieWithValue.Value.(uint8); ok {}`

pkg/intermediate/aggregate.go
```diff
@@ -168,9 +169,23 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
...
+			// Correlation is required for only InterNode flow type defined in
+			// pkg/registry/registry.go.
+			if recordFlowType == uint8(registry.InterNode) {
```
zyiou, line 180: Maybe we can define AntreaFlowType constants to be like `registry.InterNode AntreaFlowType = uint8(1)` to avoid conversion here.

pkg/intermediate/aggregate.go
```diff
@@ -228,7 +242,8 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
 }
 
 // correlateRecords correlate the incomingRecord with existingRecord using correlation
-// fields.
+// fields. This is called for records whose flowType is Internode(AntreFlowType
```
zyiou, line 245: Internode -> InterNode

zyiou (commented):
Thanks for taking care of improving integration tests. Two questions:
1. Shall we add Pod-to-External test cases in either unit or integration tests?
2. out of curiosity, how would we assign the flowType at Antrea side?

pkg/intermediate/aggregate.go
srikartati: Yes, that should be simpler.

pkg/intermediate/aggregate.go
srikartati: Removed the type AntreaFlowType completely to make it simpler.

srikartati:
> Thanks for taking care of improving integration tests. Two questions:
> 
> 1. Shall we add Pod-to-External test cases in either unit or integration tests?

Added in unit tests. Originally thought it could be covered in Antrea.

> 2. out of curiosity, how would we assign the flowType at Antrea side?

We need to match either the IPs with appropriate CIDRs or check for appropriate connection mark and figure out the correct type.

pkg/registry/registry.go
```diff
@@ -30,6 +30,14 @@ const (
 	IANAReversedEnterpriseID uint32 = 29305
 )
 
+// enum for flowType field in Antrea registry.
+const (
+	IntraNode    = uint8(1)
+	InterNode    = uint8(2)
+	ToExternal   = uint8(3)
+	FromExternal = uint(4)
```
zyiou, line 38: Why `FromExternal` type is different from others?

srikartati: It was a typo. Fixed.

zyiou (approved):

PR #150, opened 2021-03-11
Modify aggregation process to consume flow end reason

srikartati:
This is to add "flowEndReason" field in nonStatsElement list to update the field in the aggregated record as part of Aggregation process.

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -38,6 +38,14 @@ const (
 	FromExternal = uint8(4)
 )
 
+// enum for flowEndReason field in IANA registry.
+// List of RFC supported reasons: https://www.iana.org/assignments/ipfix/ipfix.xhtml#ipfix-flow-end-reason
+const (
+	IdleTimeoutReason   = uint8(0x01)
+	ActiveTimeoutReason = uint8(0x02)
+	EndOfFlowReason     = uint8(0x03)
+)
+
 var (
 	// globalRegistryByID shows mapping EnterpriseID -> Info Element ID -> Info Element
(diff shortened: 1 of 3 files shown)


pkg/registry/registry.go
```diff
@@ -30,6 +30,13 @@ const (
 	IANAReversedEnterpriseID uint32 = 29305
 )
 
+// enum for flowEndReason field in IANA registry.
+const (
+	IdleTimeoutReason   = uint8(0x01)
+	ActiveTimeoutReason = uint8(0x02)
+	EndOfFlowReason     = uint8(0x03)
+)
```
zyiou, line 38: We can support all the reasons here in case of future need.
https://www.iana.org/assignments/ipfix/ipfix.xhtml#ipfix-flow-end-reason

zyiou: What is the default value for flowEndReason? If it is (0x00) we can have a const here too to be used in Antrea.

srikartati: It is a reserved value. Where do you think we can use that in Antrea? For all flow records, the above three reasons are sufficient.

zyiou: I'm thinking that for flows that are not terminated, what value should be assigned?

srikartati: For flows, which are not terminated we keep on sending the records with either activeFlow or idleFlow tags.
Are you seeing any more cases through your work in tcp state PR?

zyiou: Got it. I haven't seen other cases. Just want to make sure.

zyiou (approved):
LGTM

PR #158, opened 2021-03-26
Modify Aggregation process methods

srikartati:
Delete the unrequired method of deleting record from record map without lock.
Add a method to get the flow updated time for the flow given flow key.

This is required for PR in Antrea: https://github.com/vmware-tanzu/antrea/pull/1949


srikartati:
@zyiou Please get to this quickly when you have time. It is a small patch and an Antrea PR is dependent on this.

zyiou (approved):
LGTM

Issue #95, opened 2020-11-30
K8s deployment manifest yaml to deploy standalone collector


srikartati:
We have a container image. For easier consumption, we can add a K8s deployment manifest. In future, parameter could be take in as configMap parameters especially certificate and keys to start collector with TLS/DTLS.

PR #159, opened 2021-04-03
Create K8s deployment yaml to deploy ipfix-collector

heanlan:
1. Add templates and configuration files to generate the collector manifest
2. Add scripts for preparing and uploading release assets
3. Add corresponding guidance in README to deploy collector

Fixes: #95

### DIFF
--- a/.github/workflows/upload_release_assets.yml
+++ b/.github/workflows/upload_release_assets.yml
@@ -9,6 +9,13 @@ jobs:
   build:
     runs-on: [ubuntu-latest]
     steps:
+      - name: Build assets
+        uses: actions/checkout@v2
+        env:
+          TAG: ${{ github.ref }}
+        run: |
+          mkdir assets
+          VERSION="${TAG:10}" ./hack/release/prepare-assets.sh ./assets
       - name: Checkout
         uses: actions/checkout@v2
         with:
@@ -17,15 +24,12 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: 1.15
-      - name: Build collector binary for Linux amd64
-        run: |
-          make collector
-      - name: Upload collector binary
+      - name: Upload ipfix-collector.yaml
         uses: actions/upload-release-asset@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ github.event.release.upload_url }}
-          asset_path: ./bin/collector
-          asset_name: ipfix-collector-linux-x86_64
+          asset_path: ./assets/ipfix-collector.yaml
+          asset_name: ipfix-collector.yaml
           asset_content_type: application/octet-stream
--- a/README.md
+++ b/README.md
@@ -6,6 +6,35 @@ go-ipfix is an IPFIX library that can be used to implement an IPFIX exporter, wh
 ## Try it out
 This IPFIX library can be used to build an exporter. Please check out the [exporter tests](https://github.com/vmware/go-ipfix/blob/main/pkg/exporter/process_test.go) to get an idea on how to build exporter on top of TCP and UDP transport protocols given a IPFIX collector.
 
+### Deploy stand alone IPFIX collector
+To deploy a released version of go-ipfix collector, which is used to decode and log the IPFIX records, please choose one deployment manifest from the list of releases. For any given release <TAG> (e.g. v0.1.0), you can deploy the collector as follows:
+
+```
+kubectl apply -f https://github.com/vmware/go-ipfix/releases/download/<TAG>/ipfix-collector.yaml
+```
+To deploy the latest version of collector (built from the main branch), use the command below: 
+```
+cd <directory containing this README file>/build/yamls
+kubectl apply -f ./ipfix-collector.yaml
+```
+
+While deploying the latest version of collector, the port and protocol can be configured by using the command:
+
+```
+cd <directory containing this README file>/hack
+./generate-manifest-collector.sh --mode dev --port <port> --proto (tcp|udp) > ../build/yamls/ipfix-collector.yaml
+```
+
+Parameter ```--mode dev``` will build the collector from the docker image with the "latest" tag.  
+Use ```--port <port>``` to specify the port used by the collector. Default is 4739.  
+Use  ```--proto (tcp|udp)``` to specify the protocol used by the collector. Default is tcp.  
+For example:
+
+```
+./generate-manifest-collector.sh --mode dev --port 4739 --proto tcp > ../build/yamls/ipfix-collector.yaml
+
+```
+
 ## Build Registry
 To build the registry from [IANA registry](https://www.iana.org/assignments/ipfix/ipfix.xhtml) or [Antrea registry](pkg/registry/registry_antrea.csv), run following commands:
 ```
--- a/README.md
+++ b/README.md
@@ -7,18 +7,18 @@ go-ipfix is an IPFIX library that can be used to implement an IPFIX exporter, wh
 This IPFIX library can be used to build an exporter. Please check out the [exporter tests](https://github.com/vmware/go-ipfix/blob/main/pkg/exporter/process_test.go) to get an idea on how to build exporter on top of TCP and UDP transport protocols given a IPFIX collector.
 
 ### Deploy stand alone IPFIX collector
-To deploy a released version of go-ipfix collector, which is used to decode and log the IPFIX records, please choose one deployment manifest from the list of releases. For any given release <TAG> (e.g. v0.1.0), you can deploy the collector as follows:
+To deploy a released version of the go-ipfix collector, which is used to collect, decode and log the IPFIX records, please choose one deployment manifest from the list of releases. For any given release <TAG> (e.g. v0.1.0), you can deploy the collector as follows:
 
 ```
 kubectl apply -f https://github.com/vmware/go-ipfix/releases/download/<TAG>/ipfix-collector.yaml
 ```
-To deploy the latest version of collector (built from the main branch), use the command below: 
+To deploy the latest version of the go-ipfix collector (built from the main branch), use the command below: 
 ```
 cd <directory containing this README file>/build/yamls
 kubectl apply -f ./ipfix-collector.yaml
 ```
 
-While deploying the latest version of collector, the port and protocol can be configured by using the command:
+While deploying the latest version of the go-ipfix collector, port and protocol can be configured by using the command:
 
 ```
 cd <directory containing this README file>/hack
--- a/build/yamls/collector/Chart.yaml
+++ b/build/yamls/collector/Chart.yaml
@@ -1,3 +1,3 @@
 apiVersion: v2
 name: ipfix-collector
-version: 0.5.0
+version: 0.6.0
--- a/build/yamls/collector/Chart.yaml
+++ b/build/yamls/collector/Chart.yaml
@@ -1,0 +1,3 @@
+apiVersion: v2
+name: ipfix-collector
+version: 0.5.0
--- a/build/yamls/collector/templates/ipfix-collector.yaml
+++ b/build/yamls/collector/templates/ipfix-collector.yaml
@@ -1,0 +1,57 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    app: ipfix-collector
+  name: ipfix
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: ipfix-collector
+  name: ipfix-collector
+  namespace: ipfix
+spec:
+  selector:
+    app: ipfix-collector
+  ports:
+  - name: ipfix-udp
+    port: 4739
+    protocol: UDP
+    targetPort: {{ .Values.port }}
+  - name: ipfix-tcp
+    port: 4739
+    protocol: TCP
+    targetPort: {{ .Values.port }}
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: ipfix-collector
+  name: ipfix-collector
+  namespace: ipfix
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: ipfix-collector
+  template:
+    metadata:
+      labels:
+        app: ipfix-collector
+    spec:
+      containers:
+      - args:
+        - --v=0
+        - --ipfix.port={{ .Values.port }}
+        - --ipfix.transport={{ .Values.protocol }}
+        image: projects.registry.vmware.com/antrea/ipfix-collector:{{ .Values.image_tag }}
+        imagePullPolicy: IfNotPresent
+        name: ipfix-collector
+        ports:
+        - containerPort: {{ .Values.port }}
+      nodeSelector:
+        kubernetes.io/os: linux
+        kubernetes.io/arch: amd64
--- a/hack/generate-manifest-collector.sh
+++ b/hack/generate-manifest-collector.sh
@@ -1,0 +1,118 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 Go-ipfix Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eo pipefail
+
+function echoerr {
+    >&2 echo "$@"
+}
+
+MODE="dev"
+
+_usage="Usage: $0 [--mode (dev|release)] [--help|-h]
+Generate a YAML manifest for ipfix collector, using Helm, and print it to stdout.
+        --mode (dev|release)  Choose the configuration variant that you need (default is 'dev')
+        --help, -h            Print this message and exit
+
+In 'dev' mode, argument [--port <port> --proto (tcp|udp)] can be provided.
+If no argument is given, collector will be configured with the default argument values.
+        --port <port>         Specify the port that will be listened on by the collector. Default is 4739.
+        --proto (tcp|udp)       Speicify the protocol. Default is tcp.
+Example: ./generate-manifest-collector.sh --mode dev --port 4739 --proto tcp > ../build/yamls/ipfix-collector.yaml
+
+In 'release' mode, environment variable IMG_TAG must be set.
+
+This tool uses helm (https://github.com/helm/helm) to generate templated manifests for
+running ipfix collector. You can set the HELM environment variable to the path of the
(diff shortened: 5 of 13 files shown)


README.md
```diff
@@ -6,6 +6,11 @@ go-ipfix is an IPFIX library that can be used to implement an IPFIX exporter, wh
 ## Try it out
 This IPFIX library can be used to build an exporter. Please check out the [exporter tests](https://github.com/vmware/go-ipfix/blob/main/pkg/exporter/process_test.go) to get an idea on how to build exporter on top of TCP and UDP transport protocols given a IPFIX collector.
 
+To deploy a standalone go-ipfix collector, which is used to decode and log the IPFIX records, you can use:
```
srikartati, line 9: Better to have a sub-section like "Deploy Go-IPFIX Collector"

build/yamls/ipfix-collector.yml
```diff
@@ -0,0 +1,47 @@
...
+  selector:
+    matchLabels:
+      app: ipfix-collector
+  template:
+    metadata:
+      labels:
+        app: ipfix-collector
+    spec:
+      containers:
+      - image: projects.registry.vmware.com/antrea/ipfix-collector:v0.4.8
```
srikartati, line 41: how will the versioning work? I think our intention should not be to make users change this file manually.
One option is to use kustomize. We use that in Antrea. Please see this file for reference: https://github.com/vmware-tanzu/antrea/blob/main/hack/generate-manifest.sh
See the mode option with dev or release as possible values.

build/yamls/ipfix-collector.yml
```diff
@@ -0,0 +1,47 @@
...
+metadata:
+  labels:
+    app: ipfix-collector
+  name: ipfix-collector
+  namespace: ipfix
+spec:
+  selector:
+    app: ipfix-collector
+  ports:
+    - protocol: TCP
```
srikartati, line 19: We support UDP too. 
https://github.com/vmware/go-ipfix/blob/main/cmd/collector/collector.go#L60

heanlan: Do we have to add a configmap to support UDP?

README.md
heanlan: Will do. Thanks!

build/yamls/ipfix-collector.yml
heanlan: Thanks. Just to clarify, dev mode uses the "latest" image tag, while release mode uses the release version as image tag. Is that what you expect?

build/yamls/ipfix-collector.yml
heanlan: Do we have to add a configmap to support UDP?

srikartati: Command argument `--ipfix.transport="udp"` in the k8s manifest should work.

build/yamls/ipfix-collector.yml
srikartati: Yes, dev mode for latest tag with ToT main branch and release mode will have a release version image in the manifest.

zyiou: @srikartati How about using a script for replacing `latest` with version when uploading release assets and keep `latest` in dev code? It will have same effect as Antrea. And we can add guide for deploying ipfix collector like [this](https://github.com/vmware-tanzu/antrea/blob/main/docs/getting-started.md#installation). Our yaml file is quite simple for now, we can probably support kustomization later.

build/yamls/ipfix-collector.yml
heanlan: It looks like in order to make it work, I have to modify the Dockerfile.build.collector. 
Line19 is changed from:
`ENTRYPOINT "collector"`
to:
`ENTRYPOINT ["collector"]`
And now collector.go can effectively take arguments.

build/yamls/ipfix-collector.yml
srikartati: Sure a script to replace yaml is ok. I think we may need a few more options in the script like collector address, port, protocol, etc. However, I do not think it will blow up (at least for now) where we require kustomize tool .

It is better to add the protocol editing option along with the image tag option in the script.

build/yamls/ipfix-collector.yml
srikartati: Thanks for trying it out and figuring out the fix.

dreamtalen:
When using ipfix-collector on a cluster with ARM and Windows Nodes, Pods should be scheduled on Linux AMD64
Nodes. Could you please add the nodeSelector in the yaml file like this PR in antrea? https://github.com/vmware-tanzu/antrea/pull/2087

heanlan:
> When using ipfix-collector on a cluster with ARM and Windows Nodes, Pods should be scheduled on Linux AMD64
> Nodes. Could you please add the nodeSelector in the yaml file like this PR in antrea? [vmware-tanzu/antrea#2087](https://github.com/vmware-tanzu/antrea/pull/2087)

heanlan:
Accidentally close the PR, reopen now...

README.md
```diff
@@ -6,6 +6,28 @@ go-ipfix is an IPFIX library that can be used to implement an IPFIX exporter, wh
 ## Try it out
 This IPFIX library can be used to build an exporter. Please check out the [exporter tests](https://github.com/vmware/go-ipfix/blob/main/pkg/exporter/process_test.go) to get an idea on how to build exporter on top of TCP and UDP transport protocols given a IPFIX collector.
 
+### Deploy stand alone IPFIX collector
+To deploy a standalone go-ipfix collector, which is used to decode and log the IPFIX records, you can use:
+```
+cd <directory containing this README file>/build/yamls
+kubectl apply -f ./ipfix-collector.yaml
```
zyiou, line 13: We can follow the way antrea does [here](https://github.com/vmware-tanzu/antrea/blob/main/docs/getting-started.md#installation): specify the release manifest here and if users want to customize the protocol etc., they can download the folder from main branch.

README.md
```diff
@@ -6,6 +6,28 @@ go-ipfix is an IPFIX library that can be used to implement an IPFIX exporter, wh
...
+```
+The collector's port and protocol can be configured by using the command:
+
+```
+cd <directory containing this README file>/hack
+./generate-manifest-collector.sh --mode dev --addr "port:proto" > ../build/yamls/ipfix-collector.yaml
```
zyiou, line 19: I think using separate params here would be better: `--port 4739 --protocol tcp`

build/yamls/collector/Chart.yaml
```diff
@@ -0,0 +1,3 @@
+apiVersion: v2
+name: ipfix-collector
+version: 0.5.0
```
zyiou, line 3: new line.

hack/.bin/darwin-amd64/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
zyiou, line 1: Out of curiosity, are files in `hack/.bin` necessary for us to deploy ipfix-collector?

heanlan: You're right. I shouldn't commit those to the repo. The verify-helm.sh will install Helm locally.

build/yamls/collector/Chart.yaml
heanlan: Fixed. Thanks.

README.md
heanlan: Yes. I was doing this because originally we have three arguments `addr:port:proto`. I feel like four separate arguments (including --mode) might be too long. Then we remove the `addr` arg. Now I'm waiting for Srikar's opinion on whether we should keep the port argument or not.

README.md
heanlan: Absolutely. The only concern I had before is that users will not be able to find any manifests under "releases assets" since we do not have the manifest for previous release. But I agree we can add the guide at this time for future need. Thanks!

build/yamls/collector/templates/ipfix-collector.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: v1
```
srikartati, line 1: Lets use "template" or "base" instead of "templates". There is only one template.

hack/generate-manifest-collector.sh
```diff
@@ -0,0 +1,116 @@
...
+If no argument is given, collector will be configured with the default argument values.
+        --port <port>         Specify the port that will be listened on by the collector. Default is 4739.
+        --proto (tcp|udp)       Speicify the protocol. Default is tcp.
+Example: ./generate-manifest-collector.sh --mode dev --port 4739 --proto tcp > ../build/yamls/ipfix-collector.yaml
+
+In 'release' mode, environment variable IMG_TAG must be set.
+
+This tool uses helm (https://github.com/helm/helm) to generate templated manifests for
+running ipfix collector. You can set the HELM environment variable to the path of the
+helm binary you want us to use. Otherwise we will look for kustomize in your PATH and your
```
srikartati, line 40: kustomize?

srikartati (commented):
I think we need a github job to make sure that the manifest is up-to-date if there is any change to the template. Look for "make manifest" and related github job in Antrea.

.github/workflows/upload_release_assets.yml
```diff
@@ -9,6 +9,13 @@ jobs:
...
     steps:
+      - name: Build assets
+        uses: actions/checkout@v2
+        env:
+          TAG: ${{ github.ref }}
+        run: |
+          mkdir assets
+          VERSION="${TAG:10}" ./hack/release/prepare-assets.sh ./assets
       - name: Checkout
         uses: actions/checkout@v2
```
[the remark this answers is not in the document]
zyiou: yes, we don't need collector binary.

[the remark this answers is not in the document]
heanlan: Removed.

build/yamls/collector/Chart.yaml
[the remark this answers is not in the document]
heanlan: I have added something to update the version number when we have a new release. It will be updated to the most updated release tag. Is that what you are suggesting?

build/yamls/collector/templates/ipfix-collector.yaml
heanlan: I also feel uncomfortable about that. But unfortunately the folder has to be named as "templates" to make it configurable. It has to follow the Helm chart structure as mentioned in the similar issue: https://github.com/helm/helm/issues/4152
I have tried to rename it and the configuration doesn't work as a result. I'm thinking maybe we can still switch back to kustomize in the future if there's a lot of inconvenience.

hack/generate-manifest-collector.sh
heanlan: Fixed. Thanks.

hack/generate-manifest-collector.sh
```diff
@@ -0,0 +1,118 @@
...
+fi
+
+if [ "$MODE" == "release" ]; then
+    # update the Chart version by the newest release version
+    sed -i '' -e "s/^version.*/version: ${IMG_TAG:1}/g" Chart.yaml
+    # replace the line starting with "image_tag", and create a backup
+    sed -i.bak -e "s/^image_tag.*/image_tag: $IMG_TAG/g" values.yaml
+
+    $HELM template "." -f "./templates/ipfix-collector.yaml"
+    cp -f "values.yaml.bak" "values.yaml"
```
srikartati, line 116: do you require this step? Why do we have to keep the values.yml updated?

build/yamls/collector/Chart.yaml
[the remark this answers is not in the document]
srikartati: My understanding is that this is a template and we can use a generic value like <x.x.x>.

Btw, why does the name has to start with capital--"Chart.yaml"?
In addition, I see that Chart.yaml is not at all used when the mode is dev during the time of generation. Is this expected?

hack/generate-manifest-collector.sh
heanlan: These lines will keep values.yaml from being updated. It creates a backup of the original values.yaml, makes changes to the newly-generated values.yaml, generates templated ipfix-collector.yaml using the newly-generated values.yaml, and restore the backup one.

build/yamls/collector/Chart.yaml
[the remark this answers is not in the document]
heanlan: Based on my understanding of the documentation https://helm.sh/docs/topics/charts/, under the "chart and versioning" subtitle, the overall structured files is called a chart, and the version implies the version of the chart. It is used when we apply command like `helm package`, which package a chart directory into a chart archive. So the value is decided by the contents of the current project. There are words like "_Helm v2 and later uses version numbers as release markers. Packages in repositories are identified by name plus version._". I'm not so sure on how should we assign the value of version, but I feel like the value is meaningful.

The naming and existence of Chart.yaml is also part of the Helm chart structure in order to make it configurable. https://helm.sh/docs/chart_template_guide/getting_started/
And it is case sensitive https://helm.sh/docs/chart_best_practices/conventions/#usage-of-the-words-helm-and-chart

heanlan:
> I think we need a github job to make sure that the manifest is up-to-date if there is any change to the template. Look for "make manifest" and related github job in Antrea.

Added.

hack/generate-manifest-collector.sh
```diff
@@ -0,0 +1,140 @@
...
+
+    if [ "$MODE" == "release" ]; then
+        sed -i -e "s/^version.*/version: ${IMG_TAG:1}/g" Chart.yaml
+        sed -i.bak -e "s/^image_tag.*/image_tag: $IMG_TAG/g" values.yaml
+
+        $HELM template "." -f "./templates/ipfix-collector.yaml"
+        cp -f "values.yaml.bak" "values.yaml"
+        rm "values.yaml.bak"
+    fi
+fi
```
heanlan, line 140: I can't figure out a good way to avoid the code repetition here. I tried a couple of ways but they didn't work.
For example introducing a variable:
`SEDOPTION=
if [[ "$OSTYPE" == "darwin"* ]]; then
  SEDOPTION="''"
fi

if [ "$MODE" == "dev" ]; then
    sed -i ${SEDOPTION} -e "s/^image_tag.*/image_tag: latest/g" values.yaml`
This would create an extra weird file named `values''.yaml`.

shihhaoli: I am not a sed expert.
How about setting SEDOPTION="-i" or "-i ''" depending on OSTYPE?

heanlan: Thanks Shih-Hao for the suggestion! I tried it but sadly the results is the same:( still bringing the extra file

shihhaoli: Just curious, is extra file caused by "sed -i.bak" on line 114, 134 or elsewhere?

heanlan: I'm testing it with command `/generate-manifest-collector.sh --mode dev > build/yamls/ipfix-collector.yaml`. 
So I think it goes to line 100, which is the code below with the variable `SEDOPTION="-i ''`:
`sed ${SEDOPTION} -e "s/^image_tag.*/image_tag: latest/g" values.yaml`

I think I can understand the logic that sed will create a file ending with '', just like what I have in line 116, where sed will create a backup file values.yaml.bak. But the strange thing is, without the SEDOPTION, the original code on line 100 works well on mac and will not create the extra file, but with SEDOPTION, it creates the file.

shihhaoli: Thanks for the info. 
One dumb question. So we need to pass "" after -i on windows machine?

heanlan: No, for MacOS we do `sed -i '' -e`, for Linux and Windows we do `sed -i -e`.

shihhaoli: Got it now. Thanks.
Maybe we can always use a dummy backup file for all the sed commands (e.g. "sed -i.dummy ..."), except "sed -i.bak ...".
Then in the end, remove the dummy file.

heanlan: Good idea! Thanks a lot. It works now.

hack/generate-manifest-collector.sh
srikartati: Sounds good.

README.md
```diff
@@ -6,6 +6,35 @@ go-ipfix is an IPFIX library that can be used to implement an IPFIX exporter, wh
 ## Try it out
 This IPFIX library can be used to build an exporter. Please check out the [exporter tests](https://github.com/vmware/go-ipfix/blob/main/pkg/exporter/process_test.go) to get an idea on how to build exporter on top of TCP and UDP transport protocols given a IPFIX collector.
 
+### Deploy stand alone IPFIX collector
+To deploy a released version of go-ipfix collector, which is used to decode and log the IPFIX records, please choose one deployment manifest from the list of releases. For any given release <TAG> (e.g. v0.1.0), you can deploy the collector as follows:
```
srikartati, line 10: "...the go-ipfix collector, which is used to collect, decode and log the IPFIX records.."

README.md
```diff
@@ -6,6 +6,35 @@ go-ipfix is an IPFIX library that can be used to implement an IPFIX exporter, wh
...
+
+```
+kubectl apply -f https://github.com/vmware/go-ipfix/releases/download/<TAG>/ipfix-collector.yaml
+```
+To deploy the latest version of collector (built from the main branch), use the command below:
```
srikartati, line 15: ..of the go-ipfix collector..

README.md
```diff
@@ -6,6 +6,35 @@ go-ipfix is an IPFIX library that can be used to implement an IPFIX exporter, wh
...
+```
+kubectl apply -f https://github.com/vmware/go-ipfix/releases/download/<TAG>/ipfix-collector.yaml
+```
+To deploy the latest version of collector (built from the main branch), use the command below: 
+```
+cd <directory containing this README file>/build/yamls
+kubectl apply -f ./ipfix-collector.yaml
+```
+
+While deploying the latest version of collector, the port and protocol can be configured by using the command:
```
srikartati, line 21: the go-ipfix collector, port and protocol...

build/yamls/collector/Chart.yaml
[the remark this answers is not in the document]
heanlan: Sure. Two questions: 
1. I'm assuming you are saying we will edit the file `VERSION` and `Chart.yaml` in the generate-manifest script. From where shall we get the version value(`0.6.0`) to put in `VERSION`? Just do release version +1? Like 0.5.0 -> 0.6.0?
2. When shall we update the version value? Every time we have a new release we also upgrade the dev version by 0.1.0?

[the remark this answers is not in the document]
srikartati: No, I meant you should edit the Chart.yaml with the version from https://github.com/vmware/go-ipfix/blob/main/VERSION when you are generating the collector manifest using helm tool. Isn't that possible?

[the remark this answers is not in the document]
heanlan: That's definitely possible. I'll do that. I thought you want the VERSION file also to be edited in the generate-manifest script.

srikartati (approved):
Thanks for adding this. LGTM.

srikartati:
I see all comments are addressed. Merging now.

PR #163, opened 2021-04-15
Fix the Set reset issue

srikartati:
Re-initialize record slice by making it nil for GC to clean
up previous records and references properly.
This led to the issue in Antrea: https://github.com/vmware-tanzu/antrea/issues/2035


zyiou (approved):
LGTM

PR #165, opened 2021-04-15
Add new fields for Antrea registry

zyiou:
Add some information elements to Antrea registry to prepare support for tcp state transition and deny connection tracking in Antrea.

### DIFF
--- a/pkg/registry/registry_antrea.csv
+++ b/pkg/registry/registry_antrea.csv
@@ -36,4 +36,7 @@ ElementID,Name,Abstract Data Type,Data Type Semantics,Status,Description,Units,R
 134,reversePacketDeltaCountFromDestinationNode,unsigned64,,current,,,,,,,,56506,
 135,reverseOctetDeltaCountFromDestinationNode,unsigned64,,current,,,,,,,,56506,
 136,tcpState,string,,current,,,,,,,,56506,
-137,flowType,unsigned8,,current,,,,,,,,56506,
+137,flowType,unsigned8,,current,The type of flow is based on the location of the source and destination Pods. Supported Types(uint8 value): IntraNode(1) InterNode(2) ToExternal(3) and FromExternal(4),,,,,,,56506,
+138,tcpStatePrevList,string,,current,,,,,,,,56506,
+139,ingressNetworkPolicyRuleAction,unsigned8,,current,Supported Actions(uint8 value): NetworkPolicyRuleActionAllow(1) NetworkPolicyRuleActionDrop(2) NetworkPolicyRuleActionReject(3),,,,,,,56506,
+140,egressNetworkPolicyRuleAction,unsigned8,,current,Supported Actions(uint8 value): NetworkPolicyRuleActionAllow(1) NetworkPolicyRuleActionDrop(2) NetworkPolicyRuleActionReject(3),,,,,,,56506,
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -38,6 +38,13 @@ const (
 	FromExternal = uint8(4)
 )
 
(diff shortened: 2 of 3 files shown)


pkg/registry/registry_antrea.csv
```diff
@@ -36,4 +36,8 @@ ElementID,Name,Abstract Data Type,Data Type Semantics,Status,Description,Units,R
 134,reversePacketDeltaCountFromDestinationNode,unsigned64,,current,,,,,,,,56506,
 135,reverseOctetDeltaCountFromDestinationNode,unsigned64,,current,,,,,,,,56506,
 136,tcpState,string,,current,,,,,,,,56506,
-137,flowType,unsigned8,,current,,,,,,,,56506,
+137,flowType,unsigned8,,current,Type of flow based on location of source and destination. 1-IntraNode 2-InterNode 3-ToExternal 4-FromExternal,,,,,,,56506,
```
srikartati, line 39: "The type of flow is based on the location of the source and destination Pods. Supported Types(uint8 value): IntraNode(1), InterNode(2), ToExternal(3) and FromExternal(4)"

pkg/registry/registry_antrea.csv
```diff
@@ -36,4 +36,8 @@ ElementID,Name,Abstract Data Type,Data Type Semantics,Status,Description,Units,R
...
+138,tcpStatePrev,string,,current,,,,,,,,56506,
+139,tcpStatePrevList,string,,current,,,,,,,,56506,
```
srikartati, line 41: why both? If there is only one state, it can still be part of the list.

pkg/registry/registry_antrea.csv
```diff
@@ -36,4 +36,8 @@ ElementID,Name,Abstract Data Type,Data Type Semantics,Status,Description,Units,R
...
+140,ingressNetworkPolicyRuleAction,string,,current,,,,,,,,56506,
```
srikartati, line 42: should we do uint8 instead of string? There are only limited action types.

zyiou: I saw in Antrea they are using string, but you are right there are only 3 types. Are you suggesting using uin8 will help reduce length of our flow record?

srikartati: I think this is still pending.

pkg/registry/registry_antrea.csv
zyiou: Removed `tcpStatePrev` since we can keep states in `tcpStatePrevList` in ordered list.

pkg/registry/registry_antrea.csv
```diff
@@ -36,4 +36,7 @@ ElementID,Name,Abstract Data Type,Data Type Semantics,Status,Description,Units,R
 134,reversePacketDeltaCountFromDestinationNode,unsigned64,,current,,,,,,,,56506,
 135,reverseOctetDeltaCountFromDestinationNode,unsigned64,,current,,,,,,,,56506,
 136,tcpState,string,,current,,,,,,,,56506,
-137,flowType,unsigned8,,current,,,,,,,,56506,
+137,flowType,unsigned8,,current,The type of flow is based on the location of the source and destination Pods. Supported Types(uint8 value): IntraNode(1) InterNode(2) ToExternal(3) and FromExternal(4),,,,,,,56506,
+138,tcpStatePrevList,string,,current,,,,,,,,56506,
+139,ingressNetworkPolicyRuleAction,unsigned8,,current,,,,,,,,56506,
```
srikartati, line 41: why not description here?

zyiou: updated. Thanks!

srikartati (approved):
LGTM

PR #172, opened 2021-04-17
Revert "Modify Aggregation process methods (#158)"

srikartati:
Reverting this commit 7c0a0ae664f3720eee1d078598e7c825956f3df1.

Reason is that the PR(vmware-tanzu/antrea#1949) that depends on this commit
is not yet ready. And to move the tag of go-ipfix in Antrea for other
PRs this commit should be present in the latest release tags. Therefore,
reverting this.


srikartati:
@zyiou Realized that this has to be checked in before we do the release v0.5.0, other wise we cannot move the tag in Antrea.

zyiou (approved):

PR #178, opened 2021-04-29
Add Kafka TLS support to the producer.

srikartati:
Also, enhance the input consumption when initializing the
Kafka producer.


pkg/producer/kafka.go
```diff
@@ -26,39 +31,67 @@ import (
...
+	if input.KafkaTLSEnabled {
+		if tlsConfig := setupTLSConfig(input.KafkaCAFile, input.KafkaTLSCertFile, input.KafkaTLSKeyFile); tlsConfig != nil {
+			kafkaConfig.Net.TLS.Config = tlsConfig
+			kafkaConfig.Net.TLS.Enable = true
+			if input.KafkaTLSSkipVerify {
+				klog.Info("kafka client TLS enabled (server certificate didn't validate)")
+			} else {
+				klog.Info("kafka client TLS enabled")
+			}
+		}
```
zyiou, line 83: I can see that errors are already handled at `setupTLSConfig` method. Then is it possible that `tlsConfig` is nil? if so, we may also need to have error handling here.

pkg/producer/kafka.go
```diff
@@ -26,39 +31,67 @@ import (
...
+	kafkaConfig.Producer.Return.Successes = input.KafkaLogSuccesses
+	kafkaConfig.Producer.Return.Errors = input.KafkaLogErrors
 
-	asyncProducer, err := sarama.NewAsyncProducer(addrs, kafkaConfig)
+	// Initialize TLS certificates
+	if input.KafkaTLSEnabled {
+		if tlsConfig := setupTLSConfig(input.KafkaCAFile, input.KafkaTLSCertFile, input.KafkaTLSKeyFile); tlsConfig != nil {
+			kafkaConfig.Net.TLS.Config = tlsConfig
+			kafkaConfig.Net.TLS.Enable = true
+			if input.KafkaTLSSkipVerify {
```
zyiou, line 78: Seems that `input.KafkaTLSSkipVerify` is not used other places other than here. Do we need to handle the case when `input.KafkaTLSSkipVerify` is set?

pkg/producer/kafka.go
```diff
@@ -26,39 +31,67 @@ import (
...
+				klog.Info("kafka client TLS enabled (server certificate didn't validate)")
+			} else {
+				klog.Info("kafka client TLS enabled")
+			}
+		}
+	}
+	if input.EnableSaramaDebugLog {
+		sarama.Logger = log.New(os.Stderr, "[sarama] ", log.LstdFlags)
+	}
+	asyncProducer, err := sarama.NewAsyncProducer(input.KafkaBrokers, kafkaConfig)
```
zyiou, line 88: As long as we put TLS related config into kafkaConfig, Kafka producer will create the tls connection, right?

pkg/producer/kafka.go
srikartati: Agree. Changed the error handling, so it can be propagated to the caller.

pkg/producer/kafka.go
zyiou: Got it. We can add code to ignore secure checking with proper reasoning description like:
```
// #nosec G402: only applicable for testing purpose
t.InsecureSkipVerify = skipVerify
```

zyiou (dismissed):
Are we planning to add tests for this PR? Otherwise LGTM overall.

srikartati:
> Are we planning to add tests for this PR? Otherwise LGTM overall.

I do not have tests for InitKafkaProducer function. I think TLS certificates warrant some unit tests. Will add those.

pkg/producer/kafka.go
shihhaoli: Why didn't we pass input.KafkaTLSSkipVerify to setupTLSConfig(), which returns a tls.Config object?
I thought that value can be directly assigned to the returning tls.Config, such as 
tls.Config{InsecureSkipVerify: input.KafkaTLSSkipVerify} ?

srikartati: golang lint test fails if we directly the value like above. However, I moved the assignment to `setupTLSConfig`.

Sorry for late response. I had put this PR on hold. Please take a look again.

pkg/producer/kafka_test.go
```diff
@@ -0,0 +1,102 @@
+// Copyright 2020 VMware, Inc.
```
zyiou, line 1: 2020 -> 2021

pkg/test/certs.go
```diff
@@ -0,0 +1,134 @@
+package test
```
zyiou, line 1: missing license header in this file

pkg/collector/process_test.go
```diff
@@ -492,11 +363,11 @@ func TestTLSCollectingProcess(t *testing.T) {
 	collectorAddr := cp.GetAddress()
 	go func() {
 		roots := x509.NewCertPool()
-		ok := roots.AppendCertsFromPEM([]byte(fakeCACert))
+		ok := roots.AppendCertsFromPEM([]byte(test.FakeCACert))
```
zyiou, line 366: We can also replace TLS related certs like this file in `pkg/exporter/process_test.go` and `pkg/test/exporter_collector_test.go`

srikartati: done

zyiou (approved):
Thanks a lot for the changes. LGTM

PR #179, opened 2021-05-01
Add and modify network policy related fields in Antrea registry

zyiou:


srikartati (commented):
What about elements for these: network policy UUID, NetworkPolicy type, and rule priority?

zyiou:
> What about elements for these: network policy UUID, NetworkPolicy type, and rule priority?

We have already had elements for these: [ElementID 114-119](https://github.com/vmware/go-ipfix/blob/9d613db9caf0690c20efa1703a3fdc7b3a4c6cab/pkg/registry/registry_antrea.csv#L16-L21) in Antrea registry

srikartati:
> > What about elements for these: network policy UUID, NetworkPolicy type, and rule priority?
> 
> We have already had elements for these: [ElementID 114-119](https://github.com/vmware/go-ipfix/blob/9d613db9caf0690c20efa1703a3fdc7b3a4c6cab/pkg/registry/registry_antrea.csv#L16-L21) in Antrea registry

Cool.. forgot that I added. Can we change the policy type element from string to unisigned8 and define the types like flowTypes and other elements? 
Also lets take this opportunity to review the Antrea registry if there any unnecessary elements.

heanlan (commented):
Thanks for adding the fields. LGTM

zyiou:
> Cool.. forgot that I added. Can we change the policy type element from string to unisigned8 and define the types like flowTypes and other elements?
> Also lets take this opportunity to review the Antrea registry if there any unnecessary elements.

Sure. Updated. 
Also we checked with Yang about network policy rule priority. There are basically three types of priority in ANP/CNP: tier priority, policy priority and rule priority. The network policy rule priority we will assign for NetworkPolicyRulePriority here is from `PolicyRule` struct ([link](https://github.com/vmware-tanzu/antrea/blob/main/pkg/agent/types/networkpolicy.go#L72-L85)), which is actually the computed value from these three priorities (calculated [here](https://github.com/vmware-tanzu/antrea/blob/main/pkg/agent/controller/networkpolicy/priority.go#L110-L128)). This value is like internal indicator for ranking priority (bigger value is of higher priority). To avoid causing misunderstandings, I changed type of `ingressNetworkPolicyRulePriority` and `egressNetworkPolicyRulePriority` and added a `-1` value for K8s network policy which don't have this field.

srikartati:
> > Cool.. forgot that I added. Can we change the policy type element from string to unisigned8 and define the types like flowTypes and other elements?
> > Also lets take this opportunity to review the Antrea registry if there any unnecessary elements.
> 
> Sure. Updated.
> Also we checked with Yang about network policy rule priority. There are basically three types of priority in ANP/CNP: tier priority, policy priority and rule priority. The network policy rule priority we will assign for NetworkPolicyRulePriority here is from `PolicyRule` struct ([link](https://github.com/vmware-tanzu/antrea/blob/main/pkg/agent/types/networkpolicy.go#L72-L85)), which is actually the computed value from these three priorities (calculated [here](https://github.com/vmware-tanzu/antrea/blob/main/pkg/agent/controller/networkpolicy/priority.go#L110-L128)). This value is like internal indicator for ranking priority (bigger value is of higher priority). To avoid causing misunderstandings, I changed type of `ingressNetworkPolicyRulePriority` and `egressNetworkPolicyRulePriority` and added a `-1` value for K8s network policy which don't have this field.

Interesting. Thanks for checking. Changing the name sounds good to me. 
What about other priorities? Are they useful for the user? Should we add those as different fields?

zyiou:
> Interesting. Thanks for checking. Changing the name sounds good to me.
> What about other priorities? Are they useful for the user? Should we add those as different fields?

Tier, tier priority and policy priority are more meaningful and useful for users. But for now it's non-trivial to add support to retrieve them. We can add these fields when we have the support later.

srikartati:
> > Interesting. Thanks for checking. Changing the name sounds good to me.
> > What about other priorities? Are they useful for the user? Should we add those as different fields?
> 
> Tier, tier priority and policy priority are more meaningful and useful for users. But for now it's non-trivial to add support to retrieve them. We can add these fields when we have the support later.

Sounds good. May be TODO in Antrea code would be helpful.

zyiou:
> > > Interesting. Thanks for checking. Changing the name sounds good to me.
> > > What about other priorities? Are they useful for the user? Should we add those as different fields?
> > 
> > 
> > Tier, tier priority and policy priority are more meaningful and useful for users. But for now it's non-trivial to add support to retrieve them. We can add these fields when we have the support later.
> 
> Sounds good. May be TODO in Antrea code would be helpful to track it.

Sure. Tag @heanlan here

pkg/registry/registry.go
```diff
@@ -46,6 +46,18 @@ const (
...
+// enum for ingressNetworkPolicyType and egressNetworkPolicyType field in Antrea registry.
+const (
+	NetworkPolicyTypeK8sNetworkPolicy           = uint8(1)
+	NetworkPolicyTypeAntreaNetworkPolicy        = uint8(2)
+	NetworkPolicyTypeAntreaClusterNetworkPolicy = uint8(3)
+)
+
+// placeholder of NetworkPolicyRulePriority for K8s Network Policy.
+const (
+	NetworkPolicyRulePriorityK8s = int16(-1)
```
srikartati, line 58: Keep it simple.. K8sNetworkPolicyRulePriority?

pkg/registry/registry.go
```diff
@@ -46,6 +46,18 @@ const (
 	NetworkPolicyRuleActionReject   = uint8(3)
 )
 
+// enum for ingressNetworkPolicyType and egressNetworkPolicyType field in Antrea registry.
+const (
+	NetworkPolicyTypeK8sNetworkPolicy           = uint8(1)
```
srikartati, line 51: NetworkPolicyType* -> PolicyType*

heanlan:
> > > > Interesting. Thanks for checking. Changing the name sounds good to me.
> > > > What about other priorities? Are they useful for the user? Should we add those as different fields?
> > > 
> > > 
> > > Tier, tier priority and policy priority are more meaningful and useful for users. But for now it's non-trivial to add support to retrieve them. We can add these fields when we have the support later.
> > 
> > 
> > Sounds good. May be TODO in Antrea code would be helpful to track it.
> 
> Sure. Tag @heanlan here

Will add a TODO. But based on our discussion with Yang, I suppose the support will still need to be added by us later. It looks like they don't have the requirement to retrieve tier priority, policy priority for now.

srikartati (approved):
LGTM

PR #181, opened 2021-05-03
Modify Aggregation process methods (#158)

srikartati:
Delete the unrequired method of deleting record from
record map without the lock.
Add a method to get the flow updated time for flow given flow key

Pushing this to main. Based on the Antrea PR progress, I will add this to release-0.5.0 branch.


zyiou (approved):
LGTM. It's a previously approved PR.

PR #183, opened 2021-05-04
Modify correlation requirements in aggregation process

zyiou:
For inter-Node flows, if destination Pod has an ingress reject rule, source node will not commit the flow to conntrack table, so that we cannot do correlation for this case.

### DIFF
--- a/pkg/intermediate/aggregate.go
+++ b/pkg/intermediate/aggregate.go
@@ -606,8 +606,9 @@ func validateDataRecord(record entities.Record) bool {
 	return true
 }
 
-// isCorrelationRequired returns true when flowType is InterNode and
-// egressNetworkPolicyRuleAction is not deny (drop/reject).
+// isCorrelationRequired returns true for InterNode flowType when
+// either the egressNetworkPolicyRuleAction is not deny (drop/reject) or
+// the ingressNetworkPolicyRuleAction is not reject.
 func isCorrelationRequired(record entities.Record) bool {
 	if ieWithValue, exist := record.GetInfoElementWithValue("flowType"); exist {
 		if recordFlowType, ok := ieWithValue.Value.(uint8); ok {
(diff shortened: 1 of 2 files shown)


pkg/intermediate/aggregate.go
```diff
@@ -606,8 +606,9 @@ func validateDataRecord(record entities.Record) bool {
 	return true
 }
 
-// isCorrelationRequired returns true when flowType is InterNode and
-// egressNetworkPolicyRuleAction is not deny (drop/reject).
+// isCorrelationRequired returns true when flowType is InterNode,
```
srikartati, line 609: _the_ flowType
_the_ egressNetworkPolicyRuleAction
_the_ ingressNetworkPolicyRuleAction

pkg/intermediate/aggregate.go
```diff
@@ -606,8 +606,9 @@ func validateDataRecord(record entities.Record) bool {
...
+// egressNetworkPolicyRuleAction is not deny (drop/reject) and
```
srikartati, line 610: When I see the logic below.. it looks like either egressNetworkPolicyRuleAction is not deny or ingressNetworkPolicyRuleAction is not reject. Please use "either.. or.."  in the sentence and not "and".

srikartati (commented):
>For inter-Node flows, if destination Pod has an ingress reject rule, source node will not commit the flow to conntrack table, so that we cannot do correlation for this case.

Does the source node send the reject packet here and we do not see any packet on destination node?
In the case of ingress drop action, does the destination node see the packet?

zyiou:
> > For inter-Node flows, if destination Pod has an ingress reject rule, source node will not commit the flow to conntrack table, so that we cannot do correlation for this case.
> 
> Does the source node send the reject packet here and we do not see any packet on destination node?
> In the case of ingress drop action, does the destination node see the packet?

Ingress reject rule at destination Pod: source Node will only commit to conntrack table with flow (src -> coredns), destination Node will send reject packetin
Ingress drop rule at destination Pod, source Node will commit to conntrack table as if there is no deny rule, destination Node will send the drop packetin

srikartati:
> Ingress reject rule at destination Pod: source Node will only commit to conntrack table with flow (src -> coredns), destination Node will send reject packetin

Why coredns? Is there any dns resolution?

zyiou:
> > Ingress reject rule at destination Pod: source Node will only commit to conntrack table with flow (src -> coredns), destination Node will send reject packetin
> 
> Why coredns? Is there any dns resolution?

My bad. src -> coredns only happens when I tested with http://web case. No flow will be committed in normal cases.

srikartati (commented):
LGTM, but I see that there are no unit tests covering these corner cases of correlation, for example I did not find any records with `egressNetworkPolicyRuleAction` field present. Do you think there is scope to add unit tests covering these correlation cases?

zyiou:
> LGTM, but I see that there are no unit tests covering these corner cases of correlation, for example I did not find any records with `egressNetworkPolicyRuleAction` field present. Do you think there is scope to add unit tests covering these correlation cases?

Yes, we need to add some unit tests to cover these cases. Updated. Thanks!

srikartati (approved):
LGTM

PR #184, opened 2021-05-05
Modify the type of Antrea Registry field

heanlan:
To accomodate ovs flow priority with range [0:65000], also to specify k8s network policy with value -1.


srikartati (approved):
LGTM

PR #185, opened 2021-05-08
Modify aggregation process to maintain record expiry data

srikartati:
Aggregation proccess is modified to maintain records in a heap-based on active and inactive expiry
timeouts provided as inputs.
This is needed for providing granular intervals when exporting flow records through flow aggregator in Antrea: https://github.com/vmware-tanzu/antrea/pull/1949

Priority queue code is maintained in a separate package. This can be reused in Antrea flow exporter.

### DIFF
--- a/pkg/intermediate/aggregate.go
+++ b/pkg/intermediate/aggregate.go
@@ -16,11 +16,13 @@ package intermediate
 
 import (
 	"bytes"
+	"container/heap"
 	"encoding/binary"
 	"fmt"
 	"net"
 	"strings"
 	"sync"
+	"time"
 
 	"k8s.io/klog/v2"
 
@@ -29,9 +31,17 @@ import (
 	"github.com/vmware/go-ipfix/pkg/util"
 )
 
+var (
+	MaxRetries    = 5
+	MinExpiryTime = 100 * time.Millisecond
+)
+
 type AggregationProcess struct {
 	// flowKeyRecordMap maps each connection (5-tuple) with its records
 	flowKeyRecordMap map[FlowKey]AggregationFlowRecord
+	// expirePriorityQueue helps to maintain a priority queue for the records given
+	// active expiry and inactive expiry timeouts.
+	expirePriorityQueue TimeToExpirePriorityQueue
 	// mutex allows multiple readers or one writer at the same time
 	mutex sync.RWMutex
 	// messageChan is the channel to receive the message
@@ -49,15 +59,26 @@ type AggregationProcess struct {
 	// TODO: Add checks to validate the lists inside such as no duplicates, order
 	// of stats etc.
 	aggregateElements *AggregationElements
+	// activeExpiryTimeout helps in identifying records that elapsed active expiry
+	// timeout. Active expiry timeout is a periodic expiry interval for every flow
+	// record in the aggregation record map.
+	activeExpiryTimeout time.Duration
+	// inactiveExpiryTimeout helps in identifying records that elapsed inactive expiry
+	// timeout. Inactive expiry timeout is an expiry interval that gets reset every
+	// time a new record is received for the existing record in the aggregation
+	// record map.
+	inactiveExpiryTimeout time.Duration
 	// stopChan is the channel to receive stop message
 	stopChan chan bool
 }
 
 type AggregationInput struct {
-	MessageChan       chan *entities.Message
-	WorkerNum         int
-	CorrelateFields   []string
-	AggregateElements *AggregationElements
+	MessageChan           chan *entities.Message
+	WorkerNum             int
+	CorrelateFields       []string
+	AggregateElements     *AggregationElements
+	ActiveExpiryTimeout   time.Duration
+	InactiveExpiryTimeout time.Duration
 }
 
 // InitAggregationProcess takes in message channel (e.g. from collector) as input
@@ -69,14 +90,22 @@ func InitAggregationProcess(input AggregationInput) (*AggregationProcess, error)
 	} else if input.WorkerNum <= 0 {
 		return nil, fmt.Errorf("worker number cannot be <= 0")
 	}
+	if input.AggregateElements != nil {
+		if (len(input.AggregateElements.StatsElements) != len(input.AggregateElements.AggregatedSourceStatsElements)) || (len(input.AggregateElements.StatsElements) != len(input.AggregateElements.AggregatedDestinationStatsElements)) {
+			return nil, fmt.Errorf("stats elements, source stats elements and destination stats elemenst length should be equal")
+		}
+	}
 	return &AggregationProcess{
 		make(map[FlowKey]AggregationFlowRecord),
+		make(TimeToExpirePriorityQueue, 0),
 		sync.RWMutex{},
 		input.MessageChan,
 		input.WorkerNum,
 		make([]*worker, 0),
 		input.CorrelateFields,
 		input.AggregateElements,
+		input.ActiveExpiryTimeout,
+		input.InactiveExpiryTimeout,
 		make(chan bool),
 	}, nil
 }
@@ -148,31 +177,93 @@ func (a *AggregationProcess) ForAllRecordsDo(callback FlowKeyRecordMapCallBack) 
 	return nil
 }
 
-// GetLastUpdatedTimeOfFlow provides the last updated time in the format of IPFIX
-// field "flowEndSeconds".
-func (a *AggregationProcess) GetLastUpdatedTimeOfFlow(flowKey FlowKey) (uint32, error) {
+func (a *AggregationProcess) deleteFlowKeyFromMap(flowKey FlowKey) error {
 	a.mutex.Lock()
 	defer a.mutex.Unlock()
-	record, exists := a.flowKeyRecordMap[flowKey]
-	if !exists {
-		return 0, fmt.Errorf("flow key is not present in the map")
-	}
-	flowEndField, exists := record.Record.GetInfoElementWithValue("flowEndSeconds")
-	if exists {
-		return flowEndField.Value.(uint32), nil
-	} else {
-		return 0, fmt.Errorf("flowEndSeconds field is not present in the record")
-	}
+	return a.deleteFlowKeyFromMapWithoutLock(flowKey)
 }
 
-func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) error {
-	a.mutex.Lock()
-	defer a.mutex.Unlock()
+func (a *AggregationProcess) deleteFlowKeyFromMapWithoutLock(flowKey FlowKey) error {
 	_, exists := a.flowKeyRecordMap[flowKey]
 	if !exists {
-		return fmt.Errorf("flow key is not present in the map")
+		return fmt.Errorf("flow key %v is not present in the map", flowKey)
 	}
 	delete(a.flowKeyRecordMap, flowKey)
+	return nil
+}
+
+// GetExpiryFromExpirePriorityQueue returns the earliest timestamp (active expiry
+// or inactive expiry) from expire priority queue.
+func (a *AggregationProcess) GetExpiryFromExpirePriorityQueue() time.Duration {
+	a.mutex.Lock()
+	defer a.mutex.Unlock()
+
+	currTime := time.Now()
+	if a.expirePriorityQueue.Len() > 0 {
+		// Get the minExpireTime of the top item in expirePriorityQueue.
+		expiryDuration := MinExpiryTime + a.expirePriorityQueue.minExpireTime(0).Sub(currTime)
+		if expiryDuration < 0 {
+			return MinExpiryTime
+		}
+		return expiryDuration
+	}
+	if a.activeExpiryTimeout < a.inactiveExpiryTimeout {
+		return a.activeExpiryTimeout
+	}
+	return a.inactiveExpiryTimeout
+}
+
+func (a *AggregationProcess) ForAllExpiredFlowRecordsDo(callback FlowKeyRecordMapCallBack) error {
+	a.mutex.Lock()
+	defer a.mutex.Unlock()
+
+	if a.expirePriorityQueue.Len() == 0 {
+		return nil
+	}
+	currTime := time.Now()
+	for a.expirePriorityQueue.Len() > 0 {
+		topItem := a.expirePriorityQueue.Peek()
+		if topItem.activeExpireTime.After(currTime) && topItem.inactiveExpireTime.After(currTime) {
+			// We do not have to check other items anymore.
+			break
+		}
+		// Pop the record item from the priority queue
+		pqItem := heap.Pop(&a.expirePriorityQueue).(*ItemToExpire)
+		if !pqItem.flowRecord.ReadyToSend {
+			// Reset the timeouts and add the record to priority queue.
+			// Delete the record after max retries.
+			pqItem.flowRecord.waitForReadyToSendRetries = pqItem.flowRecord.waitForReadyToSendRetries + 1
+			if pqItem.flowRecord.waitForReadyToSendRetries > MaxRetries {
+				klog.V(2).Infof("Deleting the record after waiting for ready to send with key: %v record: %v", pqItem.flowKey, pqItem.flowRecord)
+				if err := a.deleteFlowKeyFromMapWithoutLock(*pqItem.flowKey); err != nil {
+					return fmt.Errorf("error while deleting flow record after max retries: %v", err)
+				}
+			} else {
+				pqItem.activeExpireTime = currTime.Add(a.activeExpiryTimeout)
+				pqItem.inactiveExpireTime = currTime.Add(a.inactiveExpiryTimeout)
+				heap.Push(&a.expirePriorityQueue, pqItem)
+			}
+			continue
+		}
+		err := callback(*pqItem.flowKey, *pqItem.flowRecord)
+		if err != nil {
+			return fmt.Errorf("callback execution failed for popped flow record with key: %v, record: %v, error: %v", pqItem.flowKey, pqItem.flowRecord, err)
+		}
+		// Delete the flow record if it is expired because of inactive expiry timeout.
+		if pqItem.inactiveExpireTime.Before(currTime) {
+			if err = a.deleteFlowKeyFromMapWithoutLock(*pqItem.flowKey); err != nil {
+				return fmt.Errorf("error while deleting flow record after inactive expiry: %v", err)
+			}
+			continue
+		}
+		// Reset the expireTime for the popped item and push it to the priority queue.
+		if pqItem.activeExpireTime.Before(currTime) {
+			// Reset the active expire timeout and push the record into priority
+			// queue.
+			pqItem.activeExpireTime = currTime.Add(a.activeExpiryTimeout)
+			heap.Push(&a.expirePriorityQueue, pqItem)
+		}
+	}
 	return nil
 }
 
@@ -184,6 +275,7 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
 
 	correlationRequired := isCorrelationRequired(record)
 
+	currTime := time.Now()
 	aggregationRecord, exist := a.flowKeyRecordMap[*flowKey]
 	if exist {
 		if correlationRequired {
@@ -211,6 +303,10 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
 				return err
 			}
 		}
+		// Reset the inactive expiry time in the queue item with updated aggregate
+		// record.
+		a.expirePriorityQueue.Update(aggregationRecord.PriorityQueueItem,
+			flowKey, &aggregationRecord, aggregationRecord.PriorityQueueItem.activeExpireTime, currTime.Add(a.inactiveExpiryTimeout))
 	} else {
 		// Add all the new stat fields and initialize them.
 		if correlationRequired {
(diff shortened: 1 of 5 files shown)


pkg/intermediate/aggregate.go
```diff
@@ -49,15 +58,23 @@ type AggregationProcess struct {
 	// TODO: Add checks to validate the lists inside such as no duplicates, order
 	// of stats etc.
 	aggregateElements *AggregationElements
+	// activeExpiryTimeout helps in identifying records that elapsed active expiry
+	// timeout.
+	activeExpiryTimeout time.Duration
+	// inactiveExpiryTimeout helps in identifying records that elapsed active expiry
```
zyiou, line 64: `active` -> `inactive`

pkg/intermediate/aggregate.go
```diff
@@ -176,6 +188,78 @@ func (a *AggregationProcess) DeleteFlowKeyFromMap(flowKey FlowKey) error {
...
+				pqItem.inactiveExpireTime = time.Now().Add(a.inactiveExpiryTimeout)
+				heap.Push(&a.expirePriorityQueue, pqItem)
+			}
+			continue
+		}
+		err := callback(pqItem.flowKey, pqItem.flowRecord)
+		if err != nil {
+			return fmt.Errorf("callback execution failed for popped flow record with key: %v, record: %v, error: %v", pqItem.flowKey, pqItem.flowRecord, err)
+		}
+		// Reset the expireTime for the popped item and push it to the priority queue.
```
zyiou, line 246: this comment seems inconsistent with code below.

pkg/intermediate/priorityqueue.go
```diff
@@ -0,0 +1,87 @@
...
+
+type TimeToExpirePriorityQueue []*ItemToExpire
+
+func (pq TimeToExpirePriorityQueue) Len() int {
+	return len(pq)
+}
+
+func (pq TimeToExpirePriorityQueue) Less(i, j int) bool {
+	if pq[i].activeExpireTime.Before(pq[i].inactiveExpireTime) {
+		if pq[j].activeExpireTime.Before(pq[j].inactiveExpireTime) {
```
zyiou, line 41: Will there be cases that `activeExpireTime` of an item is after `inactiveExpireTime`?

srikartati: There can be.. consider the default values of active expiry time to be 30s and inactive expiry time to be 45s. An inactive record first sends a flow record after 30s and resets it timer to 30s, where as inactive expiry time timer is 15s away.

pkg/intermediate/aggregate.go
```diff
@@ -148,37 +169,97 @@ func (a *AggregationProcess) ForAllRecordsDo(callback FlowKeyRecordMapCallBack)
...
+func (a *AggregationProcess) GetExpiryFromExpirePriorityQueue() time.Duration {
+	a.mutex.Lock()
+	defer a.mutex.Unlock()
+
+	if a.expirePriorityQueue.Len() > 0 {
+		topItem := a.expirePriorityQueue.Peek()
+		if topItem.ActiveExpireTime.Before(topItem.InactiveExpireTime) {
+			return topItem.ActiveExpireTime.Sub(time.Now())
+		} else {
+			return topItem.InactiveExpireTime.Sub(time.Now())
```
antoninbas, line 198: as a rule (applies to the code below as well) I suggest calling `time.Now()` a single time and storing the result in a variable

pkg/intermediate/aggregate.go
```diff
@@ -148,37 +169,97 @@ func (a *AggregationProcess) ForAllRecordsDo(callback FlowKeyRecordMapCallBack)
...
-		return flowEndField.Value.(uint32), nil
-	} else {
-		return 0, fmt.Errorf("flowEndSeconds field is not present in the record")
+	delete(a.flowKeyRecordMap, flowKey)
+	return nil
+}
+
+// GetExpiryFromExpirePriorityQueue returns the earliest timestamp (active expiry
+// or inactive expiry) from expire priority queue.
+func (a *AggregationProcess) GetExpiryFromExpirePriorityQueue() time.Duration {
```
antoninbas, line 189: I wonder if we should have a minimum duration time here, to avoid always handling a single item per call
Maybe 100ms?

But I am not sure whether it's really important...

pkg/intermediate/aggregate.go
```diff
@@ -16,22 +16,32 @@ package intermediate
...
 
 type AggregationProcess struct {
 	// flowKeyRecordMap maps each connection (5-tuple) with its records
-	flowKeyRecordMap map[FlowKey]AggregationFlowRecord
+	flowKeyRecordMap map[util.FlowKey]AggregationFlowRecord
+	// expirePriorityQueue helps to maintain a priority queue for the records given
+	// active expiry and inactive expiry timeouts.
+	expirePriorityQueue priorityqueue.TimeToExpirePriorityQueue
 	// mutex allows multiple readers or one writer at the same time
 	mutex sync.RWMutex
```
antoninbas, line 46: love that we still have a single mutex :)

pkg/util/priorityqueue/priorityqueue.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/vmware/go-ipfix/pkg/entities"
+	"github.com/vmware/go-ipfix/pkg/util"
+)
+
+type ItemToExpire struct {
+	// Flow related info
+	FlowKey             util.FlowKey
+	FlowRecord          entities.Record
+	IsValid             bool
```
antoninbas, line 29: I'm curious as to why we don't just keep a pointer to the aggregationRecord here?
it seems that because we have a single mutex it would work fine.

antoninbas (commented):
Some small comments, but overall LGTM. Simpler to understand than before IMO.

pkg/intermediate/aggregate.go
```diff
@@ -148,37 +169,97 @@ func (a *AggregationProcess) ForAllRecordsDo(callback FlowKeyRecordMapCallBack)
...
+		if topItem.ActiveExpireTime.Before(topItem.InactiveExpireTime) {
+			return topItem.ActiveExpireTime.Sub(time.Now())
+		} else {
+			return topItem.InactiveExpireTime.Sub(time.Now())
+		}
+	}
+	if a.activeExpiryTimeout < a.inactiveExpiryTimeout {
+		return a.activeExpiryTimeout
 	}
+	return a.inactiveExpiryTimeout
```
shihhaoli, line 204: Is the definition of activeExpiryTimeout and inactiveExpiryTimeout described somewhere or in a doc?
I don't quiet understand their meanings from line 62 and 65.

In this GetExpiryFromExpirePriorityQueue function, it looks like the caller will not know if the return value is from activeExpiryTimeout or inactiveExpiryTimeout. 
Does that mean active or inactive is not important?
So do we need to store them separately? 
For example, can we just store the smaller one during the initialization if it is the one being returned here?

pkg/intermediate/aggregate.go
srikartati: Done.  Good suggestion.

pkg/util/priorityqueue/priorityqueue.go
srikartati: I wanted to reuse this code for Antrea flow exporter, so moved it to the `entitied.Record` interface. Considering the fact that we maintain a separate record data structure in Antrea and a reference to that has to be stored in the priority queue, maybe keeping this code separate for aggregation process is better.

pkg/intermediate/aggregate.go
srikartati: Please refer this: https://github.com/vmware-tanzu/antrea/pull/1949/files#diff-6cf2c6fa47e75165fd7d385fc7ff6cce0a2eda02cfb7cf04db849031ae126ba9R134-R147
I will also add more context to the existing text in field descriptions.

A flow record can be expired because of either active or inactive flow timeouts. They do not have a constant relationship between them. Depending on timestamps of flow records received, these active and inactive timeouts get set/reset, so tracking as part of the initialization will not work.

pkg/util/priorityqueue/priorityqueue.go
srikartati: Stored the pointer to the Aggregation record directly and kept the flow key as well, because it is slightly costly to get the flow key from the aggregation record again. Keeping this code specific to the aggregation process for now.

pkg/intermediate/priorityqueue.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+}
+
+func (pq TimeToExpirePriorityQueue) Swap(i, j int) {
+	pq[i], pq[j] = pq[j], pq[i]
+	pq[i].index = i
+	pq[j].index = j
+}
+
+func (pq *TimeToExpirePriorityQueue) Push(x interface{}) {
```
zyiou, line 58: out of curiosity, why for `Push` and `Pop` we return interface instead of `*ItemToExpire`?

zyiou (commented):
LGTM overall. Just two nits.

pkg/intermediate/priorityqueue.go
srikartati: These methods are used by the method here https://golang.org/src/container/heap/heap.go?s=1949:1986#L52
Therefore, we are required to pass the interface through this function. We call the Push and Pop functions from the heap go package.
We do not have a complete implementation of the heap here; we piggyback on the heap package.
If the conversions get costly, we may have to move the complete implementation of the heap here instead of using heap go pkg.

pkg/intermediate/aggregate.go
srikartati: Added this minimum duration here as we sometimes hit situation, where the top item's timestamp is already elapsed and returns a negative duration. The ticker in Antrea do not accept non-positive duration, so using a min duration here.

pkg/intermediate/priorityqueue.go
```diff
@@ -0,0 +1,86 @@
...
+		if pq[j].activeExpireTime.Before(pq[j].inactiveExpireTime) {
+			return pq[i].activeExpireTime.Before(pq[j].activeExpireTime)
+		}
+		return pq[i].activeExpireTime.Before(pq[j].inactiveExpireTime)
+	} else {
+		if pq[j].activeExpireTime.Before(pq[j].inactiveExpireTime) {
+			return pq[i].inactiveExpireTime.Before(pq[j].activeExpireTime)
+		}
+		return pq[i].inactiveExpireTime.Before(pq[j].inactiveExpireTime)
+	}
```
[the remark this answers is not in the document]
srikartati: Good suggestion--this will simplify the code and better for code readability.

pkg/intermediate/aggregate.go
```diff
@@ -148,31 +171,93 @@ func (a *AggregationProcess) ForAllRecordsDo(callback FlowKeyRecordMapCallBack)
...
+			if err = a.deleteFlowKeyFromMapWithoutLock(*pqItem.flowKey); err != nil {
+				return fmt.Errorf("error while deleting flow record after inactive expiry: %v", err)
+			}
+			continue
+		}
+		// Reset the expireTime for the popped item and push it to the priority queue.
+		if pqItem.activeExpireTime.Before(currTime) {
+			// Reset the active expire timeout and push the record into priority
+			// queue.
+			pqItem.activeExpireTime = currTime.Add(a.activeExpiryTimeout)
```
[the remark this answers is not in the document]
srikartati: @shihhaoli In that case, we could also have gone for a lower inactive timeout than the active timeout, where we delete the flow record as soon as the inactive timeout is elapsed A new record can be added when a new flow record comes in. 

We went this route to reduce memory allocations and cleanups for the case of long-lived flows. In addition, as Yiou mentioned this also benefits to have a continuous time-series when visualizing flow metrics.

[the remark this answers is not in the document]
shihhaoli: Thanks guys. Sorry a couple more questions.

So in the current logic, we could still export the record when inactive timeout expires because the code only checks if the popped record is ready to send without checking it is due to which timeout. Is that right?

Do we also have an attribute in the record storing the last updated time?
If so, maybe we can only use active timeout as the periodical interval for the timeout queue (i.e. priority queue). Whenever it expires, then check if the time between the last updated time and current timeout exceeds the inactive timeout value. If yes, we can stop there (i.e. no longer adding the record back to the queue). In this case, the data point will be updated regularly based on active timeout interval until inactive timeout expires.

Anyhow, this is just my thought based on my current understanding. It may not reflect what you intend to do. Thanks.

pkg/intermediate/aggregate.go
srikartati: Yes, that may not be guaranteed. I modified the logic a bit to cover that scenario. Ticker does not accept 0 too.

pkg/intermediate/aggregate.go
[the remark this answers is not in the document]
srikartati: >So in the current logic, we could still export the record when inactive timeout expires because the code only checks if the popped record is ready to send without checking it is due to which timeout. Is that right?

Yes, that is correct. This ReadyToSend is to wait for flow records to be completely correlated as per our expectations. We do not send any flow records in this scenario.

>Do we also have an attribute in the record storing the last updated time?

Yes, we do have that, but this is taken from the `flowEndSeconds` element in the flow record. This is added by Flow Exporter at Antrea Agent and not Flow Aggregator. With the logic you suggested, we cannot expire the flow record at a constant interval for all inactive flow records. It depends on the time instance when the last flow record has been received.

Thanks for the comments, Shih-hao.

pkg/intermediate/aggregate.go
```diff
@@ -148,31 +172,93 @@ func (a *AggregationProcess) ForAllRecordsDo(callback FlowKeyRecordMapCallBack)
...
+func (a *AggregationProcess) GetExpiryFromExpirePriorityQueue() time.Duration {
+	a.mutex.Lock()
+	defer a.mutex.Unlock()
+
+	currTime := time.Now()
+	if a.expirePriorityQueue.Len() > 0 {
+		// Get the minExpireTime of the top item in expirePriorityQueue.
+		expiryDuration := MinExpiryTime + a.expirePriorityQueue.minExpireTime(0).Sub(currTime)
+		if expiryDuration < 0 {
+			return MinExpiryTime
```
shihhaoli, line 201: You may no longer need to add MinExpiryTime on previous line?

srikartati: MinExpiryTime in the previous line is to process multiple records simultaneously if their expiry times are very close to each other (<100ms).

shihhaoli: Got it. Maybe then doing like this?

    expiryDuration := a.expirePriorityQueue.minExpireTime(0).Sub(currTime)
    if expiryDuration < MinExpiryTime  {
        return MinExpiryTime
    }

pkg/intermediate/aggregate.go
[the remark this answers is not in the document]
shihhaoli: Got it. Thanks.
I was trying to see if we can avoid updating the queue due to change on inactive timeout because moving objects in the queue could be more expensive than just checking certain condition.

pkg/intermediate/aggregate.go
```diff
@@ -390,6 +490,33 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
 	return nil
 }
 
+// ResetStatElementsInRecord is called by the user after the aggregation record
+// is sent after its expiry either by active or inactive expiry interval. This should
+// be called by user after acquiring the mutex in the Aggregation process.
+func (a *AggregationProcess) ResetStatElementsInRecord(record entities.Record) error {
+	statsElementList := a.aggregateElements.StatsElements
+	antreaSourceStatsElements := a.aggregateElements.AggregatedSourceStatsElements
+	antreaDestinationStatsElements := a.aggregateElements.AggregatedDestinationStatsElements
```
shihhaoli, line 499: Just double check, so all three lists always have the same length?

pkg/intermediate/priorityqueue.go
```diff
@@ -0,0 +1,84 @@
...
+	item.index = n
+	*pq = append(*pq, item)
+}
+
+func (pq *TimeToExpirePriorityQueue) Pop() interface{} {
+	n := len(*pq)
+	item := (*pq)[n-1]
+	item.index = -1
+	*pq = (*pq)[0:(n - 1)]
+	return item
```
shihhaoli, line 68: I thought we should pop the first one in the queue?
BTW, should we check if the queue is empty?

pkg/intermediate/aggregate.go
shihhaoli: Sorry not quite understand the scenario.
I thought we just want to make sure this function will return at least 100ms for each call (assume non-empty queue case)?
Did you mean there are some cases, you want it to return more than 100ms?

pkg/intermediate/priorityqueue.go
srikartati: It is because of implementation inbuilt go package of heap as shared before: https://golang.org/src/container/heap/heap.go?s=1949:1986#L52

pkg/intermediate/aggregate.go
srikartati: Yes, that is correct. This is to make sure that we do not process single expired record, if the next record expiry time instance is lower than 100ms.

shihhaoli: Got it. But in the statement

    expiryDuration := MinExpiryTime + a.expirePriorityQueue.minExpireTime(0).Sub(currTime),

if a.expirePriorityQueue.minExpireTime(0).Sub(currTime) = -20ms,
then this function will return 80ms to the caller, which is less than 100ms.
Will this be OK?

pkg/intermediate/priorityqueue.go
shihhaoli: Got it. Thanks.

pkg/intermediate/aggregate.go
srikartati: Yes that will be fine. It's a best effort to have some minimum delay to avoid single flow record processing scenarios.

shihhaoli (commented):
Hi, I have a high-level question.
Do we consider using a single timeout event for all records?
In the current design, it looks to me we are adding two timeout events into the timer queue for each record. So the queue size and heap processing time could grow depending the number of active records.

How about using a global event for all records?
For example, we export all active records at once every x seconds.
Then the timer queue processing could be much simpler.
And all exported data points will have the same time series, I believe that will be easier for the analytic tools to do correlation and detection, etc.
What do you guys think? Thanks.

srikartati:
> Hi, I have a high-level question.
> Do we consider using a single timeout event for all records?
> In the current design, it looks to me we are adding two timeout events into the timer queue for each record. So the queue size and heap processing time could grow depending the number of active records.
> 
> How about using a global event for all records?
> For example, we export all active records at once every x seconds.
> Then the timer queue processing could be much simpler.
> And all exported data points will have the same time series, I believe that will be easier for the analytic tools to do correlation and detection, etc.
> What do you guys think? Thanks.

We are following the mechanism proposed in RFC: https://tools.ietf.org/html/rfc6183#section-5.3.1
The major benefit of having inactive flow timeout is cache clearing; we do not want to clear it unnecessarily. This also has performance impacts. In addition, granular intervals will help in spreading the export events at different time instances.

We will be doing performance benchmarking and will help understand where the bottlenecks are.

pkg/intermediate/aggregate.go
srikartati: >Is index 1 in your reply the first or second item in the queue
Yes that is second item and not top item.

>So you're trying to avoid the returned duration covers only one item?
Yes for cases, where second item expiry item is close to the first one. Think of batch processing. Just a heuristic-based small enhancement.

shihhaoli: OK, got it now.
I thought configuring a reasonable MinExpiryTime is good enough since there could be always a item very close to the chosen MinExpiryTime.
Thanks.

shihhaoli (commented):
> We are following the mechanism proposed in RFC: https://tools.ietf.org/html/rfc6183#section-5.3.1
> The major benefit of having inactive flow timeout is cache clearing; we do not want to clear it unnecessarily. This also has performance impacts. In addition, granular intervals will help in spreading the export events at different time instances.
> 
> We will be doing performance benchmarking in the imminent future and will help understand where the bottlenecks are.

Sounds good. 
In my mind, we can still cleanup inactive records by piggy-back the processing at the periodical active flow timeout (assuming it is ok to delay the cleanup till next active flow timeout).
But you're more familiar with the whole picture. I only had limited experiences with writing the ML script when processing those data points and past experiences with stats processing. Thanks.

srikartati:
> In the current design, it looks to me we are adding two timeout events into the timer queue for each record. So the queue size and heap processing time could grow depending the number of active records.

Why do you say queue size increases. There is only one item per record like one timeout scenario; we just add a field for the inactive timeout. The only extra processing time overhead incurred is the calculation of minimum, which is not a big overhead IMO.

shihhaoli:
> > In the current design, it looks to me we are adding two timeout events into the timer queue for each record. So the queue size and heap processing time could grow depending the number of active records.
> 
> Why do you say queue size increases? There is only one item per record like one timeout scenario; we just add a field for the inactive timeout. The only extra processing time overhead incurred is the calculation of minimum, which is not a big overhead IMO.

Sorry it should be one extra event (instead of two events) added to the timer queue per record.
So whenever we have a new record, we will add a timeout event for it in the queue.
Thus the timer queue could have N items if we have N active records.
Does it sound right? Thanks.

srikartati:
> Sorry it should be one extra event (instead of two events) added to the timer queue per record.
> So whenever we have a new record, we will add a timeout event for it in the queue.
> Thus the timer queue could have N items if we have N active records. Does it sound right?
> 
> When adding a new item to the queue, I believe the re-heap time is O(log N), which should not be a big overhead if N is not super large.
> 
> Thanks.

Yes every time a new record is received, we need to heapify the queue. With inactive timeout, yes the number of times we heapify the queue would increase. We should keep this in mind when benchmarking the Flow Aggregator code in Antrea.

antoninbas (approved):
LGTM

zyiou (approved):
LGTM

shihhaoli (approved):

PR #189, opened 2021-05-12
Modify constant K8sNetworkPolicyRulePriority's type to int32

heanlan:

PR #190, opened 2021-05-13
Modify constant K8sNetworkPolicyRulePriority's type to int32 (#189)

zyiou:


srikartati (approved):
This is just cherrypick of original commit right?

zyiou:
> This is just cherrypick of original commit right?

Yes. It should be merged to main at first place.

PR #195, opened 2021-05-14
Fix missing klog version change

zyiou:
Also update some package versions to keep consistent with Antrea go.mod


srikartati (approved):
LGTM

PR #196, opened 2021-05-14
Add support for isMetadataFilled for aggregation process

zyiou:
If correlation is required, we will set `isMetadataFilled` to true.
if correlation is not required, we will set `isMetadataFilled` to true only when the flow is intra-Node.

`isMetadataFilled` is always true for Intra-Node and ToExternal flows and only applicable for Inter-Node flows that are not required to be correlated. (e.g. flows with Egress deny rule applied)

### DIFF
--- a/pkg/intermediate/aggregate.go
+++ b/pkg/intermediate/aggregate.go
@@ -281,7 +281,13 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
 	a.mutex.Lock()
 	defer a.mutex.Unlock()
 
-	correlationRequired := isCorrelationRequired(record)
+	var flowType uint8
+	if flowTypeIE, exist := record.GetInfoElementWithValue("flowType"); exist {
+		flowType = flowTypeIE.Value.(uint8)
+	} else {
+		klog.Warning("FlowType does not exist in current record.")
+	}
+	correlationRequired := isCorrelationRequired(flowType, record)
 
 	currTime := time.Now()
 	aggregationRecord, exist := a.flowKeyRecordMap[*flowKey]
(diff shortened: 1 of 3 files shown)


pkg/intermediate/aggregate.go
```diff
@@ -331,6 +340,12 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
 		}
 		if !correlationRequired {
 			aggregationRecord.ReadyToSend = true
+			// if no correlation is required for inter-Node record, K8s metadata will not be filled.
```
srikartati, line 343: Could you rephrase this? Also please explicitly mention that for Intra-node flows and ToExternal flows, we can set the MetaDataFilled as true by default.

srikartati (commented):
Please also add explicitly that the new field is always true for Intra-Node flows and ToExternal flows in the PR description. It is only applicable for inter-node flows and we do not expect them to be correlated.

pkg/intermediate/aggregate.go
```diff
@@ -331,6 +340,14 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
 		}
 		if !correlationRequired {
 			aggregationRecord.ReadyToSend = true
+			// if no correlation is required for inter-Node record, K8s metadata will not
+			// be filled. For Intra-Node flows and ToExternal flows, isMetaDataFilled is set
+			// to true by default.
+			if isRecordFromSrc(record) || isRecordFromDst(record) {
```
srikartati, line 346: Not your change? Do we still need `isRecordFromSrc` and `isRecordFromDst`? We already have flow type in the flow record. Can we use that to make code simple?

pkg/intermediate/aggregate.go
```diff
@@ -331,6 +340,14 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
 		}
 		if !correlationRequired {
 			aggregationRecord.ReadyToSend = true
+			// if no correlation is required for inter-Node record, K8s metadata will not
```
srikartati, line 343: "If no correlation is required for an inter-Node record, K8s metadata is expected to be not completely filled."

pkg/intermediate/aggregate.go
zyiou: Sure. We can switch to using flowType here. But function `isRecordFromSrc`, `isRecordFromDst` and `areRecordsFromSameNode` need to be kept because this information is not revealed from flowType.

srikartati: Yes we still need those methods. Moving to flowType here makes the code easily readable.

srikartati (approved):
Thanks for addressing comments. 
LGTM

PR #197, opened 2021-05-15
Fix klog version

srikartati:
Missed one change when moving to klog v2.


srikartati:
PR was already raised.

PR #200, opened 2021-05-18
Remove unnecessary allocation in AddInfoElement method

srikartati:
This showed up in performance benchmarking through Antrea.
https://github.com/antrea-io/antrea/pull/2188


zyiou (approved):
Thanks for addressing this. LGTM

PR #202, opened 2021-05-25
Improve performance of go-ipfix

zyiou:


zyiou:
Close this due to duplication as #204

PR #203, opened 2021-05-25
Add Pod Labels Field into Registry

dreamtalen:
Add source & destination Pod labels field into registry, these information will be filled on the Flow Aggregator side and are necessary in developing network policy recommendation feature.


srikartati (approved):
LGTM

PR #208, opened 2021-06-04
Modify the proto format of the message

srikartati:
Instead of using oneof proto property, we move to `protoreflect.Message` to support multiple proto schemas.

The advantage is that users consuming our Kafka producer can define their own proto directly and implement the converter interface.


pkg/producer/convertor/test/flowtype_test.go
```diff
@@ -30,7 +30,7 @@ import (
 
 var (
 	// Hard coding the kafka msg in bytes. Need to figure out a way to generate this.
-	msg1ForFlowType1 = []byte{0x0, 0x0, 0x0, 0xa7, 0xa, 0xa4, 0x1, 0x10, 0x1, 0x18,
+	msg1ForFlowType1 = []byte{0x0, 0x0, 0x0, 0xa4, 0x10, 0x1, 0x18,
```
zyiou, line 33: why are the changes in message information here? is it because we don't need to init flowMessage struct?

srikartati: Yes there is no wrapper using oneof.. we just cast all protos to `Protoreflect.Message` format.

srikartati (commented):
Hold on to the review. I realized that I need to add another method in the convertor interface.

srikartati:
> Hold on to the review. I realized that I need to add another method in the convertor interface.

Please take a look now.

pkg/producer/convertor/convertor.go
```diff
@@ -28,5 +29,10 @@ var ProtoSchemaConvertor = map[string]RegisterProtoSchema{}
 type IPFIXToKafkaConvertor interface {
 	// ConvertIPFIXMsgToFlowMsgs converts multiple data records in the IPFIX message
 	// to flow messages for any given proto schema.
-	ConvertIPFIXMsgToFlowMsgs(msg *entities.Message) []*protobuf.FlowMessage
+	ConvertIPFIXMsgToFlowMsgs(msg *entities.Message) []protoreflect.Message
+	// ConvertIPFIXRecordToFlowMsg converts a single IPFIX record to a kafka flow
+	// message for any given proto schema. This is useful for producers
+	// send single record at a time instead of a batch of records. Using the msg
+	// input parameter is optional depending on the fields in the proto schema.
+	ConvertIPFIXRecordToFlowMsg(msg *entities.Message, record entities.Record) protoreflect.Message
```
zyiou, line 37: why do we need msg as a parameter? In my view with msg this method is similar to `ConvertIPFIXMsgToFlowMsgs`. `ConvertIPFIXMsgToFlowMsgs` itself can also return a single `protoreflect.Message` for single record.

srikartati: You are right. Technically we do not need it. In an effort to reduce the duplicated code, I forced the parameter in there. Let me refactor it differently.

zyiou (approved):
Thanks for the changes. LGTM

PR #210, opened 2021-06-07
Add more methods to Kafka producer

srikartati:
Add methods to send the IPFIX record instead of the whole IPFIX
message, close the sarama kafka go client, get the sarama kafka go client etc.

### DIFF
--- a/pkg/producer/kafka.go
+++ b/pkg/producer/kafka.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"sync"
 
 	"github.com/Shopify/sarama"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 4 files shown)


pkg/producer/kafka.go
```diff
@@ -164,14 +179,40 @@ func (kp *KafkaProducer) SendFlowMessage(msg protoreflect.Message, kafkaDelimitM
...
+	if kp.producer == nil {
+		return
+	}
+
+	var wg sync.WaitGroup
+	kp.producer.AsyncClose()
+
+	wg.Add(1)
+	go func() {
+		close(kp.closeCh)
```
zyiou, line 209: We may need to add check whether kp.closeCh is nil or KafkaLogErrors is enabled before calling close.

pkg/producer/kafka.go
```diff
@@ -164,14 +179,40 @@ func (kp *KafkaProducer) SendFlowMessage(msg protoreflect.Message, kafkaDelimitM
...
+		if kp.input.KafkaLogSuccesses {
+			for range kp.producer.Successes() {
+				klog.Error("Unexpected message on Successes()")
+			}
```
zyiou, line 213: under what circumstances would this case happen?

srikartati: The producer is async producer. When `sendFlowMessage` and `Close` called simultaneously, this may happen.

pkg/producer/kafka.go
srikartati: Good catch. Done.

zyiou (approved):
LGTM

PR #214, opened 2021-06-15
Add consumer and broker to consume records from producer

zyiou:


zyiou:
@srikartati Please take a quick look of this PR when you get a chance. I still have some testing for integrating with producer. Just want to see whether I'm on the right track. Thanks!

srikartati (commented):
I see that you are not using sarama kafka library to create the broker. You seem to be just starting TCP/UDP server to receive protobuf messages and decode those.

In theory, kafka producer (which is sarama based) and broker can be independent, so it should be ok. Did you go this route to keep things simple? I think Kafka config may not be very mature if we go this route like version, creating Kafka topics, etc. However, enabling TLS communication is possible with this route--we just have to start server with TLS certs.

I see a broker test with a simple client. Could you add an integration test involving kafka producer and broker in go-ipfix library to see if the communication can be established or not? This is like IPFIX exporter-collector test.

Another comment is the code needs some reorganization.. we should put producer and broker under one package something like `pkg/kafka`.

zyiou:
> I see that you are not using sarama kafka library to create the broker. You seem to be just starting TCP/UDP server to receive protobuf messages and decode those.
> 
> In theory, kafka producer (which is sarama based) and broker can be independent, so it should be ok. Did you go this route to keep things simple? I think Kafka config may not be very mature if we go this route like version, creating Kafka topics, etc. However, enabling TLS communication is possible with this route--we just have to start server with TLS certs.

I'm considering two aspects when adopting this method: 1. I didn't see which part from sarama library we can use for our case: for brokers, they simply return the broker address and id instead of actually creating brokers 2. it is simple and can meet our needs for decoding. However, I do encounter the problem when connecting it with producer. I will need to do further investigation on this before proceeding.

README.md
```diff
@@ -35,6 +35,18 @@ For example:
...
+### Start a Kafka instance
+To start a Kafka server with minimum Kubernetes setup, run
+```shell
+kubectl apply -f ./build/yamls/kafka-server.yaml -n [NAMESPACE]
+``` 
+Broker address will be `kafka-service:9092`.
+Then we can connect a Kafka consumer to this server by running: 
+```shell
+kubectl create deployment --image=antrea/kafka-consumer consumer -n [NAMESPACE]
+```
```
srikartati, line 47: We should have a single manifest for kafka broker and consumer if possible to keep the deployment simple. We already have a workflow to upload the cosumer image.

I think it is ok to pull the image from the confluent repo in the case of Kafka borker. This is similar to ELK flow collector. However, there may be an issue in Antrea e2e tests if there are any pull errors from docker registry. It is better to get a copy and upload it to the vmware harbor registry. 
Maybe we should have a check on the version and upload the image through a daily job. What do you think?

build/yamls/kafka-server.yaml
```diff
@@ -0,0 +1,81 @@
+apiVersion: v1
```
srikartati, line 1: Is this taken from any open source project? Should we add a reference?

cmd/consumer/protobuf/antrea.pb
```diff
@@ -0,0 +1,66 @@
+// Copyright 2021 VMware, Inc.
```
srikartati, line 1: We should probably move the protobuf folder to pkg/consumer.

pkg/consumer/consumer.go
```diff
@@ -0,0 +1,102 @@
+// Copyright 2021 VMware, Inc.
```
srikartati, line 1: Let's keep both producer and consumer in pkg/kafka?

pkg/producer/convertor/test/flowtype_test.go
```diff
@@ -26,74 +26,7 @@ import (
...
-		0x70, 0xe8, 0x7, 0x78, 0x32, 0x80, 0x1, 0x32, 0x88, 0x1, 0x32, 0x90, 0x1,
-		0x32, 0x9a, 0x1, 0x5, 0x4, 0x70, 0x6f, 0x64, 0x31, 0xa2, 0x1, 0x7, 0x6, 0x70,
-		0x6f, 0x64, 0x6e, 0x73, 0x31, 0xaa, 0x1, 0x6, 0x5, 0x6e, 0x6f, 0x64, 0x65,
-		0x31, 0xb2, 0x1, 0x5, 0x4, 0x70, 0x6f, 0x64, 0x32, 0xba, 0x1, 0x7, 0x6, 0x70,
-		0x6f, 0x64, 0x6e, 0x73, 0x32, 0xc2, 0x1, 0x6, 0x5, 0x6e, 0x6f, 0x64, 0x65,
-		0x32, 0xca, 0x1, 0xb, 0x31, 0x39, 0x32, 0x2e, 0x31, 0x36, 0x38, 0x2e, 0x30,
-		0x2e, 0x31, 0xd2, 0x1, 0x5, 0x4, 0x73, 0x76, 0x63, 0x31, 0xea, 0x1, 0x1,
-		0x0, 0xf2, 0x1, 0x1, 0x0, 0xfa, 0x1, 0x1, 0x0, 0x82, 0x2, 0x1, 0x0, 0x8a,
-		0x2, 0x9, 0x31, 0x32, 0x37, 0x2e, 0x30, 0x2e, 0x30, 0x2e, 0x31, 0x90, 0x2,
-		0x83, 0x25}
```
srikartati, line 46: were these moved for consumer tests?

cmd/consumer/protobuf/antrea.pb
srikartati: Thought more about this. Maybe we could use the proto code (protobuf and generated code) directly from Antrea repo, so that we will not have any compatibility issues later. What do you think?

build/yamls/kafka-server.yaml
zyiou: I created from scratch using deployment template. So no need to add special reference I think.

pkg/consumer/consumer.go
zyiou: Sure. Updated

pkg/producer/convertor/test/flowtype_test.go
zyiou: yes, planning to add some testing code for consumer.

cmd/consumer/protobuf/antrea.pb
zyiou: Will it cause dependency cycle if we import the proto code from Antrea and Antrea import other lib from go-ipfix?

README.md
zyiou: Yes it makes sense to keep image copies at our side. Will create a job in Antrea and update our yaml accordingly.

srikartati: It doesn't have to be Antrea. We can do it in go-ipfix? Do you see an issue? We have access to Antrea repo on harbor registry.

cmd/consumer/protobuf/antrea.pb
srikartati: I see that antrea proto is only used in cmd/consumer. I do not think we have to import this package to Antrea.
Lets keep it separate for now. We can make the change when Antrea code gets merged later.

README.md
zyiou: Oh right. we can do it here.

.github/workflows/upload_release_assets.yml
```diff
@@ -25,3 +25,18 @@ jobs:
           asset_path: ./assets/ipfix-collector.yaml
           asset_name: ipfix-collector.yaml
           asset_content_type: application/octet-stream
+  build-consumer:
+    runs-on: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build assets
+        run: cp ./build/yamls/kafka-server.yaml ./assets
+      - name: Upload kafka-server.yaml
```
srikartati, line 34: let's call kafka-server.yaml as kafka-flow-collector.yaml to be more specific and it could cover both broker and consumer?

cmd/consumer/consumer.go
```diff
@@ -0,0 +1,129 @@
...
+		case <-signalCh:
+			close(stopCh)
+			return
+		}
+	}
+}
+
+func run() error {
+	klog.Info("Starting Kafka consumer")
+	flowType := &protobuf.AntreaFlowMsg{}
```
srikartati, line 77: this is ok for now.. we should consider an interface that takes proto schema name as input in the future.

pkg/kafka/consumer/consumer.go
```diff
@@ -0,0 +1,102 @@
...
+	"strings"
+
+	"github.com/Shopify/sarama"
+	"google.golang.org/protobuf/proto"
+	"k8s.io/klog/v2"
+)
+
+const msgDelimitLen = 4
+
+type ConsumerInput struct {
```
srikartati, line 28: We may want to consider having TLS certificate support in the future. It is ok for now.

srikartati (commented):
I think only thing missing is the github workflow to upload confluent broker and zookeeper images

srikartati:
> I think only thing missing is the github workflow to upload confluent broker and zookeeper images

Or may be we can use docker proxy cache instead of copying.
Proxy Cache: https://confluence.eng.vmware.com/pages/viewpage.action?spaceKey=AO&title=Corporate+Harbor
Do you want to try that instead of duplicating images?

zyiou:
> > I think only thing missing is the github workflow to upload confluent broker and zookeeper images
> 
> Or may be we can use docker proxy cache instead of copying.
> Proxy Cache: https://confluence.eng.vmware.com/pages/viewpage.action?spaceKey=AO&title=Corporate+Harbor
> Do you want to try that instead of duplicating images?

Do you mean we can get the image from link like `harbor-repo.vmware.com/dockerhub-proxy-cache/go-ipfix/cp-kafka` and set up a cron job for uploading the latest images to harbor? In this case we don't need to set up daily job since cache is there.

srikartati:
> > 
> > Or may be we can use docker proxy cache instead of copying.
> > Proxy Cache: https://confluence.eng.vmware.com/pages/viewpage.action?spaceKey=AO&title=Corporate+Harbor
> > Do you want to try that instead of duplicating images?
> 
> Do you mean we can get the image from link like `harbor-repo.vmware.com/dockerhub-proxy-cache/go-ipfix/cp-kafka` and set up a cron job for uploading the latest images to harbor? In this case we don't need to set up daily job since cache is there.

Yes, no need for a job, and also we can avoid making a copy of the images. Instead directly use `harbor-repo.vmware.com/dockerhub-proxy-cache/confluent/*` in our manifests.

cmd/consumer/consumer.go
zyiou: Do you mean similar to what producer does right now?

zyiou:
> > > Or may be we can use docker proxy cache instead of copying.
> > > Proxy Cache: https://confluence.eng.vmware.com/pages/viewpage.action?spaceKey=AO&title=Corporate+Harbor
> > > Do you want to try that instead of duplicating images?
> > 
> > 
> > Do you mean we can get the image from link like `harbor-repo.vmware.com/dockerhub-proxy-cache/go-ipfix/cp-kafka` and set up a cron job for uploading the latest images to harbor? In this case we don't need to set up daily job since cache is there.
> 
> Yes, no need for a job, and also we can avoid making a copy of the images. Instead directly use `harbor-repo.vmware.com/dockerhub-proxy-cache/confluent/*` in our manifests.

Not sure I get it correctly. Created a PR #224  for uploading images and replaced the docker images we will use in kafka-flow-collector.yml with cached ones.

srikartati:
> Not sure I get it correctly. Created a PR #224 for uploading images and replaced the docker images we will use in kafka-flow-collector.yml with cached ones.

When we push to Antrea docker repo, as Antrea repo is authorized, we can only access from harbor registry:
https://projects.registry.vmware.com/harbor/projects/539/repositories/ipfix-collector
`docker pull projects.registry.vmware.com/antrea/ipfix-collector:latest`

Where as docker proxy cache registry is somewhat different. It maintains copies for all unauthorized repos on docker.io registry including confluentc. Following command worked for me on my local machine. This does not have any rate limiting issue. 
`docker pull harbor-repo.vmware.com/dockerhub-proxy-cache/confluentinc/cp-kafka:latest`

We can use these docker proxy cache images in our manifest directly instead of copying images to antrea and pulling them from projects.registry.vmware.com/antrea again. We can avoid the github workflow. Does this make sense?

cmd/consumer/consumer.go
srikartati: I meant the user of kafka producer can take in multiple proto schema as input and create producer with appropriate convertor. Consumer application should take the proto schema name as input and decode based on that. We are currently supporting only Antrea proto schema.

zyiou:
> When we push to Antrea docker repo, as Antrea repo is authorized, we can only access from harbor registry:
> https://projects.registry.vmware.com/harbor/projects/539/repositories/ipfix-collector
> `docker pull projects.registry.vmware.com/antrea/ipfix-collector:latest`
> 
> Where as docker proxy cache registry is somewhat different. It maintains copies for all unauthorized repos on docker.io registry including confluentc. Following command worked for me on my local machine. This does not have any rate limiting issue.
> `docker pull harbor-repo.vmware.com/dockerhub-proxy-cache/confluentinc/cp-kafka:latest`
> 
> We can use these docker proxy cache images in our manifest directly instead of copying images to antrea and pulling them from projects.registry.vmware.com/antrea again. We can avoid the github workflow. Does this make sense?

Got it. Thanks a lot for clarifying. Yes it makes sense in our case. Removed the unnecessary PR.

.github/workflows/upload_release_assets.yml
```diff
@@ -25,3 +25,18 @@ jobs:
           asset_path: ./assets/ipfix-collector.yaml
           asset_name: ipfix-collector.yaml
           asset_content_type: application/octet-stream
+  build-consumer:
```
srikartati, line 28: build-kafka-collector here and build-ipfix-collector above

srikartati (commented):
LGTM except for one nit.

.github/workflows/upload_release_assets.yml
zyiou: Sorry just notice this comment. Done. Thanks!

srikartati (approved):
Thanks for the PR

PR #215, opened 2021-06-16
Move bytes.buffer to []byte for set and message

zyiou:
This PR is about modifying `bytes.Buffer` struct to `[]byte` for set and message. It also optimizes the way of encoding record, set and message to reduce memory allocs.
Before
```
BenchmarkExportConntrackConns-2   	     154	   9073336 ns/op	  378045 B/op	   15034 allocs/op
BenchmarkExportDenyConns-2   	     252	   5222296 ns/op	  231724 B/op	    9411 allocs/op
```
After
```
BenchmarkExportConntrackConns-2   	     168	   7869668 ns/op	  297198 B/op	    5651 allocs/op
BenchmarkExportDenyConns-2   	     259	   5044816 ns/op	  183199 B/op	    3385 allocs/op
```
For conntrack connections exporting, CPU time gets reduced by 13.27%, memory usage reduced by 21.39%, allocs reduced by 62.41%. 
For deny connections exporting, CPU time gets reduced by 3.40%, memory usage reduced by 20.94%, allocs reduced by 64.03%


zyiou:
I tried this change with Flow Exporter performance benchmarking but unfortunately, didn't see any improvement. Wondering whether we need to keep this PR or close it. Pros I can think of keeping is that it makes encoding of record, set, message consistent with []byte. @srikartati Please also have a try on your side for us to decide. Thanks!

pkg/exporter/process.go
```diff
@@ -248,7 +243,7 @@ func (ep *ExportingProcess) createAndSendMsg(set entities.Set) (int, error) {
 
 	// Append the byte slices together to send on the exporter connection rather
 	// than copying the set buffer to message buffer again.
-	bytesSlice := append(msg.GetMsgBuffer().Bytes(), set.GetBuffer().Bytes()...)
+	bytesSlice := append(msg.GetMsgBuffer(), set.GetBuffer()...)
```
srikartati, line 246: could you try the following instead of append? Clearly the premise in the comment is wrong :)
https://github.com/golang/go/wiki/SliceTricks#copy

pkg/exporter/process.go
```diff
@@ -215,15 +215,10 @@ func (ep *ExportingProcess) NewTemplateID() uint16 {
 func (ep *ExportingProcess) createAndSendMsg(set entities.Set) (int, error) {
 	// Create a new message and use it to send the set.
 	msg := entities.NewMessage(false)
```
srikartati, line 217: Since we have the set, we know the memory size that needs to be allocated right? Can't we use that here?

srikartati (commented):
Thanks for making this change. Couple of comments.
I will try this patch on my perf test too.

pkg/exporter/process.go
zyiou: Sure. Updated.

pkg/exporter/process.go
zyiou: Found that we don't actually append set buffer into message, so the message size is fixed (header length). When sending message, we just combine message header and set buffer. So we don't need to consider message buffer pre-alloc here.

srikartati: Sounds good. Did you see better performance?

zyiou: Sadly no :(

srikartati:
Here is the comparison in my runs:
Before fix:
```
vagrant@k8s-node-worker-1:~/antrea/pkg/flowaggregator$ go test -test.v -run=none -test.benchmem  -bench=. -count=2 -memprofile memprofile.out -cpuprofile profile.out 
goos: linux
goarch: amd64
pkg: antrea.io/antrea/pkg/flowaggregator
BenchmarkIntraNodeFlowRecords
    flowaggregator_perf_test.go:458: Num messages received: 1550264
BenchmarkIntraNodeFlowRecords-2   	       1	120004095225 ns/op	8876354472 B/op	332625061 allocs/op
PASS
```

After fix:
```
vagrant@k8s-node-worker-1:~/antrea/pkg/flowaggregator$ go test -test.v -run=none -test.benchmem  -bench=. -memprofile memprofile.out -cpuprofile profile.out 
goos: linux
goarch: amd64
pkg: antrea.io/antrea/pkg/flowaggregator
BenchmarkIntraNodeFlowRecords
    flowaggregator_perf_test.go:479: Num messages received: 1536208
BenchmarkIntraNodeFlowRecords-2   	       1	120002065282 ns/op	9408062880 B/op	340186942 allocs/op
PASS
ok  	antrea.io/antrea/pkg/flowaggregator	120.304s
```

There is no real improvement. 
As mentioned offline, there are different entities (record, set, message) and we are allocating memory for bytes at each entity. We should think of a way to read the memory at each entity rather than writing them again and again.
We should look at chaining readers like [here](https://medium.com/learning-the-go-programming-language/streaming-io-in-go-d93507931185)?

pkg/entities/set.go
```diff
@@ -127,13 +136,7 @@ func (s *set) AddRecord(elements []*InfoElementWithValue, templateID uint16) err
...
 	if !s.isDecoding {
 		recordBytes := record.GetBuffer()
-		bytesWritten, err := s.buffer.Write(recordBytes)
-		if err != nil {
-			return fmt.Errorf("error in writing the buffer to set: %v", err)
-		}
-		if bytesWritten != len(recordBytes) {
-			return fmt.Errorf("bytes written length is not expected")
-		}
+		s.buffer = append(s.buffer, recordBytes...)
```
srikartati, line 139: could we avoid this append(write)?

srikartati (commented):
I like the approach of maintaining all element lengths and then encoding. The encoding function can be futrther enhanced by taking the allocated bytes and filling them. We can reduce the allocations in the encoding function. What do you think?

zyiou:
> I like the approach of maintaining all element lengths and then encoding. The encoding function can be futrther enhanced by taking the allocated bytes and filling them. We can reduce the allocations in the encoding function. What do you think?

Yes, but we will need to keep element list instead which is not necessary for encoding. Tested with latest change with pre-allocation for set and record. But the result seems not improving:
Before:
```
BenchmarkPoll-2  	  112	  9965818 ns/op	 902472 B/op	  54534 allocs/op
BenchmarkExportConntrackConns-2  106	  9684719 ns/op	 685126 B/op	  15782 allocs/op
BenchmarkExportDenyConns-2  	  247	  5021390 ns/op	 215250 B/op	  6539 allocs/op

```
After
```
BenchmarkPoll-2  	  127	  8602612 ns/op	 887633 B/op	  54184 allocs/op
BenchmarkExportConntrackConns-2  123	  8935159 ns/op	 679445 B/op	  18822 allocs/op
BenchmarkExportDenyConns-2  	  204	  4964911 ns/op	 304444 B/op	  10967 allocs/op
```

srikartati:
> Yes, but we will need to keep element list instead which is not necessary for encoding. Tested with latest change with pre-allocation for set and record. But the result seems not improving:
> Before:
> 
> ```
> BenchmarkPoll-2  	  112	  9965818 ns/op	 902472 B/op	  54534 allocs/op
> BenchmarkExportConntrackConns-2  106	  9684719 ns/op	 685126 B/op	  15782 allocs/op
> BenchmarkExportDenyConns-2  	  247	  5021390 ns/op	 215250 B/op	  6539 allocs/op
> ```
> 
> After
> 
> ```
> BenchmarkPoll-2  	  127	  8602612 ns/op	 887633 B/op	  54184 allocs/op
> BenchmarkExportConntrackConns-2  123	  8935159 ns/op	 679445 B/op	  18822 allocs/op
> BenchmarkExportDenyConns-2  	  204	  4964911 ns/op	 304444 B/op	  10967 allocs/op
> ```

CPU time seems to be better even though allocations have gone up.
I looked at the change. We are still writing to a slice at all entities: record, set and message. Once we write to bytes as part of encoding.. could we just read them in other entities. Can this alternative work?

zyiou:
> CPU time seems to be better even though allocations have gone up.
> I looked at the change. We are still writing to a slice at all entities: record, set and message. Once we write to bytes as part of encoding.. could we just read them in other entities. Can this alternative work?

Yes, thanks for the suggestion. I removed the write to set (record and message will need the write for storing record and send message respectively) and saw some improvement in CPU and memory:
Before
```
BenchmarkExportConntrackConns-2   	     154	   9073336 ns/op	  378045 B/op	   15034 allocs/op
BenchmarkExportDenyConns-2   	     252	   5222296 ns/op	  231724 B/op	    9411 allocs/op
```
After
```
BenchmarkExportConntrackConns-2   	     187	   8077640 ns/op	  317389 B/op	   12054 allocs/op
BenchmarkExportDenyConns-2   	     301	   4548239 ns/op	  188549 B/op	    7381 allocs/op
```

srikartati:
> Yes, thanks for the suggestion. I removed the write to set (record and message will need the write for storing record and send message respectively) and saw some improvement in CPU and memory:
> Before
> 
> ```
> BenchmarkExportConntrackConns-2   	     154	   9073336 ns/op	  378045 B/op	   15034 allocs/op
> BenchmarkExportDenyConns-2   	     252	   5222296 ns/op	  231724 B/op	    9411 allocs/op
> ```
> 
> After
> 
> ```
> BenchmarkExportConntrackConns-2   	     187	   8077640 ns/op	  317389 B/op	   12054 allocs/op
> BenchmarkExportDenyConns-2   	     301	   4548239 ns/op	  188549 B/op	    7381 allocs/op
> ```

Good improvement!
One more thing with `EncodeToIEDataType`. We are allocating bytes two times. Once for the record and doing it again inside `EncodeToIEDataType`. Could we send the pointer of allocated memory and fill it inside the Encode function?

zyiou:
> One more thing with `EncodeToIEDataType`. We are allocating bytes two times. Once for the record and doing it again inside `EncodeToIEDataType`. Could we send the pointer of allocated memory and fill it inside the Encode function?

Sure. After sending allocated memory to encoding function, alloc gets great improvement :)
Before
```
BenchmarkExportConntrackConns-2   	     187	   8077640 ns/op	  317389 B/op	   12054 allocs/op
BenchmarkExportDenyConns-2   	     301	   4548239 ns/op	  188549 B/op	    7381 allocs/op
```
After
```
BenchmarkExportConntrackConns-2   	     168	   7869668 ns/op	  297198 B/op	    5651 allocs/op
BenchmarkExportDenyConns-2   	     259	   5044816 ns/op	  183199 B/op	    3385 allocs/op
```

pkg/entities/record.go
```diff
@@ -121,6 +116,27 @@ func (d *dataRecord) PrepareRecord() error {
...
 }
 
+func (d *dataRecord) GetBuffer() []byte {
+	if len(d.buffer) == d.len || d.isDecoding {
+		return d.buffer
+	}
+	d.buffer = make([]byte, d.len)
+	index := 0
+	for _, element := range d.orderedElementList {
+		len := GetInfoElementLen(element)
```
srikartati, line 126: can this length be store in the element first time instead of calling `GetInfoElementLen` again?

pkg/entities/ie.go
```diff
@@ -388,3 +388,158 @@ func EncodeToIEDataType(dataType IEDataType, val interface{}) ([]byte, error) {
 	}
 	return nil, fmt.Errorf("API supports only valid information elements with datatypes given in RFC7011")
 }
+
+// EncodeToBuff is to encode data to specific type to the buff
+func EncodeToBuff(dataType IEDataType, val interface{}, buffer []byte, index int) error {
```
srikartati, line 393: This is good. Just worried about the scenario if the buffer does not have enough length, where it will cause panic. Maybe have a check before calling this function?

srikartati (commented):
The new results are great!

pkg/entities/record.go
zyiou: Yes. Updated.

pkg/entities/ie.go
zyiou: Makes sense. Added the check and make this method private since it will not be used elsewhere

srikartati (approved):
Thanks for making changes.
LGTM

PR #216, opened 2021-06-17
Add more metadata to aggregated flow record

srikartati:
Add isIPv4 and isExporterIPv4 fields to aggregated flow record.

This is related to [the issue](https://github.com/antrea-io/antrea/issues/2282) raised in Antrea dual-stack cluster tests.

### DIFF
--- a/pkg/intermediate/types.go
+++ b/pkg/intermediate/types.go
@@ -45,6 +45,12 @@ type AggregationFlowRecord struct {
 	// false when creating new AggregationFlowRecord. Setting and utilizing this
 	// field is upto the user and not used in go-ipfix library code.
 	areExternalFieldsFilled bool
+	// isIPv4 indicates whether the source and destination addresses are IPv4 or
+	// IPv6 in the aggregated flow record.
+	isIPv4 bool
+	// isExporterIPv4 indicates whether the exporter address of the received flow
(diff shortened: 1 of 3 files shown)


pkg/intermediate/types.go
```diff
@@ -45,6 +45,12 @@ type AggregationFlowRecord struct {
 	// false when creating new AggregationFlowRecord. Setting and utilizing this
 	// field is upto the user and not used in go-ipfix library code.
 	areExternalFieldsFilled bool
+	// isIPv4 indicates whether the source and destination addresses are IPv4 or
+	// IPv6 in the aggregated flow record.
+	isIPv4 bool
+	// isExporterIPv4 indicates whether the exporter address of the received flow
+	// aggregator is IPv4 or IPv6.
+	isExporterIPv4 bool
```
zyiou, line 53: We should either make these two fields public or have methods to access them if we want to use at Antrea side.

srikartati: Agreed. Missed it. Will add Get interface for these metadata fields and they are needed.

zyiou (approved):
LGTM. Thanks for the changes. It's more clear for us to have IPv4/v6 check at go-ipfix side.

PR #220, opened 2021-07-02
Clear original export fields in the aggregation process

srikartati:
Original export fields should be removed in the intermediate process
in their current state.
In cases, where correlated flow records are there, current original export
fields do not make sense as records can come from two different
exporters.

More details in an Antrea issue, who is main user of go-ipfix library:
https://github.com/antrea-io/antrea/issues/2336


zyiou (approved):
LGTM

PR #223, opened 2021-07-12
Add convertor interface into the input of Kafka producer

srikartati:
Simplify proto schema management for Kafka producer.
Directly send the convertor interface as input instead of maintaining the map.


pkg/producer/kafka.go
```diff
@@ -47,21 +46,20 @@ type ProducerInput struct {
...
 	// closeCh is required to exit the go routine that captures error messages from
 	// sarama go client.
 	closeCh chan struct{}
 }
 
 func NewKafkaProducer(input ProducerInput) *KafkaProducer {
 	return &KafkaProducer{
-		input:                input,
-		protoSchemaConvertor: convertor.ProtoSchemaConvertor[input.KafkaProtoSchema](),
+		input: input,
```
zyiou, line 62: Not from this PR, do we need to validate whether protoSchemaConvertor is nil since we are using it directly when publishing record?
Also, for other required field like version, it's better to check whether it is assigned with value otherwise the producer will not work well.

zyiou (commented):
Overall LGTM. Only one nit in validating input.

pkg/producer/kafka.go
srikartati: Yeah better to have another check.. user is going to have a check for supported schemas anyway when fetching the protoschema convertors. 
Added check for version too. Thanks.

zyiou (approved):
LGTM

PR #224, opened 2021-07-13
Change Docker images versions used in kafka server deployment yaml 

zyiou:


zyiou:
Close this since it is no longer applicable.
see https://github.com/vmware/go-ipfix/pull/214#issuecomment-879297231

srikartati:
@zyiou I think we need this PR. `docker-proxy-cache` is not exposed to the public, so image pull does not work on github CI. 
I am reopening the PR.

.github/workflows/build.yml
```diff
@@ -25,4 +25,22 @@ jobs:
           DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
         run: |
           echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
-          docker push antrea/ipfix-collector:latest
\ No newline at end of file
+          docker push antrea/ipfix-collector:latest
+  upload-kafka-images:
```
srikartati, line 29: can we make this job daily instead of doing it for every push on main branch? These are not dependent on our code right?

.github/workflows/build.yml
```diff
@@ -25,4 +25,22 @@ jobs:
...
+      - name: Pull, tag and push confluentinc/cp-zookeeper and confluentinc/cp-kafka Docker images to Antrea Docker registry
+        if: ${{ github.repository == 'vmware/go-ipfix' && github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+        env:
+          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+        run: |
+          echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+          docker pull confluentinc/cp-zookeeper:latest
+          docker tag confluentinc/cp-zookeeper:latest antrea/kafka-zookeeper:latest
+          docker push antrea/kafka-zookeeper:latest
```
srikartati, line 43: I will request Antonin for permission to upload these images on to Antrea repo on dockerhub.

srikartati:
Also, please change the image names with project.registry.vmware.com in here: https://github.com/vmware/go-ipfix/blob/main/build/yamls/kafka-flow-collector.yaml

.github/workflows/build.yml
srikartati: @zyiou I am chatting with Antonin. He suggested using a version tag instead of latest. That means we can directly upload the image directly to Antrea repo when the version change is required. In this case, we do not need any job.

I think the hardcoded version for images work here. Do you see any issue?

.github/workflows/build.yml
srikartati: Permissions are provided for these images: `antrea/confluentinc-zookeeper` and `antrea/confluentinc-kafka`

.github/workflows/build.yml
srikartati: The reason is that there are CDN issues when using latest as the tag name.

.github/workflows/build.yml
srikartati: I uploaded following images directly using goipfixbot docker login onto Antrea repo:
`antrea/confluentinc-zookeeper:6.2.0` and `antrea/confluentinc-kafka:6.2.0`

We do not need any github workflow to upload images. We use this versioned image as long as it works for us. When we have to move to a new one, we directly upload the new one. 
Please edit the manifest with the above tags and remove the GitHub workflow. Please add this info in README if possible.

.github/workflows/build.yml
zyiou: Got it. I was trying with Github action schedule function. Will change back since it is no longer needed. Thanks

build/yamls/kafka-flow-collector.yaml
```diff
@@ -77,7 +77,7 @@ spec:
     spec:
       containers:
         - name: kafka
-          image: harbor-repo.vmware.com/dockerhub-proxy-cache/confluentinc/cp-kafka:latest
+          image: antrea/confluentinc-kafka:v6.2.0
```
srikartati, line 80: Let's use the following VMware registry with no rate limiting like other images:
projects.registry.vmware.com/antrea

zyiou: Updated. Thanks!

srikartati (approved):
LGTM

srikartati:
Did the merge immediately as I am consuming this in Antrea CI.

PR #226, opened 2021-07-15
Fix tcp collector server close issue

### LINKED ISSUE #219: Bug when stopping the collector process: TCP and UDP servers are not closed properly.
**Describe the bug**
When the following extra checks are added in `TestTCPCollectingProcess_ReceiveDataRecord` and `TestUDPCollectingProcess_ReceiveDataRecord` after stopping the collector process, then both the tests fail.

```
cp.Stop()
// Check if connection has closed properly or not by creating a new connection.
_, err = net.Dial(collectorAddr.Network(), collectorAddr.String())
assert.Error(t, err) 
```
```
cp.Stop()
// Check if connection has closed properly or not by creating a new connection.
_, err =  net.DialUDP(udpTransport, nil, resolveAddr)
assert.Error(t, err) 
```

**To Reproduce**
Run the tests by adding above lines

**Expected behavior**
Both TCP and UDP servers should be closed properly after calling `Stop` method of the collector process.

**Actual Behavior**
Servers are not closed properly. We can still create the connections.

**Versions**
ToT main branch

**Additional context**

zyiou:
This PR is for making tcp collector server close properly.

partially fixes #219

### DIFF
--- a/pkg/collector/tcp.go
+++ b/pkg/collector/tcp.go
@@ -26,7 +26,7 @@ func (cp *CollectingProcess) startTCPServer() {
 			return
 		}
 		cp.updateAddress(listener.Addr())
-		klog.Infof("Started TLS collecting process on %s", cp.address)
+		klog.Infof("Started TLS collecting process on %s", cp.netAddress)
 	} else {
 		listener, err = net.Listen("tcp", cp.address)
 		if err != nil {
@@ -34,15 +34,14 @@ func (cp *CollectingProcess) startTCPServer() {
 			return
 		}
 		cp.updateAddress(listener.Addr())
-		klog.Infof("Start TCP collecting process on %s", cp.address)
+		klog.Infof("Start TCP collecting process on %s", cp.netAddress)
 	}
-
+	defer listener.Close()
 	go func() {
-		defer listener.Close()
 		for {
 			conn, err := listener.Accept()
 			if err != nil {
-				klog.Errorf("Cannot start collecting process on %s: %v", cp.address, err)
+				klog.Errorf("Cannot accept connection on %s: %v", cp.netAddress, err)
 				return
 			}
 			go cp.handleTCPClient(conn)
--- a/pkg/collector/process_test.go
+++ b/pkg/collector/process_test.go
@@ -131,6 +131,8 @@ func TestTCPCollectingProcess_ReceiveDataRecord(t *testing.T) {
 	}()
 	<-cp.GetMsgChan()
 	cp.Stop()
+	_, err = net.Dial(collectorAddr.Network(), collectorAddr.String())
+	assert.Error(t, err)
 }
 
 func TestUDPCollectingProcess_ReceiveDataRecord(t *testing.T) {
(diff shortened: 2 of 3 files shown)


srikartati (commented):
Yiou, Could you make sure by running the unit test multiple times with a count of 50 to 100?
I am suspecting the proper closing of the listener is not sufficient enough to solve this issue.

zyiou:
For udp server, I didn't find a good way to check its connectivity (Basically if trying `net.Dial` on any valid port on localhost, it will not return error, so following checks do not work for udp)
```
cp.Stop()
// Check if connection has closed properly or not by creating a new connection.
_, err = net.Dial(collectorAddr.Network(), collectorAddr.String())
assert.Error(t, err) 
```
One way we can do is sending response back to client every time a record is received. But it would cause unnecessary traffic.

srikartati:
> For udp server, I didn't find a good way to check its connectivity (Basically if trying `net.Dial` on any valid port on localhost, it will not return error, so following checks do not work for udp)
> 
> ```
> cp.Stop()
> // Check if connection has closed properly or not by creating a new connection.
> _, err = net.Dial(collectorAddr.Network(), collectorAddr.String())
> assert.Error(t, err) 
> ```
> 
> One way we can do is sending response back to client every time a record is received. But it would cause unnecessary traffic.

Are you saying, even though UDP sever port is not open on localhost, we can connect to it? Or Dial does something other than returning error?
Focusing on TCP is good for now.

zyiou:
> Are you saying, even though UDP sever port is not open on localhost, we can connect to it? Or Dial does something other than returning error?

Yes even I input a random address `127.0.0.1:1234` or `127.0.0.1:1000`, it will not return error.

> Focusing on TCP is good for now.

Sure

pkg/collector/tcp.go
```diff
@@ -26,23 +26,22 @@ func (cp *CollectingProcess) startTCPServer() {
...
 			klog.Errorf("Cannot start collecting process on %s: %v", cp.address, err)
 			return
 		}
 		cp.updateAddress(listener.Addr())
-		klog.Infof("Start TCP collecting process on %s", cp.address)
+		klog.Infof("Start TCP collecting process on %s", cp.netAddress)
 	}
-
+	defer listener.Close()
 	go func() {
```
srikartati, line 40: Lets have some logic to exit go routine properly. What do you think?

```
go func(stopCh chan bool) {
	for {
		conn, err := listener.Accept()
		if err != nil {
		     select {
			case <-stopCh:
				return
			default:
				klog.Errorf("Cannot start the connection on the collecting process at %s: %v", cp.address, err)
					return}
		}
		go cp.handleTCPClient(conn)
	}
}(cp.stopChan)
```

srikartati:
Does this have any changes related to UDP? If not, change the title.
I think there listener close bug in UDP too.

pkg/collector/tcp.go
zyiou: With this change, we will need to send multiple signal to stopChan to stop both this goroutine and unblock the rest statements of the function. Do you think that is necessary?

srikartati: You are right. It will be easier if it's not a bool channel.
Could we define the channel as empty strict and close the channel to handle multiple go routines like in a conventional way?

zyiou:
Close this since it is better resolved with proper channel close in #227

PR #227, opened 2021-07-16
Fix bug while stopping the TCP collector process

srikartati:

PR #229, opened 2021-07-17
Properly close tcp and udp collector server

zyiou:
In the previous fix #227, we ensured listener is closed when calling `cp.Stop()` so that when user wants to establish a new connection, they will have error, which is expected. But that is not enough since user can still write to existing connection because corresponding connection was not closed properly. 
At the time we close connection, if user writes to existing connection, collector will still process one record because it already waits at the next iteration of `conn.Read(buff)`. And it takes a little time to close the connection, so set wait for 1 millisecond before sending next record to make the results more steady.
Additionally, since `conn.Write` is not a reliable way for checking whether the connection is open from client side, we add a `CheckConnToCollector` method for clients to check. It works for tcp/tls connections but not udp protocol.

### DIFF
--- a/pkg/collector/process.go
+++ b/pkg/collector/process.go
@@ -57,6 +57,7 @@ type CollectingProcess struct {
 	caCert     []byte
 	serverCert []byte
 	serverKey  []byte
+	wg         sync.WaitGroup
 }
 
 type CollectorInput struct {
@@ -77,7 +78,6 @@ type CollectorInput struct {
 
 type clientHandler struct {
 	packetChan chan *bytes.Buffer
-	errChan    chan bool
 }
 
 func InitCollectingProcess(input CollectorInput) (*CollectingProcess, error) {
@@ -108,8 +108,10 @@ func (cp *CollectingProcess) Start() {
 }
 
 func (cp *CollectingProcess) Stop() {
-	klog.V(4).Info("stopping the collecting process")
 	close(cp.stopChan)
+	// wait for all connections to be safely deleted and returned
+	cp.wg.Wait()
+	klog.Info("stopping the collecting process")
 }
 
 func (cp *CollectingProcess) GetAddress() net.Addr {
@@ -131,7 +133,6 @@ func (cp *CollectingProcess) CloseMsgChan() {
 func (cp *CollectingProcess) createClient() *clientHandler {
 	return &clientHandler{
 		packetChan: make(chan *bytes.Buffer),
-		errChan:    make(chan bool),
 	}
 }
 
@@ -151,14 +152,6 @@ func (cp *CollectingProcess) getClientCount() int {
 	cp.mutex.RLock()
 	defer cp.mutex.RUnlock()
 	return len(cp.clients)
-}
-
-func (cp *CollectingProcess) closeAllClients() {
-	cp.mutex.Lock()
-	defer cp.mutex.Unlock()
-	for _, client := range cp.clients {
-		client.errChan <- true
-	}
 }
 
 func (cp *CollectingProcess) decodePacket(packetBuffer *bytes.Buffer, exportAddress string) (*entities.Message, error) {
--- a/pkg/collector/process_test.go
+++ b/pkg/collector/process_test.go
@@ -120,18 +120,24 @@ func TestTCPCollectingProcess_ReceiveDataRecord(t *testing.T) {
 
 	// wait until collector is ready
 	waitForCollectorReady(t, cp)
+	var conn net.Conn
 	collectorAddr := cp.GetAddress()
 	go func() {
-		conn, err := net.Dial(collectorAddr.Network(), collectorAddr.String())
+		conn, err = net.Dial(collectorAddr.Network(), collectorAddr.String())
 		if err != nil {
 			t.Errorf("Cannot establish connection to %s", collectorAddr.String())
 		}
-		defer conn.Close()
 		conn.Write(validDataPacket)
 	}()
 	<-cp.GetMsgChan()
 	cp.Stop()
-	// Check if connection has closed properly or not by creating a new connection.
+	// Check if connection has closed properly or not by trying to write to it
+	_, err = conn.Write(validDataPacket)
+	time.Sleep(time.Millisecond)
+	_, err = conn.Write(validDataPacket)
+	assert.Error(t, err)
+	conn.Close()
+	// Check if connection has closed properly or not by trying to create a new connection.
 	_, err = net.Dial(collectorAddr.Network(), collectorAddr.String())
 	assert.Error(t, err)
 }
@@ -156,15 +162,21 @@ func TestUDPCollectingProcess_ReceiveDataRecord(t *testing.T) {
 		t.Errorf("UDP Address cannot be resolved.")
 	}
 	go func() {
-		conn, err := net.DialUDP(udpTransport, nil, resolveAddr)
-		if err != nil {
-			t.Errorf("UDP Collecting Process does not start correctly.")
-		}
-		defer conn.Close()
-		conn.Write(validDataPacket)
+		<-cp.GetMsgChan()
+		cp.Stop()
 	}()
-	<-cp.GetMsgChan()
-	cp.Stop()
+	conn, err := net.DialUDP(udpTransport, nil, resolveAddr)
+	if err != nil {
+		t.Errorf("UDP Collecting Process does not start correctly.")
+	}
+	conn.Write(validDataPacket)
+	time.Sleep(time.Millisecond)
+	// Check if connection has closed properly or not by trying to write to it
+	_, _ = conn.Write(validDataPacket)
+	time.Sleep(time.Millisecond)
+	_, err = conn.Write(validDataPacket)
+	assert.Error(t, err)
+	conn.Close()
 }
 
 func TestTCPCollectingProcess_ConcurrentClient(t *testing.T) {
@@ -188,7 +200,7 @@ func TestTCPCollectingProcess_ConcurrentClient(t *testing.T) {
 			t.Errorf("Cannot establish connection to %s", collectorAddr.String())
 		}
 		time.Sleep(time.Millisecond)
-		assert.Equal(t, 2, cp.getClientCount(), "There should be two tcp clients.")
+		assert.GreaterOrEqual(t, cp.getClientCount(), 2, "There should be at least two tcp clients.")
 		cp.Stop()
 	}()
 	cp.Start()
@@ -201,32 +213,25 @@ func TestUDPCollectingProcess_ConcurrentClient(t *testing.T) {
 	// wait until collector is ready
 	waitForCollectorReady(t, cp)
 	collectorAddr := cp.GetAddress()
-	go func() {
-		resolveAddr, err := net.ResolveUDPAddr(collectorAddr.Network(), collectorAddr.String())
-		if err != nil {
-			t.Errorf("UDP Address cannot be resolved.")
-		}
-		conn, err := net.DialUDP(udpTransport, nil, resolveAddr)
-		if err != nil {
-			t.Errorf("UDP Collecting Process does not start correctly.")
-		}
-		defer conn.Close()
-		conn.Write(validTemplatePacket)
-	}()
-	go func() {
-		resolveAddr, err := net.ResolveUDPAddr(collectorAddr.Network(), collectorAddr.String())
-		if err != nil {
-			t.Errorf("UDP Address cannot be resolved.")
-		}
-		conn, err := net.DialUDP(udpTransport, nil, resolveAddr)
-		if err != nil {
-			t.Errorf("UDP Collecting Process does not start correctly.")
-		}
-		defer conn.Close()
-		conn.Write(validTemplatePacket)
-		time.Sleep(time.Millisecond)
-		assert.Equal(t, 2, cp.getClientCount(), "There should be two udp clients.")
-	}()
+	resolveAddr, err := net.ResolveUDPAddr(collectorAddr.Network(), collectorAddr.String())
+	if err != nil {
+		t.Errorf("UDP Address cannot be resolved.")
+	}
+	conn1, err := net.DialUDP(udpTransport, nil, resolveAddr)
+	if err != nil {
+		t.Errorf("UDP Collecting Process does not start correctly.")
+	}
+	defer conn1.Close()
+	conn1.Write(validTemplatePacket)
+
+	conn2, err := net.DialUDP(udpTransport, nil, resolveAddr)
+	if err != nil {
+		t.Errorf("UDP Collecting Process does not start correctly.")
+	}
+	defer conn2.Close()
+	conn2.Write(validTemplatePacket)
+	time.Sleep(time.Millisecond)
+	assert.GreaterOrEqual(t, cp.getClientCount(), 2, "There should be at least two udp clients.")
 	// there should be two messages received
 	<-cp.GetMsgChan()
 	<-cp.GetMsgChan()
@@ -364,6 +369,8 @@ func TestTLSCollectingProcess(t *testing.T) {
 	// wait until collector is ready
 	waitForCollectorReady(t, cp)
 	collectorAddr := cp.GetAddress()
+	var conn net.Conn
+	var config *tls.Config
 	go func() {
 		roots := x509.NewCertPool()
 		ok := roots.AppendCertsFromPEM([]byte(test.FakeCACert))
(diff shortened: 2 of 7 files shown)


pkg/collector/tcp.go
```diff
@@ -49,24 +49,20 @@ func (cp *CollectingProcess) startTCPServer() {
 					return
 				}
 			}
-			go cp.handleTCPClient(conn, stopCh)
+			go cp.handleTCPClient(conn)
```
srikartati, line 52: >When closing connections, we are not supposed to depend on cp.stopChan because it will cause firstly exit from startTCPServer() before we are able to take care of closing connection in handleTCPClient().

What is the reason for not closing the connection properly? We do have to defer call for connection close in L62 in `handleTCPClient` method.
Is it because the connection is created in other go routine?

pkg/collector/tcp.go
```diff
@@ -49,24 +49,20 @@ func (cp *CollectingProcess) startTCPServer() {
...
 		}
 	}(cp.stopChan)
 	<-cp.stopChan
```
srikartati, line 55: Don't you want to include `CloseAllClients` method here?

srikartati (commented):
Does this patch resolve the flow aggregator restart issue?

pkg/collector/tcp.go
zyiou: I think so. when we close `stopChan`, the signal in `stopChan` is not caught by `handleTCPClient` so `handleTCPClient` is not able to return correctly before this function `startTCPServer()` exits. Also detect the same condition at Antrea side.

pkg/collector/tcp.go
zyiou: `CloseAllClients` is put in the `Stop()` function

zyiou:
> Does this patch resolve the flow aggregator restart issue?

Not completely. As I mentioned, signal from `close(stopChan)` is often not caught by newly created go routine before main function exits, so when flow aggregator stops, `stopChan` [here](https://github.com/antrea-io/antrea/blob/f80cbd9970404f1dc586ef4dacc7275d3da86d96/pkg/flowaggregator/flowaggregator.go#L365) does not get executed for most of times so that collector does not stop correctly. Will create a fix in Antrea too.

pkg/collector/tcp.go
srikartati: Maybe this is because of the reason that we exit the parent go routine (inline one that calls handleTCPClient) before all the other client go routines exit. Could we use wg.Add and wg.Wait here to resolve that. 
I think that may work. What do you think?

zyiou: Yes, using wait group to ensure connections are safely closed before exiting makes sense.

pkg/collector/process_test.go
```diff
@@ -120,19 +120,21 @@ func TestTCPCollectingProcess_ReceiveDataRecord(t *testing.T) {
...
 			t.Errorf("Cannot establish connection to %s", collectorAddr.String())
 		}
-		defer conn.Close()
 		conn.Write(validDataPacket)
 	}()
 	<-cp.GetMsgChan()
 	cp.Stop()
-	// Check if connection has closed properly or not by creating a new connection.
-	_, err = net.Dial(collectorAddr.Network(), collectorAddr.String())
+	// Check if connection has closed properly or not by writing to it
```
srikartati, line 134: We need to check the connection is closed properly and the collector server is properly stopped or not (server close), right? For the latter, we can still use the `Dial` as the check.

pkg/collector/process.go
```diff
@@ -144,6 +148,12 @@ func (cp *CollectingProcess) addClient(address string, client *clientHandler) {
 func (cp *CollectingProcess) deleteClient(name string) {
 	cp.mutex.Lock()
 	defer cp.mutex.Unlock()
+	client, exist := cp.clients[name]
+	if !exist {
+		klog.Warning("client does not exist or already been deleted.")
+		return
+	}
+	close(client.errChan)
```
srikartati, line 156: do we need both channel and wait group?

pkg/collector/process.go
```diff
@@ -156,8 +166,9 @@ func (cp *CollectingProcess) getClientCount() int {
 func (cp *CollectingProcess) closeAllClients() {
 	cp.mutex.Lock()
 	defer cp.mutex.Unlock()
-	for _, client := range cp.clients {
-		client.errChan <- true
+	for name, client := range cp.clients {
+		close(client.errChan)
+		delete(cp.clients, name)
```
srikartati, line 171: Presuming that this is required, we can use deleteClient method right?

pkg/collector/process_test.go
```diff
@@ -225,11 +227,12 @@ func TestUDPCollectingProcess_ConcurrentClient(t *testing.T) {
 		defer conn.Close()
 		conn.Write(validTemplatePacket)
 		time.Sleep(time.Millisecond)
-		assert.Equal(t, 2, cp.getClientCount(), "There should be two udp clients.")
+		assert.GreaterOrEqual(t, cp.getClientCount(), 2, "There should be at least two udp clients.")
 	}()
 	// there should be two messages received
 	<-cp.GetMsgChan()
 	<-cp.GetMsgChan()
+	time.Sleep(2 * time.Millisecond)
```
srikartati, line 235: what is the reason for sleep here?

pkg/collector/process_test.go
```diff
@@ -391,6 +395,10 @@ func TestTLSCollectingProcess(t *testing.T) {
 	<-cp.GetMsgChan()
 	cp.Stop()
 	assert.NotNil(t, cp.templatesMap[1], "TLS Collecting Process should receive and store the received template.")
+	// Check if connection has closed properly or not by writing to it
```
srikartati, line 398: Here as well, it is better to check if server is closed properly or not in addition to the connection.

pkg/collector/udp.go
```diff
@@ -57,7 +55,7 @@ func (cp *CollectingProcess) startUDPServer() {
 			return
 		}
 		cp.updateAddress(listener.Addr())
-		klog.Infof("Start dtls collecting process on %s", cp.address)
+		klog.Infof("Start dtls collecting process on %s", cp.netAddress)
 		conn, err = listener.Accept()
```
srikartati, line 59: We should close the listener.

pkg/collector/udp.go
```diff
@@ -108,24 +106,21 @@ func (cp *CollectingProcess) startUDPServer() {
...
-func (cp *CollectingProcess) handleUDPClient(address net.Addr, wg *sync.WaitGroup) {
+func (cp *CollectingProcess) handleUDPClient(address net.Addr) {
 	if _, exist := cp.clients[address.String()]; !exist {
 		client := cp.createClient()
 		cp.addClient(address.String(), client)
-		wg.Add(1)
-		defer wg.Done()
+		cp.wg.Add(1)
 		go func() {
+			defer cp.wg.Done()
```
srikartati, line 123: If we are handling UDP server stop as well, we should have a test in UDP tests. Or let's handle it in a different PR?

srikartati:
Not sure if we should have a solution that involves arbitrary sleep in the user code: https://github.com/antrea-io/antrea/pull/2431/files#diff-7ead2c15ecd071e2b67d6243a9a79aeaa6b65dd9d7f44fa511e20e5b564839a6R122

IMHO, we should avoid this kind of usage. If required, we can take input from the user to handle the synchronization.

zyiou:
@srikartati please hold on the review for this PR. There is still instability at Antrea side. will request for review when I figure out the fix. Thanks!

pkg/collector/udp.go
zyiou: Done. Thanks

pkg/collector/udp.go
zyiou: You are right. Added it in current PR since it is simple test. But checking connection openness by reading does not work for udp, so in this PR we only solves the closing of udp.

pkg/collector/process_test.go
zyiou: Makes sense. Updated. Thanks!

pkg/collector/process.go
```diff
@@ -108,8 +109,11 @@ func (cp *CollectingProcess) Start() {
 }
 
 func (cp *CollectingProcess) Stop() {
-	klog.V(4).Info("stopping the collecting process")
 	close(cp.stopChan)
+	cp.closeAllClients()
```
srikartati, line 113: Looks like we are using this only for UDP.
Wondering if UDP server and client connections can be gracefully shutdown without this.
If not, we should only call this for UDP and skip for TCP as it is not used there.

pkg/exporter/process.go
```diff
@@ -204,6 +207,16 @@ func (ep *ExportingProcess) CloseConnToCollector() {
 	}
 }
 
+// CheckConnToCollector checks whether the connection from exporter is still open
+// by trying to read from connection. Closed connection will return EOF from read.
+func (ep *ExportingProcess) CheckConnToCollector() bool {
```
srikartati, line 212: I see that we are piggybacking on the export goroutine in Antrea. This adds some overhead (including some delay) in the exporter goroutine. Prefer to handle this in a separate goroutine in the library. You probably want to avoid locking on the client connection.

Did you get a chance to look at this connection struct from net library: https://pkg.go.dev/net#TCPConn.SetKeepAlive ?

pkg/exporter/process_test.go
```diff
@@ -590,3 +590,26 @@ func TestExportingProcess_GetMsgSizeLimit(t *testing.T) {
 	t.Logf("Created exporter connecting to local server with address: %s", conn.LocalAddr().String())
 	assert.Equal(t, entities.DefaultUDPMsgSize, exporter.GetMsgSizeLimit())
 }
+
+func TestExportingProcess_CheckConnToCollector(t *testing.T) {
```
srikartati, line 594: can we add this in our integration test of exporter and collector?

pkg/exporter/process.go
zyiou: I have considered using `setKeepAlive`. In that case we will need to keep a `TCPConn` for tcp and `net.Conn` for udp instead of `net.Conn` in our collecting process and handle write and close differently for tcp and udp connections. It does not have read overhead or blocking the connection. 
Using current check will have multiple reads to one byte memory. If we use goroutine for checking the connection, it will not block the connection too.
These two solutions look similar to me. Do you have any preference?

srikartati: I see. In the first option, there is quite a bit of refactoring needed. The benefit of this option is utilizing the standardized library code that takes care of all scenarios.
I am ok with the second option too. As the read is done in a separate goroutine, the overhead is not so much. If we are using the lock on the connection, we might be blocking the connection while doing either read or write. Isn't that correct? For reading on the connection, we can set a deadline so that we will not block for a long time. 
Any thoughts on the periodic interval for this goroutine?

zyiou: Got it. Haven't thought about using lock to prevent concurrent write and read. In that case, I think refactoring to TCPConn will be more convenient since we don't need to handle the locking and clean up for goroutine (also good for encapsulate the checking for connection openness). Just had a draft test on refactoring. Will also test at Antrea side to see whether it will work well.

pkg/collector/tcp.go
```diff
@@ -49,24 +51,23 @@ func (cp *CollectingProcess) startTCPServer() {
...
 	buff := make([]byte, cp.maxBufferSize)
-	for {
-		select {
-		case <-stopCh:
-			cp.deleteClient(address)
-			return
-		default:
+	defer cp.wg.Done()
+	defer conn.Close()
+	go func() {
```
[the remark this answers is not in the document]
zyiou: After connection is closed, `conn.Read(buff)` will have `io.EOF` error then it will return from the goroutine and finish corresponding waitgroup. Reason we cannot use `for{ select{ case: <-cp.stopChan default:}}` is that each iteration we will enter `conn.Read` and wait for input there, the stop message cannot be caught.

pkg/collector/process.go
zyiou: Went through its usage and found that we can remove the usage of `client.errChan`. Done. Thanks!

pkg/exporter/process.go
zyiou: Toke a closer look and try on refactoring  to `TCPConn`, it does not work well with tls settings and will have multiple retries before return. Using `Read` can directly return the result and we can customize the periodic interval. I re-write the way of checking so that not much change will be needed in Antrea. Please take a look Thanks!

pkg/test/exporter_collector_test.go
```diff
@@ -267,6 +271,51 @@ func testExporterToCollector(address net.Addr, isSrcNode, isIPv6 bool, isMultipl
...
+		t.Fatalf("Got error when connecting to %s", cp.GetAddress().String())
+	}
+	templateID := export.NewTemplateID()
+	templateSet := createTemplateSet(templateID, false)
+	// Send template record
+	export.SendSet(templateSet)
+	<-cp.GetMsgChan()
+	cp.CloseMsgChan()
+	cp.Stop() // Close collecting process
+	time.Sleep(2 * time.Millisecond)
```
srikartati, line 314: Instead of sleep, can we add wait.Poll for the next line with a timeout?

pkg/exporter/process.go
```diff
@@ -127,6 +133,30 @@ func InitExportingProcess(input ExporterInput) (*ExportingProcess, error) {
 		templateRefCh:   make(chan struct{}),
 	}
 
+	// Start a goroutine for checking whether connection to collector is still open
+	if input.CollectorProtocol == "tcp" {
```
srikartati, line 137: is this not an issue with UDP?

pkg/exporter/process.go
```diff
@@ -29,6 +30,9 @@ import (
 )
 
 const startTemplateID uint16 = 255
+const defaultCheckConnInterval = 10 * time.Second
+
+var oneByte = make([]byte, 1)
```
srikartati, line 35: why global variable?

pkg/exporter/process.go
```diff
@@ -50,6 +54,7 @@ type ExportingProcess struct {
 	templatesMap    map[uint16]templateValue
 	templateRefCh   chan struct{}
 	mutex           sync.Mutex
```
srikartati, line 56: change this mutex name to something more appropriate like the new one?

pkg/collector/process.go
```diff
@@ -108,8 +108,10 @@ func (cp *CollectingProcess) Start() {
 }
 
 func (cp *CollectingProcess) Stop() {
-	klog.V(4).Info("stopping the collecting process")
 	close(cp.stopChan)
+	// wait for all connections to be safely deleted and returned
+	cp.wg.Wait()
+	klog.V(4).Info("stopping the collecting process")
```
srikartati, line 114: Remove the verbosity level? I think this doesn't happen that often.

srikartati (commented):
New approach looks good to me. Few more comments.

pkg/exporter/process.go
zyiou: For udp, we cannot check whether the connection is open by read (probably because of its stateless property), so only set the goroutine for tcp.

pkg/exporter/process.go
zyiou: I don't want to recreate it every time when testing the connection. Moved it to ExportingProcess struct then.

pkg/exporter/process.go
```diff
@@ -125,6 +130,31 @@ func InitExportingProcess(input ExporterInput) (*ExportingProcess, error) {
 		pathMTU:         input.PathMTU,
 		templatesMap:    make(map[uint16]templateValue),
 		templateRefCh:   make(chan struct{}),
+		oneByte:         make([]byte, 1),
```
srikartati, line 133: >I don't want to recreate it every time when testing the connection. Moved it to ExportingProcess struct then.

This seems like a local variable required by a method rather than the struct. 
How about taking in this as an argument in `checkConnToCollector` and defining it along with the inline goroutine that is calling this method. If the argument is a pointer we could make sure it is only defined and allocated once.

pkg/exporter/process.go
srikartati: Will the connection be reset quickly in the case of UDP or we do write on the closed connection too?

pkg/exporter/process.go
zyiou: Makes sense to me. Updated. Thanks!

pkg/exporter/process.go
zyiou: Had some experiments on Antrea, the connection will be reset quickly so we cannot check the right connection for sending the records.

srikartati (commented):
One comment about stray code. Otherwise LGTM.

pkg/exporter/process.go
```diff
@@ -49,7 +51,9 @@ type ExportingProcess struct {
 	pathMTU         int
 	templatesMap    map[uint16]templateValue
 	templateRefCh   chan struct{}
-	mutex           sync.Mutex
+	templateMutex   sync.Mutex
+	connMutex       sync.Mutex
+	oneByte         []byte
```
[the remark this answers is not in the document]
zyiou: Forgot to remove it. Done. Thanks!

srikartati (approved):
LGTM

PR #230, opened 2021-07-23
Improve Kafka consumer application

srikartati:
We should not assume that Kafka broker is already present and ready. 
Have a retry mechanism for consumer initialization.

### DIFF
--- a/cmd/consumer/consumer.go
+++ b/cmd/consumer/consumer.go
@@ -20,10 +20,12 @@ import (
 	"os/signal"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/Shopify/sarama"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/component-base/logs"
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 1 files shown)


srikartati:
Verified on Antrea with the updated local Kafka consumer image.

cmd/consumer/consumer.go
```diff
@@ -85,14 +90,24 @@ func run() error {
 		KafkaLogErrors:    true,
 	}
 	consumer := consumer.NewKafkaConsumer(input)
+	consumerInitError := make(chan struct{})
```
zyiou, line 93: Are we supposed to pass some error into this channel in goroutine?

srikartati: I was trying something else and figured it's not needed. Forgot to remove all the related code.

cmd/consumer/consumer.go
```diff
@@ -86,9 +88,18 @@ func run() error {
 	}
 	consumer := consumer.NewKafkaConsumer(input)
 	go func() {
-		if err := consumer.InitSaramaConsumer(); err != nil {
+		err := wait.PollImmediateInfinite(500*time.Millisecond, func() (bool, error) {
+			if err := consumer.InitSaramaConsumer(); err != nil {
+				return false, nil
```
zyiou, line 93: Not sure whether it will affect condition of Poll, we may want to `return false, err` here

srikartati: If we return error, then we error out from the wait method. This behavior will be same as before.

zyiou: Got it. Thanks for clarifying

zyiou (approved):
LGTM

PR #234, opened 2021-07-30
add delimiter when sending tcp message

zyiou:


pkg/exporter/process.go
```diff
@@ -282,25 +284,36 @@ func (ep *ExportingProcess) createAndSendMsg(set entities.Set) (int, error) {
...
+	//	index += len
+	//}
+
+	message := make(map[string]interface{})
+	record := set.GetRecords()[0]
+	for _, element := range record.GetOrderedElementList() {
+		message[element.Element.Name] = element.Value
 	}
+	message["@timestamp"] = time.Now().Format(time.RFC3339)
+	writer := bytes.NewBuffer(make([]byte, 0, 1.2*300 * msgLen))
```
srikartati, line 303: we don't need to have such capacity. We could directly specify msgLen like below:

`writer := bytes.NewBuffer(make([]byte, msgLen))`

I know it is a draft. Just pitching in..

PR #236, opened 2021-08-05
Fix collector close issue and add perf test MultipleExportersToCollector

zyiou:
This PR fixes a collector close issue that when decoding error happens, we should break instead of return.
It also adds perf test for benchmarking the case when multiple exporters are talking to one collector, collector receives and decodes the message, then stops. Some of the common functions are moved to another folder for reuse.
It adjust maximum tries before deleting idle records in aggregator from 5 to 2 to better accommodate with scale cluster when number of connections accumulates greatly and waits for retries.


pkg/collector/tcp.go
```diff
@@ -86,20 +86,20 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
 				if err != nil {
 					klog.Error(err)
 					cp.deleteClient(address)
-					return
+					break
```
srikartati, line 89: If we are deleting the client, shouldn't we exit the handTCPclient function?

pkg/collector/tcp.go
```diff
@@ -86,20 +86,20 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
...
 				}
 				if size < length {
 					klog.Errorf("Message length %v is larger than size read from buffer %v", length, size)
 					cp.deleteClient(address)
-					return
+					break
```
srikartati, line 94: Here we do not have to delete the client.

pkg/collector/tcp.go
```diff
@@ -86,20 +86,20 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
...
 				}
 				size = size - length
 				// get the message here
 				message, err := cp.decodePacket(bytes.NewBuffer(buffBytes[0:length]), address)
 				if err != nil {
 					klog.Error(err)
 					cp.deleteClient(address)
-					return
+					break
```
srikartati, line 102: same as above

pkg/test/collector_perf_test.go
```diff
@@ -0,0 +1,121 @@
...
+			b.Errorf("cannot start exporting process: %v", err)
+		}
+		templateID := exporter.NewTemplateID()
+		exporter.SendSet(createTemplateSet(templateID, false))
+		exporter.SendSet(createDataSet(templateID, true, false, true))
+	}
+	count := 0
+	for range cp.GetMsgChan() {
+		count++
+		if count == 2000 {
```
srikartati, line 86: We can configure this number. Make each exporter send 10K records?

srikartati (commented):
Thanks for adding the performance test. 
Do you see any bottlenecks in memory/cpu profiling?

pkg/collector/tcp.go
zyiou: yes you are right, we should not delete client here. Updated.

pkg/test/collector_perf_test.go
zyiou: Sure. When I make each exporter send 10K records, found that `Message length 95 is larger than size read from buffer 45` you have previously mentioned. I feel it should be related to the Logstash tcp decoding issue. Working on triaging it.

zyiou:
I set number of exporter to 100 and number of records per exporter to 600. Under this case the test will only run once and the time is 1.847s. This setting is similar to what we want to test ~30k/s. There is a bottleneck in both CPU and memory profile, which is `NewInfoElementWithValue` method.
CPU
```
go tool pprof profile.out                                                                                                                    
Type: cpu
Time: Aug 6, 2021 at 5:19pm (PDT)
Duration: 1.61s, Total samples = 1.64s (101.96%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 1450ms, 88.41% of 1640ms total
Showing top 10 nodes out of 135
      flat  flat%   sum%        cum   cum%
     670ms 40.85% 40.85%      670ms 40.85%  syscall.syscall
     460ms 28.05% 68.90%      460ms 28.05%  runtime.kevent
      90ms  5.49% 74.39%       90ms  5.49%  runtime.madvise
      50ms  3.05% 77.44%      500ms 30.49%  runtime.netpoll
      40ms  2.44% 79.88%       40ms  2.44%  runtime.usleep
      30ms  1.83% 81.71%       30ms  1.83%  github.com/vmware/go-ipfix/pkg/entities.NewInfoElementWithValue
      30ms  1.83% 83.54%       30ms  1.83%  runtime.nanotime1
      30ms  1.83% 85.37%       30ms  1.83%  runtime.procyield
      30ms  1.83% 87.20%       30ms  1.83%  runtime.pthread_kill
      20ms  1.22% 88.41%       50ms  3.05%  runtime.mallocgc
```
Memory
```
go tool pprof memprofile.out
Type: alloc_space
Time: Aug 6, 2021 at 5:20pm (PDT)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 208.08MB, 91.74% of 226.81MB total
Dropped 21 nodes (cum <= 1.13MB)
Showing top 10 nodes out of 32
      flat  flat%   sum%        cum   cum%
      61MB 26.90% 26.90%       61MB 26.90%  github.com/vmware/go-ipfix/pkg/entities.NewInfoElementWithValue (inline)
   41.51MB 18.30% 45.20%    73.01MB 32.19%  github.com/vmware/go-ipfix/pkg/test.getDataRecordElements
   31.50MB 13.89% 59.09%       92MB 40.57%  github.com/vmware/go-ipfix/pkg/collector.(*CollectingProcess).decodeDataSet
   27.50MB 12.13% 71.21%    27.50MB 12.13%  github.com/vmware/go-ipfix/pkg/entities.NewDataRecord (inline)
   10.25MB  4.52% 75.73%   116.89MB 51.54%  github.com/vmware/go-ipfix/pkg/collector.(*CollectingProcess).handleTCPClient.func1
      10MB  4.41% 80.14%       10MB  4.41%  github.com/vmware/go-ipfix/pkg/entities.DecodeToIEDataType
       8MB  3.53% 83.67%        8MB  3.53%  github.com/vmware/go-ipfix/pkg/entities.NewSet (inline)
    6.50MB  2.87% 86.53%     6.50MB  2.87%  github.com/vmware/go-ipfix/pkg/entities.(*dataRecord).GetBuffer
    6.50MB  2.87% 89.40%     6.50MB  2.87%  bytes.NewBuffer (inline)
    5.32MB  2.35% 91.74%     5.32MB  2.35%  github.com/vmware/go-ipfix/pkg/collector.(*CollectingProcess).handleTCPClient
```

Also, about tcp decoding issue. Assume length of record is 700 and we send 100 consecutive records to a collector with 65535 read buffer. Chances are the buffer will read 70000 length of bytes, larger than 65535. So the data will be truncated and cannot be correctly decoded. In Flow Aggregator, we were sometimes sending 10000 consecutive records in scale setting, that probably cause the data loss in tcp input plugin. (I am under impression the tcp read buffer is around 65535 in Logstash)

pkg/collector/tcp.go
srikartati: If length cannot be retrieved, then shouldn't we empty the buffer before reading on the connection again.

pkg/collector/tcp.go
```diff
@@ -85,21 +85,18 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
...
 				}
 				size = size - length
 				// get the message here
 				message, err := cp.decodePacket(bytes.NewBuffer(buffBytes[0:length]), address)
 				if err != nil {
 					klog.Error(err)
-					cp.deleteClient(address)
-					return
+					break
```
srikartati, line 99: same here

zyiou: Do you mean `buffBytes` or `buff`? In my understanding reading on connections should be an overwrite to the buffer and size can determine how long we need to read and decode.

srikartati: I meant buffBytes. I know that you are reinitializing, but It is better to make it nil explicitly as we are allocating the buffer again and again.

zyiou: Make sense. Updated.

srikartati (commented):
LGTM
I think PR needs a rebase.

srikartati (approved):
LGTM

PR #237, opened 2021-08-07
Add support for sending JSON record

zyiou:


pkg/exporter/process.go
```diff
@@ -84,7 +89,9 @@ type ExporterInput struct {
 func InitExportingProcess(input ExporterInput) (*ExportingProcess, error) {
 	var conn net.Conn
 	var err error
-
+	if input.CollectorProtocol == "udp" && input.SendJSONRecord {
```
srikartati, line 92: Hi Yiou, I thought about this more. Even though UDP packet length is large here.. IP fragmentation happens and packet get reassembled in Logstash instance. There is possibility of record loss, but that is ok. Logstash received records fine with IP fragmentation. I checked with large records and UDP transport. 
You can remove this check.

srikartati (commented):
LGTM except for comment on UDP transport.
Hope we got a chance to test this on local setup.

pkg/exporter/process.go
zyiou: Got it. Updated. will add some local tests later.

srikartati: I did not mean local unit tests. I just meant a test on local setup with Antrea+local ELK flow collector.

pkg/exporter/process.go
```diff
@@ -31,6 +33,7 @@ import (
 
 const startTemplateID uint16 = 255
 const defaultCheckConnInterval = 10 * time.Second
+const msgBufferLen = 4000
```
srikartati, line 36: we may end up needing a bigger record if metadata in ipfix is long, right?

zyiou: Correct. Do you have any suggested size in mind?

srikartati: May be allocate bigger size with a factor based on the length of ipfix message length?

pkg/exporter/process.go
```diff
@@ -305,6 +316,32 @@ func (ep *ExportingProcess) createAndSendMsg(set entities.Set) (int, error) {
 	return bytesSent, nil
 }
 
+// createAndSendJSONMsg takes in a set as input, creates the JSON record, and sends it out.
+func (ep *ExportingProcess) createAndSendJSONMsg(set entities.Set) (int, error) {
+	var bytesSent int
+	for _, record := range set.GetRecords() {
+		records := make(map[string]interface{})
```
kais66, line 323: nit: is "records" an accurate variable name? It seems to be a collection of fields from records.

kais66:
Can we benchmark the record sizes for IPFIX and JSON formats? Some data for a few sample records should suffice. This is to get an understanding how much overhead in terms of additional bytes JSON format is causing.

pkg/exporter/process.go
zyiou: Updated. Thanks

pkg/exporter/process.go
```diff
@@ -305,6 +327,32 @@ func (ep *ExportingProcess) createAndSendMsg(set entities.Set) (int, error) {
...
+		elements := make(map[string]interface{})
+		for _, element := range record.GetOrderedElementList() {
+			elements[element.Element.Name] = element.Value
+		}
+		message := make(map[string]interface{}, 2)
+		message["ipfix"] = elements
+		message["@timestamp"] = time.Now().Format(time.RFC3339)
+		writer := bytes.NewBuffer(make([]byte, 0, ep.jsonBufferLen))
+		encoder := json.NewEncoder(writer)
+		encoder.Encode(message)
```
srikartati, line 343: What happens if we message buffer in the encoder? Does it return error?

zyiou: you mean encode(writer)?

srikartati: I mean encoder.Encode function. Looked into it. I think we need to handle the error here.
https://github.com/square/go-jose/blob/v2.6.0/json/stream.go#L188

srikartati (approved):
LGTM

PR #238, opened 2021-08-07
Avoid sending duplicate records at aggregator

zyiou:
There are chances that current implementation will send duplicate records at the same time when time is the least common multiple of active timeout and inactive timeout. This will cause throughput calculation at Logstash to have divided by 0 issue. Also it can avoid sending duplicate records at both active & inactive timeout.

### DIFF
--- a/pkg/intermediate/aggregate.go
+++ b/pkg/intermediate/aggregate.go
@@ -254,9 +254,10 @@ func (a *AggregationProcess) ForAllExpiredFlowRecordsDo(callback FlowKeyRecordMa
 		}
 		// Reset the expireTime for the popped item and push it to the priority queue.
 		if pqItem.activeExpireTime.Before(currTime) {
-			// Reset the active expire timeout and push the record into priority
-			// queue.
+			// Reset the active and inactive expire timeout and push the record into
+			// priority queue.
 			pqItem.activeExpireTime = currTime.Add(a.activeExpiryTimeout)
+			pqItem.inactiveExpireTime = currTime.Add(a.inactiveExpiryTimeout)
 			heap.Push(&a.expirePriorityQueue, pqItem)
 		}
 	}


pkg/intermediate/aggregate.go
```diff
@@ -254,9 +254,10 @@ func (a *AggregationProcess) ForAllExpiredFlowRecordsDo(callback FlowKeyRecordMa
 		}
 		// Reset the expireTime for the popped item and push it to the priority queue.
 		if pqItem.activeExpireTime.Before(currTime) {
-			// Reset the active expire timeout and push the record into priority
-			// queue.
+			// Reset the active and inactive expire timeout and push the record into
+			// priority queue.
 			pqItem.activeExpireTime = currTime.Add(a.activeExpiryTimeout)
+			pqItem.inactiveExpireTime = currTime.Add(a.inactiveExpiryTimeout)
```
srikartati, line 260: Even though, we reset here, we may not receive any new records for this flow. We are just delaying the inactive flow record by some time. It will still be duplicate and flow end seconds will still be the same in the received record.
Isn't the observation correct?

zyiou: It can avoid the duplications when the flow is active for a long time. The case you mentioned is when we send the last record for an inactive flow right? To deal with that case, maybe if we find the least common multiple is rather small, we can hard code 1 more second to inactive flow timeout. For example, active timeout 60s, inactive timeout 90s, we can set inactive timeout to 91s, so that we would not send duplicate records at time 180s.

srikartati: Could you elaborate on how the duplications are avoided for long-lived flow? We are delaying the records for sometime, but it may still get deleted if we do not receive the records in that specific time right?

In my mind, flow is deemed active only when we receive new records for that flow, and then we can reset the inactive expiry time. By sending active flow record, we cannot say that flow is active.

zyiou: Took a closer look at the logic. I saw that as you said we have already updated the inactive timeout when receiving records. I thought the logic was not there in the first place. 
When testing in scale setup, I found some records from same flow have same flowEndSeconds. Deduced from that condition, thought it is caused by sending active and inactive flow record at the same time.

srikartati: Oh "flowEndSeconds" is added by the Flow Exporter. At the Flow Aggregator we just update that if it's latest.

zyiou:
Close it as the issue is been taken care of in this PR https://github.com/antrea-io/antrea/pull/2559

PR #240, opened 2021-08-10
Fix consecutive read issues on TCP connection in the collector process

srikartati:
Partial buffer needs to be retained when doing reads in
a loop; we should not discard it.
Modify the decoding code to make the buffer usage efficient.

Added unit test to verify that.

### DIFF
--- a/pkg/collector/process.go
+++ b/pkg/collector/process.go
@@ -15,6 +15,7 @@
 package collector
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/binary"
 	"fmt"
@@ -160,10 +161,9 @@ func (cp *CollectingProcess) getClientCount() int {
 }
 
 func (cp *CollectingProcess) decodePacket(packetBuffer *bytes.Buffer, exportAddress string) (*entities.Message, error) {
-	var version, msgLen, setID, setLen uint16
+	var length, version, setID, setLen uint16
 	var exportTime, sequencNum, obsDomainID uint32
-	err := util.Decode(packetBuffer, binary.BigEndian, &version, &msgLen, &exportTime, &sequencNum, &obsDomainID, &setID, &setLen)
-	if err != nil {
+	if err := util.Decode(packetBuffer, binary.BigEndian, &version, &length, &exportTime, &sequencNum, &obsDomainID, &setID, &setLen); err != nil {
 		return nil, err
 	}
 	if version != uint16(10) {
@@ -172,7 +172,7 @@ func (cp *CollectingProcess) decodePacket(packetBuffer *bytes.Buffer, exportAddr
 
 	message := entities.NewMessage(true)
 	message.SetVersion(version)
-	message.SetMessageLen(msgLen)
+	message.SetMessageLen(length)
 	message.SetExportTime(exportTime)
 	message.SetSequenceNum(sequencNum)
 	message.SetObsDomainID(obsDomainID)
@@ -185,6 +185,7 @@ func (cp *CollectingProcess) decodePacket(packetBuffer *bytes.Buffer, exportAddr
 	message.SetExportAddress(exportAddress)
 
 	var set entities.Set
+	var err error
 	if setID == entities.TemplateSetID {
 		set, err = cp.decodeTemplateSet(packetBuffer, obsDomainID)
 		if err != nil {
@@ -356,11 +357,13 @@ func (cp *CollectingProcess) updateAddress(address net.Addr) {
 }
 
 // getMessageLength returns buffer length by decoding the header
-func getMessageLength(msgBuffer *bytes.Buffer) (int, error) {
-	var version, msgLen uint16
-	// We do not really need to decode whole header. Support an utility function
-	// that decodes header based on the offset.
-	err := util.Decode(msgBuffer, binary.BigEndian, &version, &msgLen)
+func getMessageLength(reader *bufio.Reader) (int, error) {
+	partialHeader, err := reader.Peek(4)
+	if err != nil {
+		return 0, err
+	}
+	var msgLen uint16
+	err = util.Decode(bytes.NewBuffer(partialHeader[2:]), binary.BigEndian, &msgLen)
 	if err != nil {
 		return 0, fmt.Errorf("cannot decode message: %v", err)
 	}
--- a/pkg/collector/tcp.go
+++ b/pkg/collector/tcp.go
@@ -1,6 +1,7 @@
 package collector
 
 import (
+	"bufio"
 	"bytes"
 	"crypto/tls"
 	"crypto/x509"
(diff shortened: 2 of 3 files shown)


pkg/collector/tcp.go
```diff
@@ -79,31 +82,39 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
...
+					copy(remainingBytes, copiedBytes)
+					copiedBytes = nil
+					break
 				}
 				size = size - length
-				// get the message here
-				message, err := cp.decodePacket(bytes.NewBuffer(buffBytes[0:length]), address)
+				// Truncate the bytesBuffer here to decode the packet. 4B are already
+				// read as part of getMessageLength.
+				bytesBuffer.Truncate(length - 4)
```
antoninbas, line 108: `Truncate` removes data "at the end" of the buffer, not at the front

Since you already should have consumed these 4 bytes in `getMessageLength`, I am not sure why this line is needed at all.

pkg/collector/tcp.go
```diff
@@ -79,31 +82,39 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
 				return
 			}
 			klog.V(2).Infof("Receiving %d bytes from %s", size, address)
-			buffBytes := make([]byte, size)
-			copy(buffBytes, buff[:size])
+			klog.V(2).Infof("Remaining bytes from last read length: %v", len(remainingBytes))
+			size = size + len(remainingBytes)
+			copiedBytes := make([]byte, size)
+			copy(copiedBytes[:len(remainingBytes)], remainingBytes)
+			copy(copiedBytes[len(remainingBytes):], buff[:size-len(remainingBytes)])
```
antoninbas, line 89: this whole function is too complicated in my mind and with too many data copies. I think you should use a `bufio.Reader` object for performance and to abstract away the socket.

```go
buff := make([]byte, cp.maxBufferSize)
r := bufio.NewReader(conn)
for {
    length, version, err := getMessageLength(r)
    // ...

    // read the full message
    _, err := io.ReadFull(r, buff[:length])
    // ...

    // at this stage you have one IPFIX message stored at the beginning of `buff`
}
```

pkg/collector/process.go
```diff
@@ -351,15 +360,15 @@ func (cp *CollectingProcess) updateAddress(address net.Addr) {
 }
 
 // getMessageLength returns buffer length by decoding the header
-func getMessageLength(msgBuffer *bytes.Buffer) (int, error) {
+func getMessageLength(msgBuffer *bytes.Buffer) (int, uint16, error) {
```
zyiou, line 363: We can name the function `getMsgLenAndVersion` to be more clear

zyiou, line 363: we can change the name to `getMsgLenAndVersion` to be more clear

pkg/collector/tcp.go
```diff
@@ -79,31 +82,39 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
...
 			for size > 0 {
-				length, err := getMessageLength(bytes.NewBuffer(buffBytes))
+				bytesBuffer := bytes.NewBuffer(copiedBytes)
```
zyiou, line 91: Would it be better to have one `bytesBuffer` instead of `copiedBytes` before this for loop and keep consuming it as we always use `Decode` for consuming buffer?

pkg/collector/tcp.go
srikartati: Thanks @antoninbas for the suggestion. It simplified the implementation logic quite a bit.
However, I feel we are reading bytes of the packet twice. First into the buffer from `bufio.Reader` and then next from the `buffer.Bytes` to decode the packet. I feel there is room for improvement there. Any thoughts?

pkg/collector/tcp.go
srikartati: Modified the logic.

pkg/collector/process.go
srikartati: Changed the function to get only the length.

srikartati (commented):
Ran the perf test we have in pkg/test:
Before the change:
```
BenchmarkMultipleExportersToCollector-16    	       1	1362995650 ns/op	354837000 B/op	 5775736 allocs/op
PASS
ok  	github.com/vmware/go-ipfix/pkg/test	2.126s
```
After the change:
```
BenchmarkMultipleExportersToCollector-16    	       1	1375070617 ns/op	349695224 B/op	 5715157 allocs/op
PASS
ok  	github.com/vmware/go-ipfix/pkg/test	1.814s
```

pkg/collector/tcp.go
```diff
@@ -63,48 +64,36 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
...
+			// process.
+			copiedBytes := make([]byte, length)
+			copy(copiedBytes, buff[:length])
+			message, err := cp.decodePacket(bytes.NewBuffer(copiedBytes), address)
+			if err != nil {
+				klog.Error(err)
+				continue
 			}
+			klog.V(4).Infof("Processed message from exporter %v, number of records: %v, observation domain ID: %v",
+				message.GetExportAddress(), message.GetSet().GetNumberOfRecords(), message.GetObsDomainID())
```
zyiou, line 96: not in this PR, I think we should replace `message.GetExportAddress()` with other useful information since export address will be Pod host IP.

pkg/collector/tcp.go
```diff
@@ -63,48 +64,36 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
...
 		for {
-			size, err := conn.Read(buff)
+			length, err := getMessageLength(reader)
 			if err != nil {
-				if err == io.EOF {
-					klog.Infof("Connection from %s has been closed.", address)
-				} else {
-					klog.Errorf("Error in collecting process: %v", err)
-				}
+				klog.Errorf("error when retreiving message length: %v", err)
```
antoninbas, line 76:
```suggestion
				klog.Errorf("error when retrieving message length: %v", err)
```

pkg/collector/tcp.go
```diff
@@ -63,48 +64,36 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
...
-				message, err := cp.decodePacket(bytes.NewBuffer(buffBytes[0:length]), address)
-				if err != nil {
-					klog.Error(err)
-					buffBytes = nil
-					break
-				}
-				klog.V(4).Infof("Processed message from exporter %v, number of records: %v, observation domain ID: %v",
-					message.GetExportAddress(), message.GetSet().GetNumberOfRecords(), message.GetObsDomainID())
-				buffBytes = buffBytes[length:]
+			_, err = io.ReadFull(reader, buff[:length])
```
antoninbas, line 80: you don't need the full-sized `buff` at all, and the copy is not needed.
this should read:
```go
buff := make([]byte, length)
_, err = io.ReadFull(reader, buff)
```

at that stage, and based on the rest of the code, I don't think you will have any superfluous copies. It is a bit of  a shame that we still have to do a dynamic buffer allocation for each packet, but there is no _easy_ way around that. The next logical step would be to use a large ring buffer and avoid these dynamic allocations.

pkg/collector/process_test.go
```diff
@@ -142,6 +142,46 @@ func TestTCPCollectingProcess_ReceiveDataRecord(t *testing.T) {
...
+	go func() {
+		conn, err = net.Dial(collectorAddr.Network(), collectorAddr.String())
+		if err != nil {
+			t.Errorf("Cannot establish connection to %s", collectorAddr.String())
+		}
+		conn.Write(validDataPacket)
+		conn.Write(validDataPacket)
+	}()
+	<-cp.GetMsgChan()
+	<-cp.GetMsgChan()
```
zyiou, line 172: Like this unit test: if we are able to decode the packet without any error, we should be able to receive them correctly.

zyiou (dismissed):
LGTM

antoninbas:
> However, I feel we are reading bytes of the packet twice. First into the buffer from bufio.Reader and then next from the buffer.Bytes to decode the packet. I feel there is room for improvement there. Any thoughts?

You do have an unnecessary copy as I pointed out in the review, but it's unrelated to this. If you fix that unneeded copy:
* we copy the data into the buffer using `bufio.Reader`
* the buffer is wrapped into a `bytes.Buffer` object, which does not do any copy or allocation IIRC
* calls to `Next()` on the `bytes.Buffer` object do not copy anything, they return a slice using the same underlying storage, which is why the slice becomes invalid if there is a `Read` / `Write` operation

pkg/collector/process_test.go
srikartati: This test is not applicable anymore. Removed it.

pkg/collector/tcp.go
srikartati: done

pkg/collector/tcp.go
srikartati: Agreed.. we can avoid the copy. This means that there is no buffer required with `MaxBufferSize` which is specified by the user. I have seen some improvements, especially with CPU/runtime because of no copy.

Thanks for the insight on dynamic buffer allocation. We will look into ring bugger approach in a follow up PR.

pkg/collector/tcp.go
srikartati: Modified it with set type. Agree that we have to improve debugging in the intermediate process.

srikartati:
> * calls to `Next()` on the `bytes.Buffer` object do not copy anything, they return a slice using the same underlying storage, which is why the slice becomes invalid if there is a `Read` / `Write` operation

We do `Next()` when decoding the field elements, but for decoding the header, we still use `util.Decode` which uses `binary.Read` which does `io.ReadAll` again. However, the header is smaller when compared to the field elements.

pkg/collector/tcp.go
```diff
@@ -63,48 +64,31 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
...
-					buffBytes = nil
-					break
-				}
-				klog.V(4).Infof("Processed message from exporter %v, number of records: %v, observation domain ID: %v",
-					message.GetExportAddress(), message.GetSet().GetNumberOfRecords(), message.GetObsDomainID())
-				buffBytes = buffBytes[length:]
+			buff := make([]byte, length)
+			_, err = io.ReadFull(reader, buff)
+			if err != nil {
+				klog.Errorf("error when reading the message: %v", err)
```
antoninbas, line 81: I don't know if you capitalize all logs by convention, like we do in Antrea

antoninbas (approved):

pkg/collector/tcp.go
srikartati: We do not have that convention but going forward we will look at that.

pkg/test/collector_perf_test.go
```diff
@@ -54,14 +54,15 @@ const (
...
+// TODO: This should not be benchmark test if it is not run multiple times. Test should use "b.N".
 func BenchmarkMultipleExportersToCollector(b *testing.B) {
 	disableLogToStderr()
 	ctrl := gomock.NewController(b)
 	defer ctrl.Finish()
 	collectorInput := collector.CollectorInput{
 		Address:       "127.0.0.1:0",
 		Protocol:      "tcp",
-		MaxBufferSize: 65535,
+		MaxBufferSize: 10000,
```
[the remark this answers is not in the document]
srikartati: I wanted to check if it is used but skipped my mind when I last pushed the commit. Removing it.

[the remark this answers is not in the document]
srikartati: Let me take back that. UDP is still using it. I will keep it as it is for UDP. Will fix that later in a different PR.

PR #241, opened 2021-08-11
Remove UDP message size check

srikartati:
We decided to rely on IP fragmentation in the case of UDP transport.
This is going against the IPFIX RFC suggestion. We will revisit if
there are any issues with UDP transport in the future.

### DIFF
--- a/pkg/exporter/process.go
+++ b/pkg/exporter/process.go
@@ -45,13 +45,15 @@ type templateValue struct {
 // 2. Only one observation point per observation domain is supported,
 //    so observation point ID not defined.
 // 3. Supports only TCP and UDP; one session at a time. SCTP is not supported.
-// TODO:UDP needs to send MTU size packets as per RFC7011
+// 4. UDP needs to send MTU size packets as per RFC7011. We are not honoring that,
+//    and relying on IP fragmentation and assuming data loss in the network is minimal.
+//    We will revisit this if there are any issues, and get PathMTU from the user
+//    as part of exporter input.
 type ExportingProcess struct {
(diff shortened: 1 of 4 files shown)


srikartati:
@zyiou Ping for review.

pkg/exporter/process.go
```diff
@@ -280,13 +273,9 @@ func (ep *ExportingProcess) createAndSendIPFIXMsg(set entities.Set) (int, error)
...
 	msgLen := entities.MsgHeaderLength + set.GetSetLength()
 	if ep.connToCollector.LocalAddr().Network() == "tcp" {
-		if msgLen > entities.MaxTcpSocketMsgSize {
+		if msgLen > entities.MaxSocketMsgSize {
 			return 0, fmt.Errorf("TCP transport: message size exceeds max socket buffer size")
 		}
-	} else {
-		if msgLen > ep.pathMTU {
-			return 0, fmt.Errorf("UDP transport: message size exceeds max pathMTU (set as %v)", ep.pathMTU)
-		}
```
zyiou, line 289: I feel we should at least have some limit checking on UDP message size (even though it may not happen, we can have 65535 limit check). I cannot remember why we are setting 1500 as MaxUDPMsgSize before, is it because 1500 is the general public internet MTU?

srikartati: Yes, it makes sense to extend the max socket buffer size check to both tcp and udp.
1500 is the default path MTU we set, if a user provides no path MTU in the input for the exporter process.

zyiou (approved):
LGTM

PR #242, opened 2021-08-11
Skip sending template record for JSON records

zyiou:


pkg/exporter/process.go
```diff
@@ -225,7 +225,9 @@ func (ep *ExportingProcess) SendSet(set entities.Set) (int, error) {
 	if !ep.sendJSONRecord {
 		bytesSent, err = ep.createAndSendIPFIXMsg(set)
 	} else {
-		bytesSent, err = ep.createAndSendJSONMsg(set)
+		if setType == entities.Data {
```
srikartati, line 228: Looks like we do the check on the user side too right?
https://github.com/antrea-io/antrea/pull/2559/files#diff-6494662cfeb5d7510bb66484de54275b401bd33ffacacb6c857d4c3a3354c29aR360

zyiou: I want to remove the user side check otherwise no templates will be stored in exporting process when sending JSON record.

srikartati: Ok. I think templates are required to build the ipfix packet before encoding to json right?

zyiou: Not required when building but necessary for doing sanity check before sending the record.

srikartati: Ok we need to revisit the utility of the sanity check.

srikartati (approved):

PR #243, opened 2021-08-11
Add new method for adding record with extra elements

srikartati:
This is to keep the old method as it is and introduce new
one for better performance when adding data records, where additional elements are added in the future.


zyiou (approved):
LGTM

PR #248, opened 2021-08-23
Remove conn mutex for exporting process

antoninbas:
Golang Conn objects are "thread-safe" and multiple goroutines may invoke methods
on a Conn simultaneously. The Golang standard library will take care of
serializing the call correctly. So we can remove this extra mutex.


antoninbas:
@srikartati @zyiou for review (I don't have write access to this repo so cannot assign reviewers directly)

zyiou (approved):
Thanks for taking care of this issue. LGTM

srikartati (approved):
Thanks for fixing this. LGTM.

srikartati:
Merging this as there are no comments to be resolved.

Issue #250, opened 2021-08-26
Inaccurate delta stats of aggregated records


zyiou:
**Is your feature request related to a problem? Please describe.**
Currently when aggregating records, we added delta stats https://github.com/vmware/go-ipfix/blob/main/pkg/intermediate/aggregate.go#L522-L528. Additionally, throughput calculation will make more sense when using accurate `octetDeltaCount` instead of source or destination specific delta stats.

**Describe the solution you'd like**
Assign maximum of the delta stats instead of summing up.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #251, opened 2021-08-26
Improve octetDeltaCount calculation

heanlan:
Currently when we calculate `octetDeltaCount` for aggregated records, we are summing up the values from all incoming records. This approach leads to an issue that, in the case of inter-node flows, we will receive records from both source and destination node for the same flow, which makes `octetDeltaCount` approximate to 2*real value. 

To make it more accurate, we change the update logic to: 
`new_octetDeltaCount = max(new_octetDeltaCountFromSource, old_octetDeltaCount)`, if the incoming record is from source node;
`new_octetDeltaCount = max(new_octetDeltaCountFromDestination, old_octetDeltaCount)`, if the incoming record is from destination node

fixes: #250

### DIFF
--- a/pkg/intermediate/aggregate.go
+++ b/pkg/intermediate/aggregate.go
@@ -472,6 +472,24 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
 			existingIeWithValue, index, _ := existingRecord.GetInfoElementWithValue(element)
 			switch ieWithValue.Element.Name {
 			case "flowEndSeconds":
+				srcPodName, _, _ := existingRecord.GetInfoElementWithValue("sourcePodName")
+				dstPodName, _, _ := existingRecord.GetInfoElementWithValue("destinationPodName")
+				octetDeltaCount, _, _ := incomingRecord.GetInfoElementWithValue("octetDeltaCount")
+				octetTotalCount, _, _ := incomingRecord.GetInfoElementWithValue("octetTotalCount")
+				if srcPodName.Value.(string) == "perftest-a" {
+					klog.Info(srcPodName.Value.(string))
+					klog.Info("FLOWENDSECONDS: ", ieWithValue.Value)
+					klog.Info("OCTETDELTACOUNT: ", octetDeltaCount.Value.(uint64))
+					klog.Info("OCTETTOTALCOUNT: ", octetTotalCount.Value.(uint64))
+					klog.Info("\n\n")
+				}
+				if dstPodName.Value.(string) == "perftest-c" {
+					klog.Info(dstPodName.Value.(string))
+					klog.Info("FLOWENDSECONDS: ", ieWithValue.Value)
+					klog.Info("OCTETDELTACOUNT: ", octetDeltaCount.Value.(uint64))
+					klog.Info("OCTETTOTALCOUNT: ", octetTotalCount.Value.(uint64))
+					klog.Info("\n\n")
+				}
 				// Update flow end timestamp if it is latest.
 				if isLatest {
 					existingIeWithValue.Value = ieWithValue.Value


pkg/intermediate/aggregate.go
```diff
@@ -515,45 +516,43 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
...
+				srcExistingIeWithValue, index, _ := existingRecord.GetInfoElementWithValue(antreaSourceStatsElements[i])
 				if !isDelta {
-					existingIeWithValue.Value = ieWithValue.Value
+					srcExistingIeWithValue.Value = ieWithValue.Value
 				} else {
-					existingIeWithValue.Value = existingIeWithValue.Value.(uint64) + ieWithValue.Value.(uint64)
+					srcExistingIeWithValue.Value = srcExistingIeWithValue.Value.(uint64) + ieWithValue.Value.(uint64)
+					// delta count takes the maximum of the current delta count and
+					// the updated delta count from source node
+					existingIeWithValue.Value = util.MaxUint64(existingIeWithValue.Value.(uint64), srcExistingIeWithValue.Value.(uint64))
```
srikartati, line 530: how about filling up src and dst stat elements first and then fill the common stat elements? I think that will improve the readability of the code and we dont have to update the common stat elements at two different places.

heanlan: make sense to me. Thanks Srikar!

srikartati (approved):
LGTM

pkg/intermediate/aggregate.go
```diff
@@ -506,54 +507,53 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
...
 				if !isDelta {
-					existingIeWithValue.Value = ieWithValue.Value
+					srcExistingIeWithValue.Value = ieWithValue.Value
 				} else {
-					existingIeWithValue.Value = existingIeWithValue.Value.(uint64) + ieWithValue.Value.(uint64)
+					srcExistingIeWithValue.Value = srcExistingIeWithValue.Value.(uint64) + ieWithValue.Value.(uint64)
+					srcDeltaVal = srcExistingIeWithValue.Value.(uint64)
 				}
-				if err := existingRecord.SetInfoElementWithValue(index, *existingIeWithValue); err != nil {
+				klog.Info(srcExistingIeWithValue.Element.Name, srcExistingIeWithValue.Value)
```
zyiou, line 523: Why wee need a log here?

pkg/intermediate/aggregate.go
```diff
@@ -506,54 +507,53 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
...
 				if !isDelta {
-					existingIeWithValue.Value = ieWithValue.Value
+					dstExistingIeWithValue.Value = ieWithValue.Value
 				} else {
-					existingIeWithValue.Value = existingIeWithValue.Value.(uint64) + ieWithValue.Value.(uint64)
+					dstExistingIeWithValue.Value = dstExistingIeWithValue.Value.(uint64) + ieWithValue.Value.(uint64)
+					dstDeltaVal = dstExistingIeWithValue.Value.(uint64)
 				}
-				if err := existingRecord.SetInfoElementWithValue(index, *existingIeWithValue); err != nil {
+				klog.Info(dstExistingIeWithValue.Element.Name, dstExistingIeWithValue.Value)
```
zyiou, line 537: same here

heanlan: forgot to remove... thanks for catching that

zyiou (dismissed):
LGTM. How does it work in Antrea?

heanlan:
> LGTM. How does it work in Antrea?

It works well in Antrea, referring to https://github.com/antrea-io/antrea/pull/2692
although I think the accuracy is approximately equally good as the previous approach.

PR #254, opened 2021-09-10
Remove duplicated code in ie_value.go

srikartati:
There is lot of repeated code. Base structure
can help in removing that.

### DIFF
--- a/pkg/entities/ie_value.go
+++ b/pkg/entities/ie_value.go
@@ -1,7 +1,6 @@
 package entities
 
 import (
-	"fmt"
 	"net"
 )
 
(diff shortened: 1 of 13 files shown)


antoninbas (dismissed):

pkg/entities/ie_value.go
```diff
@@ -43,2317 +42,608 @@ type InfoElementWithValue interface {
...
-func NewDateTimeSecondsInfoElement(element *InfoElement, val uint32) *DateTimeSecondsInfoElement {
-	return &DateTimeSecondsInfoElement{
-		element: element,
-		value:   val,
+func NewUnsigned8InfoElement(element *InfoElement, val uint8) *Unsigned8InfoElement {
+	infoElem := &Unsigned8InfoElement{
+		value: val,
 	}
+	infoElem.element = element
+	return infoElem
```
zyiou, line 183: How about writing it in this way:
```
return &Unsigned8InfoElement{
	value: val,
        element: element,
}
```

pkg/entities/ie_value.go
```diff
@@ -43,2317 +42,608 @@ type InfoElementWithValue interface {
...
+		value: val,
+	}
+	infoElem.element = element
+	return infoElem
 }
 
-func (dsec *DateTimeSecondsInfoElement) GetFloat64Value() (float64, error) {
-	return 0, fmt.Errorf("accessing value of wrong data type")
+func (u16 *Unsigned16InfoElement) GetUnsigned16Value() (uint16, error) {
+	return u16.value, nil
```
zyiou, line 217: Do we still need to return error here?

pkg/entities/ie_value.go
```diff
@@ -43,2317 +42,608 @@ type InfoElementWithValue interface {
...
 }
 
-func (dsec *DateTimeSecondsInfoElement) GetBooleanValue() (bool, error) {
-	return false, fmt.Errorf("accessing value of wrong data type")
+func (u16 *Unsigned16InfoElement) SetUnsigned16Value(val uint16) error {
+	u16.value = val
+	return nil
```
zyiou, line 222: Same here.

pkg/entities/ie_value.go
srikartati: Embedded element cannot be accessed like that.

pkg/entities/ie_value.go
srikartati: Good catch. Not needed. The rest of the code that uses these get and set functions becomes simpler.

pkg/entities/ie_value.go
```diff
@@ -10,2365 +9,636 @@ type InfoElementWithValue interface {
...
+	SetStringValue(val string)
+	SetIPAddressValue(val net.IP)
 	IsValueEmpty() bool
 	GetLength() int
 	ResetValue()
 }
 
-type Unsigned8InfoElement struct {
-	value   uint8
+type BaseInfoElement struct {
```
zyiou, line 45: Is it necessary to make this type public? I didn't see it is used outside this package.

zyiou (dismissed):
Overall LGTM. only one nit.

pkg/entities/ie_value.go
srikartati: Not needed. Modified.

zyiou (approved):
LGTM

PR #261, opened 2021-10-01
Add GetRecords method to AggregationProcess

yanjunz97:
This commit adds GetRecords method to get string format flow records.
It accepts FlowKey to filter the flow records.

### DIFF
--- a/pkg/intermediate/aggregate.go
+++ b/pkg/intermediate/aggregate.go
@@ -215,6 +215,38 @@ func (a *AggregationProcess) GetExpiryFromExpirePriorityQueue() time.Duration {
 	return a.inactiveExpiryTimeout
 }
 
+// GetRecords returns the string format flow records by FlowKey from flowKeyRecordMap.
+// Returns partial match if FlowKey is not complete.
+// Returns all the flow records if FlowKey is not provided.
+func (a *AggregationProcess) GetRecords(flowKey *FlowKey) []string {
+	a.mutex.Lock()
+	defer a.mutex.Unlock()
+
+	var records []string
+	// Complete filter
+	if flowKey != nil && flowKey.SourceAddress != "" && flowKey.DestinationAddress != "" &&
+		flowKey.Protocol != 0 && flowKey.SourcePort != 0 && flowKey.DestinationPort != 0 {
+		if record, ok := a.flowKeyRecordMap[*flowKey]; ok {
+			records = append(records, record.Record.GetString())
+		}
+		return records
+	}
+	// Partial filter
+	for currentFlowKey, record := range a.flowKeyRecordMap {
+		if flowKey != nil {
+			if (flowKey.SourceAddress != "" && flowKey.SourceAddress != currentFlowKey.SourceAddress) ||
+				(flowKey.DestinationAddress != "" && flowKey.DestinationAddress != currentFlowKey.DestinationAddress) ||
+				(flowKey.Protocol != 0 && flowKey.Protocol != currentFlowKey.Protocol) ||
+				(flowKey.SourcePort != 0 && flowKey.SourcePort != currentFlowKey.SourcePort) ||
+				(flowKey.DestinationPort != 0 && flowKey.DestinationPort != currentFlowKey.DestinationPort) {
+				continue
+			}
+		}
+		records = append(records, record.Record.GetString())
+	}
+	return records
+}
+
 func (a *AggregationProcess) ForAllExpiredFlowRecordsDo(callback FlowKeyRecordMapCallBack) error {
 	a.mutex.Lock()
 	defer a.mutex.Unlock()
--- a/pkg/intermediate/aggregate_test.go
+++ b/pkg/intermediate/aggregate_test.go
@@ -694,6 +694,103 @@ func TestGetExpiryFromExpirePriorityQueue(t *testing.T) {
 	}
 }
 
+func assertRecordString(t *testing.T, recordString string, ipv6 bool) {
+	if ipv6 {
+		assert.Contains(t, recordString, "    sourceIPv6Address: 2001:0:3238:dfe1:63::fefb \n")
+		assert.Contains(t, recordString, "    destinationIPv6Address: 2001:0:3238:dfe1:63::fefc \n")
+		assert.Contains(t, recordString, "    destinationClusterIPv6: 2001:0:3238:bbbb:63::aaaa \n")
+	} else {
+		assert.Contains(t, recordString, "    sourceIPv4Address: 10.0.0.1 \n")
+		assert.Contains(t, recordString, "    destinationIPv4Address: 10.0.0.2 \n")
+		assert.Contains(t, recordString, "    destinationClusterIPv4: 192.168.0.1 \n")
+	}
+	assert.Contains(t, recordString, "    sourceTransportPort: 1234 \n")
+	assert.Contains(t, recordString, "    destinationTransportPort: 5678 \n")
+	assert.Contains(t, recordString, "    protocolIdentifier: 6 \n")
+	assert.Contains(t, recordString, "    sourcePodName: pod1 \n")
+	assert.Contains(t, recordString, "    destinationPodName: pod2 \n")
+	assert.Contains(t, recordString, "    destinationServicePort: 4739 \n")
+	assert.Contains(t, recordString, "    flowEndSeconds: 1 \n")
+	assert.Contains(t, recordString, "    flowType: 2 \n")
+	assert.Contains(t, recordString, "    flowEndReason: 2 \n")
+	assert.Contains(t, recordString, "    tcpState: ESTABLISHED \n")
+	assert.Contains(t, recordString, "    ingressNetworkPolicyRuleAction: 0 \n")
+	assert.Contains(t, recordString, "    egressNetworkPolicyRuleAction: 0 \n")
+	assert.Contains(t, recordString, "    ingressNetworkPolicyRulePriority: 50000 \n")
+	assert.Contains(t, recordString, "    packetTotalCount: 500 \n")
+	assert.Contains(t, recordString, "    packetDeltaCount: 0 \n")
+	assert.Contains(t, recordString, "    reversePacketTotalCount: 500 \n")
+	assert.Contains(t, recordString, "    reversePacketDeltaCount: 0 \n")
+}
+
+func TestGetRecords(t *testing.T) {
+	messageChan := make(chan *entities.Message)
+	input := AggregationInput{
+		MessageChan:           messageChan,
+		WorkerNum:             2,
+		CorrelateFields:       fields,
+		ActiveExpiryTimeout:   testActiveExpiry,
+		InactiveExpiryTimeout: testInactiveExpiry,
+	}
+	ap, _ := InitAggregationProcess(input)
+
+	// Add records with IPv4 fields.
+	recordIPv4Src := createDataMsgForSrc(t, false, false, false, false, false).GetSet().GetRecords()[0]
+	recordIPv4Dst := createDataMsgForDst(t, false, false, false, false, false).GetSet().GetRecords()[0]
+	// Add records with IPv6 fields.
+	recordIPv6Src := createDataMsgForSrc(t, true, false, false, false, false).GetSet().GetRecords()[0]
+	recordIPv6Dst := createDataMsgForDst(t, true, false, false, false, false).GetSet().GetRecords()[0]
+	records := []entities.Record{recordIPv4Src, recordIPv4Dst, recordIPv6Src, recordIPv6Dst}
+	for _, record := range records {
+		flowKey, isIPv4, _ := getFlowKeyFromRecord(record)
+		err := ap.addOrUpdateRecordInMap(flowKey, record, isIPv4)
+		assert.NoError(t, err)
+	}
+
+	flowKeyIPv4, _, _ := getFlowKeyFromRecord(recordIPv4Src)
+	partialFlowKeyIPv6 := &FlowKey{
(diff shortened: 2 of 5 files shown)


pkg/intermediate/aggregate.go
```diff
@@ -215,6 +215,71 @@ func (a *AggregationProcess) GetExpiryFromExpirePriorityQueue() time.Duration {
 	return a.inactiveExpiryTimeout
 }
 
+// GetRecordStrings returns the flow records by FlowKey from flowKeyRecordMap.
+// Returns all the flow records if FlowKey is not provided.
+func (a *AggregationProcess) GetRecordStrings(flowKey *FlowKey) []string {
```
srikartati, line 220: I think GetRecords is sufficient. In the function description, you can add in string format.

pkg/intermediate/aggregate.go
```diff
@@ -215,6 +215,71 @@ func (a *AggregationProcess) GetExpiryFromExpirePriorityQueue() time.Duration {
...
+	defer a.mutex.Unlock()
+
+	var records []string
+	for currentFlowKey, record := range a.flowKeyRecordMap {
+		if flowKey != nil {
+			if (flowKey.SourceAddress != "" && flowKey.SourceAddress != currentFlowKey.SourceAddress) ||
+				(flowKey.DestinationAddress != "" && flowKey.DestinationAddress != currentFlowKey.DestinationAddress) ||
+				(flowKey.Protocol != 0 && flowKey.Protocol != currentFlowKey.Protocol) ||
+				(flowKey.SourcePort != 0 && flowKey.SourcePort != currentFlowKey.SourcePort) ||
+				(flowKey.DestinationPort != 0 && flowKey.DestinationPort != currentFlowKey.DestinationPort) {
```
srikartati, line 231: Could we do a hash check before checking individual elements to improve the performance?

pkg/intermediate/aggregate.go
```diff
@@ -215,6 +215,71 @@ func (a *AggregationProcess) GetExpiryFromExpirePriorityQueue() time.Duration {
...
+			case entities.Ipv4Address, entities.Ipv6Address:
+				recordString += fmt.Sprintf("    %s: %v \n", ie.GetName(), ie.GetIPAddressValue())
+			case entities.String:
+				recordString += fmt.Sprintf("    %s: %v \n", ie.GetName(), ie.GetStringValue())
+			default:
+				err := fmt.Errorf("API supports only valid information elements with datatypes given in RFC7011")
+				recordString += fmt.Sprintf("    %s: %v \n", ie.GetName(), err)
+			}
+		}
+		records = append(records, recordString)
```
srikartati, line 278: As there is only one unique flow key in the map, if the flowKey is not nil, then we can break after this point.

srikartati (commented):
Please add unit tests for the function.

pkg/intermediate/aggregate.go
yanjunz97: Would add a hash check before the loop. Here I would like to add some flexibility in the search. For example, if the user only provides the source address, all the flow records with that source address would be returned. In this case, it seems we cannot add the break here.

yanjunz97:
> Please add unit tests for the function.

I am thinking about the unit tests. As the function returns strings, should I hard code the output in the test? As the flow records might be long, I am looking for some more elegant way but does not have a good idea. Is there any suggestion?

pkg/intermediate/aggregate.go
srikartati: Ok. If that is the case, we can break it into two scenarios: complete filter and partial filter.
In the case of complete filter, you can directly check in the map. And with partial filter, you can use your current approach.

In the future, we can explore moving to `cache.Indexer` from the [kubernetes client-go package](https://github.com/kubernetes/client-go/blob/v0.22.2/tools/cache/index.go#L35). This makes partial filter search efficient.

srikartati:
> > Please add unit tests for the function.
> 
> I am thinking about the unit tests. As the function returns strings, should I hard code the output in the test? As the flow records might be long, I am looking for some more elegant way but does not have a good idea. Is there any suggestion?

Hardcoding the output sounds fine. You can have a smaller record with fewer elements to make things easy. Multiple of them covering all fields.

pkg/intermediate/aggregate.go
```diff
@@ -215,6 +215,71 @@ func (a *AggregationProcess) GetExpiryFromExpirePriorityQueue() time.Duration {
 	return a.inactiveExpiryTimeout
 }
 
+// GetRecordStrings returns the flow records by FlowKey from flowKeyRecordMap.
+// Returns all the flow records if FlowKey is not provided.
+func (a *AggregationProcess) GetRecordStrings(flowKey *FlowKey) []string {
+	a.mutex.Lock()
+	defer a.mutex.Unlock()
+
+	var records []string
```
zyiou, line 224: We can first have a size check of records, then initialize the string slice with fixed size and assign record by its index instead of using append. In that way we can save reduce memory allocations.

pkg/intermediate/aggregate.go
```diff
@@ -215,6 +215,71 @@ func (a *AggregationProcess) GetExpiryFromExpirePriorityQueue() time.Duration {
...
+			case entities.MacAddress:
+				recordString += fmt.Sprintf("    %s: %v \n", ie.GetName(), ie.GetMacAddressValue())
+			case entities.Ipv4Address, entities.Ipv6Address:
+				recordString += fmt.Sprintf("    %s: %v \n", ie.GetName(), ie.GetIPAddressValue())
+			case entities.String:
+				recordString += fmt.Sprintf("    %s: %v \n", ie.GetName(), ie.GetStringValue())
+			default:
+				err := fmt.Errorf("API supports only valid information elements with datatypes given in RFC7011")
+				recordString += fmt.Sprintf("    %s: %v \n", ie.GetName(), err)
+			}
```
zyiou, line 276: Maybe we can put this part into a separate function inside entities package (something like `record.GetString()` or `record.String()`). It will be useful for other loggings in collecting process or exporting process.

pkg/intermediate/aggregate.go
yanjunz97: The size of the records may depend on the flow key provided. Like if the user only provides a source address in the flow key, all the records with that source address will be return. Should I allocate the same size as the flowKeyRecordMap, which is an upper bound. Or should I iterate the map twice, once for the size and the other to store the string? Or is there any other way?

pkg/intermediate/aggregate.go
dreamtalen: I agree, our ipfix-collector (https://github.com/vmware/go-ipfix/blob/36dc9443b23c980188bd96970c28fdd778b45caf/cmd/collector/collector.go#L88) could utilize this when printing out record

pkg/intermediate/aggregate.go
zyiou: I see your point. So the function does not actually return the flow records that have exact match with flow key right? If so, better to have some comment/description clarifying that. When we call the function, we will need to create a new flow key even we only want to search the match of sourceIPAddress?

yanjunz97: Yes to both questions. If we provide a complete flowkey, it would be an exact match, otherwise it is like a partial filter. I will add some descriptions. I am also going to add a hash check first as suggested by [Srika's comments above](https://github.com/vmware/go-ipfix/pull/261#discussion_r720428027).

pkg/intermediate/aggregate.go
yanjunz97: Updated

pkg/intermediate/aggregate.go
yanjunz97: Updated

pkg/intermediate/aggregate.go
yanjunz97: Updated

pkg/intermediate/aggregate.go
yanjunz97: Updated description

pkg/intermediate/aggregate.go
yanjunz97: Updated

pkg/intermediate/aggregate_test.go
```diff
@@ -694,6 +694,66 @@ func TestGetExpiryFromExpirePriorityQueue(t *testing.T) {
...
+
+	partialFlowKeyIPv6 := &FlowKey{
+		SourceAddress: "2001:0:3238:dfe1:63::fefb",
+	}
+	recordStrings = ap.GetRecords(partialFlowKeyIPv6)
+	assert.Equalf(t, 1, len(recordStrings), "Number of IPv6 records string is incorrect, expected 1 got %d", len(recordStrings))
+	assert.Contains(t, recordStrings[0], "    sourceIPv6Address: 2001:0:3238:dfe1:63::fefb \n")
+	assert.Contains(t, recordStrings[0], "    destinationIPv6Address: 2001:0:3238:dfe1:63::fefc \n")
+	assert.Contains(t, recordStrings[0], "    destinationClusterIPv6: 2001:0:3238:bbbb:63::aaaa \n")
+}
```
dreamtalen, line 755: Could we add a test case here: if no flow key provided return all flow records?

pkg/intermediate/aggregate_test.go
```diff
@@ -694,6 +694,66 @@ func TestGetExpiryFromExpirePriorityQueue(t *testing.T) {
...
+	assert.Contains(t, recordStrings[0], "    ingressNetworkPolicyRulePriority: 50000 \n")
+	assert.Contains(t, recordStrings[0], "    packetTotalCount: 500 \n")
+	assert.Contains(t, recordStrings[0], "    packetDeltaCount: 0 \n")
+	assert.Contains(t, recordStrings[0], "    reversePacketTotalCount: 500 \n")
+	assert.Contains(t, recordStrings[0], "    reversePacketDeltaCount: 0 \n")
+
+	partialFlowKeyIPv6 := &FlowKey{
+		SourceAddress: "2001:0:3238:dfe1:63::fefb",
+	}
+	recordStrings = ap.GetRecords(partialFlowKeyIPv6)
```
dreamtalen, line 750: Understand there are lots of same fields for v4 and v6 records so you omitted some assert checks. 
Could we have a seperate function containing these asserts with ip version as an argument to reuse asserts?

pkg/intermediate/aggregate_test.go
yanjunz97: Sounds good. Added

pkg/intermediate/aggregate_test.go
yanjunz97: Sounds good. Updated

dreamtalen (dismissed):
LGTM

pkg/intermediate/aggregate_test.go
```diff
@@ -713,45 +742,34 @@ func TestGetRecords(t *testing.T) {
...
 		SourceAddress: "2001:0:3238:dfe1:63::fefb",
 	}
 	recordStrings = ap.GetRecords(partialFlowKeyIPv6)
 	assert.Equalf(t, 1, len(recordStrings), "Number of IPv6 records string is incorrect, expected 1 got %d", len(recordStrings))
-	assert.Contains(t, recordStrings[0], "    sourceIPv6Address: 2001:0:3238:dfe1:63::fefb \n")
-	assert.Contains(t, recordStrings[0], "    destinationIPv6Address: 2001:0:3238:dfe1:63::fefc \n")
-	assert.Contains(t, recordStrings[0], "    destinationClusterIPv6: 2001:0:3238:bbbb:63::aaaa \n")
+	assertRecordString(t, recordStrings[0], true)
+
+	recordStrings = ap.GetRecords(nil)
```
srikartati, line 763: How about initializing all three testcases with a struct and use t.Run? https://gobyexample.com/testing
It will be easy to follow all testcases.

yanjunz97: Updated

pkg/intermediate/aggregate_test.go
```diff
@@ -694,6 +694,103 @@ func TestGetExpiryFromExpirePriorityQueue(t *testing.T) {
...
+			"IPv6 flowkey",
+			partialFlowKeyIPv6,
+			1,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			recordStrings := ap.GetRecords(tc.flowKey)
+			assert.Equalf(t, tc.expectedLen, len(recordStrings), "%s: Number of records string is incorrect, expected %d got %d", tc.name, tc.expectedLen, len(recordStrings))
+			if tc.expectedLen == 1 {
```
srikartati, line 779: flowKey != nil is better here?

pkg/intermediate/aggregate_test.go
```diff
@@ -694,6 +694,103 @@ func TestGetExpiryFromExpirePriorityQueue(t *testing.T) {
...
+				assertRecordString(t, recordStrings[0], tc.name == "IPv6 flowkey")
+			} else {
+				if strings.Contains(recordStrings[0], "sourceIPv6Address") {
+					assertRecordString(t, recordStrings[0], true)
+					assertRecordString(t, recordStrings[1], false)
+				} else {
+					assertRecordString(t, recordStrings[0], false)
+					assertRecordString(t, recordStrings[1], true)
+				}
```
srikartati, line 788: The prior else block is only for one test case when no flow key is given, right?
What is the use of this if-else block--don't we expect same records every time?

pkg/intermediate/aggregate_test.go
yanjunz97: Make sense! Updated

srikartati (approved):
LGTM

PR #262, opened 2021-10-05
Modify flow visibility methods and return type

zyiou:
This commit modifies some of the methods used for flow visibility
and their return type to keep consistent with Antrea repository.

### DIFF
--- a/pkg/collector/process.go
+++ b/pkg/collector/process.go
@@ -137,19 +137,19 @@ func (cp *CollectingProcess) CloseMsgChan() {
 	close(cp.messageChan)
 }
 
-func (cp *CollectingProcess) GetNumOfRecordsReceived() uint64 {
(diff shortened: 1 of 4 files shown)


pkg/collector/process.go
```diff
@@ -137,19 +137,19 @@ func (cp *CollectingProcess) CloseMsgChan() {
 	close(cp.messageChan)
 }
 
-func (cp *CollectingProcess) GetNumOfRecordsReceived() uint64 {
+func (cp *CollectingProcess) GetNumRecordsReceived() int64 {
 	cp.mutex.RLock()
 	defer cp.mutex.RUnlock()
-	return cp.numOfRecordsReceived
+	return int64(cp.numOfRecordsReceived)
```
srikartati, line 143: Why do we need to use `int64` here? Is there a possibility for negative records?

srikartati (approved):

Issue #265, opened 2021-10-21
Add connection throughput calculation into aggregation process


heanlan:
**Is your feature request related to a problem? Please describe.**
We want to include the throughput calculation in flow aggregator.

**Describe the solution you'd like**
Add field `throughput` to statsElementList
Add field `throughputFromSource` to antreaSourceStatsElementList
Add field `throughputFromDestination` to antreaDestinationStatsElementList

Let's say a new record arrives from the source node. We update the source node specific fields. We compute `throughputFromSource` by:
`throughputFromSource = (octetTotalCountFromSource - prevOctetTotalCountFromSource) / (flowEndSecondsFromSource - prevFlowEndSecondsFromSource)`

If `flowEndSeconds` of the new record is the latest, we update common fields. We update `throughput` by:
`throughput = throughputFromSource`

Especially, for the first record which doesn't have a "prev" record, 
`throughputFromSource = octetTotalCountFromSource / (flowEndSecondsFromSource - flowStartSecondsFromSource)`

**Describe alternatives you've considered**

Two open questions:
- Shall we also add the calculation of `reverseThroughput`, following the similar format?
- Shall we keep the node specific fields as they are for now, or remove them all, and store those info in memory?

**Additional context**
The reason for adding `throughputFromSource` and `throughputFromDestination` is we also need to report node throughput to users. We can compute it at flow aggregator side.
Based on our previous discussion, the current only use case of node specific fields is use deltaCount to compute node throughput at the front end. Since we are now computing it in flow aggregator, currently there's no use case of node specific fields.

PR #270, opened 2021-10-26
Add throughput calculation logic

heanlan:
- Background
Originally we don't provide throughput information in the IPFIX record, we compute throughput(byte/s) at the front end, e.g. aggregating `octetDeltaCount` of all the records of a connection in last 60s, and divide by 60s. In order to decrease the computation complexity at the front end, we decide to add throughput calculation into flow-aggregator, and report the latest throughput in the form of IPFIX record fields. We take this PR as the first step of supporting flow-aggregator's throughput calculation, by simply adding the fields and basic calculation logic. As a TODO, we will consider making changes on when to trigger the records exporting, based on whether we receive the equal number of records from both source/destination nodes, to make the throughput more accurate.

- Methods
For each connection, we calculate throughput, reverseThroughput, node throughput, reverse node throughput in the aggregation process. 
Node throughput of an aggregated flow record is defined as: `diff(octetTotalCount) / diff(flowEndSeconds)` between the newly-received record and the last received record from the same node.
Common field throughput is defined as: if the newly-received record has the latest `flowEndSeconds` timestamp, common field throughput will be updated by the node-throughput calculated by the newly-received record.

- Fields added to IPFIX record:

```
antreaFlowEndSecondsElementList = []string{
"flowEndSecondsFromSourceNode",
"flowEndSecondsFromDestinationNode",
}

antreaThroughputElementList = []string{
"throughput",
"reverseThroughput",
}

antreaSourceThroughputElementList = []string{
"throughputFromSourceNode",
"reverseThroughputFromSourceNode",
}

antreaDestinationThroughputElementList = []string{
"throughputFromDestinationNode",
"reverseThroughputFromDestinationNode",
}
```

- Rationale of adding the fields:
    - `antreaFlowEndSecondsElementList` are added as I suppose, when we report throughputFromNode values, it makes more sense if they are bonding with timestampsFromNode. For example, in the case when we receive n records from src node, while only n-1 records from dst node, and we need to export the aggregated record now due to active timeout. If we don't provide `flowEndSecondsFromSourceNode`, `flowEndSecondsFromDestinationNode`, the front end will treat `throughputFromSourceNode` and `throughputFromDestinationNode` both w.r.t the common field `flowEndSeconds`, which is updated by the n-th record from src node. In fact, we haven't received the record from the destination node w.r.t. the time `flowEndSeconds`.

    - `antreaThroughputElementList`,  `antreaSourceThroughputElementList`,  `antreaDestinationThroughputElementList` are added as we have the requirements to display per-connection throughput, reverse throughput, node throughput, reverse node throughput information at the front end. They are grouped in three separate groups in order to make use of their index to make the code more compact and clear, reference to statsElementList, antreaSourceStatsElementList, antreaDestinationStatsElementList.

Fixes: #265


heanlan:
@srikartati @zyiou could you please help take a look at the overall logic to see whether it's on the right track, mainly in aggregate.go? Thanks

I know there's a lot to be optimized on the coding style, e.g. avoiding duplicated code. I'll work on that at the same time.

pkg/intermediate/aggregate_test.go
```diff
@@ -52,18 +52,32 @@ var (
...
+		"octetTotalCountFromDestinationNode",
+		"reverseOctetTotalCountFromDestinationNode",
+	}
+	antreaThroughputElements = []string{
+		"throughput",
+		"reverseThroughput",
+		"throughputFromSourceNode",
+		"throughputFromDestinationNode",
+		"reverseThroughputFromSourceNode",
+		"reverseThroughputFromDestinationNode",
```
zyiou, line 80: It would be better to group the stats by non-reverse and reverse like we did before.

pkg/intermediate/aggregate.go
```diff
@@ -524,59 +548,101 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
...
+	// Update the reverseThroughput fields
+	reverseThroughput := reverseTotalCountDiff / uint64(flowEndSecondsDiff)
+	if fillSrcStats {
+		ie, _, _ := existingRecord.GetInfoElementWithValue("reverseThroughputFromSourceNode")
+		ie.SetUnsigned64Value(reverseThroughput)
+	}
+	if fillDstStats {
+		ie, _, _ := existingRecord.GetInfoElementWithValue("reverseThroughputFromDestinationNode")
+		ie.SetUnsigned64Value(reverseThroughput)
+	}
```
zyiou, line 642: I don't think we should hardcode creation of fields here. Instead, we can take these fields as input like how we handle stats. We can add input like `timestampFields` to take all flowStartSeconds, flowEndSeconds, flowEndSecondsFromSourceNode etc. related fields.

pkg/intermediate/aggregate.go
```diff
@@ -524,59 +548,101 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
...
+						commonExistingIeWithValue.SetUnsigned64Value(dstIe.GetUnsigned64Value())
+					}
+				}
+			}
 		} else {
 			return fmt.Errorf("element with name %v in statsElements not present in the incoming record", element)
 		}
 	}
+
+	if flowEndSecondsDiff == 0 {
```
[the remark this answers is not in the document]
heanlan: although it's very unlikely - in the case when an agent is down and restarted, it will replay some flow records, when the first replayed record is exactly the one received before it's down.

I guess adding this zero check can avoid division by zero warning, in some extreme cases

pkg/intermediate/aggregate.go
heanlan: things are a little bit different for throughput elements.. As for the stats, we have all the elements in statsElementList in the flow record, len(statsElementList) = len(antreaSourceStatsElements) = len(antreaDestinationStatsElements), we can take advantage of the index, and expand to source and destination fields. So this is the challenge.

I also agree this section can be optimized. I'll think about how to re-structure it.

pkg/intermediate/aggregate.go
```diff
@@ -645,9 +693,122 @@ func (a *AggregationProcess) addFieldsForStatsAggregation(record entities.Record
...
+	}
+	reverseByteCount, err = getUnsigned64ValueByIeName(record, "reverseOctetTotalCount")
+	if err != nil {
+		return err
+	}
+	incomingVal := byteCount / (uint64(timeEnd - timeStart))
+	reverseIncomingVal := reverseByteCount / (uint64(timeEnd - timeStart))
+	throughputVals := []uint64{incomingVal, reverseIncomingVal}
+
+	for i, element := range antreaThroughputElements {
```
heanlan, line 720: I re-wrote the code where we initialize throughput elements, grouping the throughput elements by common fields/src fields/ dst fields. Does the current format make more sense to you? @zyiou

pkg/intermediate/aggregate.go
```diff
@@ -645,9 +693,122 @@ func (a *AggregationProcess) addFieldsForStatsAggregation(record entities.Record
...
+	// Initialize the elements flowEndSecondsFromSource/DestinationNode.
+	if timeEnd != uint32(0) {
+		var srcTime, dstTime uint32
+		if fillSrcStats {
+			srcTime = timeEnd
+		}
+		if fillDstStats {
+			dstTime = timeEnd
+		}
+		ie, err := registry.GetInfoElement("flowEndSecondsFromSourceNode", registry.AntreaEnterpriseID)
```
heanlan, line 765: I also agree hardcode the element name is not good. Do you think it makes more sense to have a group containing only `flowEndSecondsFromSourceNode` `flowEndSecondsFromDestinationNode`?

pkg/intermediate/aggregate.go
```diff
@@ -645,9 +693,122 @@ func (a *AggregationProcess) addFieldsForStatsAggregation(record entities.Record
...
+			return err
+		}
+		if err = record.AddInfoElement(entities.NewUnsigned32InfoElement(ie, dstTime)); err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
+func (a *AggregationProcess) updateFlowEndSecondsFromNodes(incomingRecord, existingRecord entities.Record, ieName string, ieVal uint32) uint32 {
```
heanlan, line 783: Is it considering not good to pass the record as a parameter? I added multiple common functions to avoid code repetition, but they all need to have record as a parameter.

pkg/intermediate/aggregate.go
```diff
@@ -482,14 +482,24 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
...
 			}
+			// Update the flowEndSecondsFromSource/DestinationNode fields, and compute
+			// the time difference between the incoming record and the last record.
+			if fillSrcStats {
+				prevFlowEndSeconds = a.updateFlowEndSecondsFromNodes(incomingRecord, existingRecord, "flowEndSecondsFromSourceNode", incomingVal)
+			}
+			if fillDstStats {
+				prevFlowEndSeconds = a.updateFlowEndSecondsFromNodes(incomingRecord, existingRecord, "flowEndSecondsFromDestinationNode", incomingVal)
+			}
+			flowEndSecondsDiff = incomingVal - prevFlowEndSeconds
```
dreamtalen, line 502: Is it possible that `incomingVal - prevFlowEndSeconds` is less than zero? I noticed that you are using `uint32` type for `flowEndSecondsDiff`, just worried about if there could be a corner case here.

heanlan: yes, i believe that's possible in the case when agent down and restart, replaying previous records. Although we currently don't have the replay mechanism as a reference.

I added the logic to ensure we only do the aggregation when the incoming record is the latest wrt its coming node. That's to say, if a record from source node doesn't have the latest `flowEndSecondsFromSourceNode`, we will skip the aggregation of this record, since it doesn't bring new information.

pkg/intermediate/aggregate.go
```diff
@@ -482,13 +482,27 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
...
+			}
+			// Update the flowEndSecondsFromSource/DestinationNode fields, and compute
+			// the time difference between the incoming record and the last record.
+			if fillSrcStats {
+				prevFlowEndSecondsFromNode = a.updateFlowEndSecondsFromNodes(incomingRecord, existingRecord, "flowEndSecondsFromSourceNode", newFlowEndSeconds)
+			}
+			if fillDstStats {
+				prevFlowEndSecondsFromNode = a.updateFlowEndSecondsFromNodes(incomingRecord, existingRecord, "flowEndSecondsFromDestinationNode", newFlowEndSeconds)
+			}
+			flowEndSecondsDiff = newFlowEndSeconds - prevFlowEndSecondsFromNode
```
dreamtalen, line 502: Since `flowEndSecondsDiff` is uint32 type, it would not be smaller than 0. How about changing these lines to:
```
if newFlowEndSeconds < prevFlowEndSecondsFromNode {
   return nil
}
flowEndSecondsDiff = newFlowEndSeconds - prevFlowEndSecondsFromNode
```

heanlan: updated. That's right, thanks for catching it.

pkg/intermediate/aggregate.go
```diff
@@ -645,9 +694,124 @@ func (a *AggregationProcess) addFieldsForStatsAggregation(record entities.Record
...
+		incomingIe, _, _ := incomingRecord.GetInfoElementWithValue("flowStartSeconds")
+		existingVal = incomingIe.GetUnsigned32Value()
+	}
+	if existingVal < ieVal {
+		existingIe.SetUnsigned32Value(ieVal)
+	}
+	return existingVal
+}
+
+func getUnsigned64ValueByIeName(record entities.Record, ieName string) (uint64, error) {
```
zyiou, line 799: we can consider adding similar methods for record interface in the future

pkg/intermediate/aggregate.go
zyiou: I haven't figured out a good workaround. We can add a TODO here since it will not be a big change.

pkg/intermediate/aggregate.go
```diff
@@ -645,9 +694,124 @@ func (a *AggregationProcess) addFieldsForStatsAggregation(record entities.Record
...
+	existingIe, _, _ := existingRecord.GetInfoElementWithValue(ieName)
+	existingVal := existingIe.GetUnsigned32Value()
+	// When the incoming record is the first record from one node, the previous record
+	// doesn't exist. time_diff = flowEndSeconds - flowStartSeconds
+	if existingVal == 0 {
+		incomingIe, _, _ := incomingRecord.GetInfoElementWithValue("flowStartSeconds")
+		existingVal = incomingIe.GetUnsigned32Value()
+	}
+	if existingVal < ieVal {
+		existingIe.SetUnsigned32Value(ieVal)
```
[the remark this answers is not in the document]
heanlan: As I discussed with Yongming above, although I think it is very unlikely, it happens when an agent is down and restart, replaying some previous flow records. 

Do you think it worth to be considered as an edge case? I'll add a comment to explain it if we decide to keep it. We haven't implemented the flow exporter replay mechanism yet. If based on current flow exporter design, existingVal is larger or equal to ieVal is impossible.

pkg/intermediate/aggregate.go
heanlan: marked as a TODO

pkg/intermediate/aggregate.go
```diff
@@ -482,14 +491,35 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
...
 			}
+			// Update the flowEndSecondsFromSource/DestinationNode fields, and compute
+			// the time difference between the incoming record and the last record.
+			if fillSrcStats {
+				prevFlowEndSeconds = a.updateFlowEndSecondsFromNodes(incomingRecord, existingRecord, true, incomingVal)
+			}
+			if fillDstStats {
+				prevFlowEndSeconds = a.updateFlowEndSecondsFromNodes(incomingRecord, existingRecord, false, incomingVal)
+			}
+			// For inter-node flows, skip the aggregation process if the incoming record
```
heanlan, line 511: Change: Include the consideration of intra-node, we skip the aggregation for intra-node if we receive the same record as the last one. (normally we will receive one record from source node, one record from destination node. Here source node and destination is the same node, so we will receive two identical records from the node.

pkg/intermediate/aggregate.go
```diff
@@ -648,6 +712,132 @@ func (a *AggregationProcess) addFieldsForStatsAggregation(record entities.Record
...
+	if err != nil {
+		return err
+	}
+	reverseByteCount, err = getUnsigned64ValueByIeName(record, "reverseOctetTotalCount")
+	if err != nil {
+		return err
+	}
+
+	// Initialize flowEndSecondsFromSourceNode and flowEndSecondsFromDestinationNode.
+	for _, ieName := range antreaFlowEndSecondsElements {
```
heanlan, line 744: Change: Adding a group `antreaFlowEndSecondsElements ` instead of "hard-coding" the fields name. Also to make the pattern more consistent with other extra fields at flowaggregator.go on Antrea side - [code](https://github.com/antrea-io/antrea/blob/86c414e3a566538ffd8d6ca01f0e84ca25031498/pkg/flowaggregator/flowaggregator.go#L124)

pkg/intermediate/aggregate.go
```diff
@@ -648,6 +712,132 @@ func (a *AggregationProcess) addFieldsForStatsAggregation(record entities.Record
...
+			value = timeEnd
+		}
+		if err = record.AddInfoElement(entities.NewUnsigned32InfoElement(ie, value)); err != nil {
+			return err
+		}
+	}
+
+	// Initialize the throughput elements.
+	var incomingVal, reverseIncomingVal uint64
+	// For the edge case when the record has the same timeEnd and timeStart values,
```
heanlan, line 760: Change: Adding an edge case handling, since for some of the conntrack connections, when the start time is zero, both start and end time will be set reset to the current time - [code](https://github.com/antrea-io/antrea/blob/c135609aaa9b63d74842c7cf2f7981d1c44379f4/pkg/agent/flowexporter/connections/conntrack_connections.go#L242)

heanlan (commented):
Hi @zyiou @dreamtalen , thanks for the reviews.
I made the corresponding changes on Antrea side in [PR2692](https://github.com/antrea-io/antrea/pull/2692), along with some small changes in this PR since last review, which are pointed out below.
I apologized for mixing it up with the previous commit. Please take a look when you have time. Thanks!

pkg/intermediate/aggregate.go
zyiou: For intra-node, do you mean we will receive two records for each flow?

heanlan: I made it wrong in the comment. Sorry for the confusion. I've corrected it. For both intra-node and inter-node, we will send warning if the flowEndSeconds value is not the latest.

zyiou (approved):
LGTM

PR #271, opened 2021-10-28
Update GetRecords in AggregationProcess to return map format records

yanjunz97:
This commit updates the return type of GetRecords in AggregationProcess from string to map, which makes it more easier to be structured.

### DIFF
--- a/pkg/intermediate/aggregate.go
+++ b/pkg/intermediate/aggregate.go
@@ -215,19 +215,20 @@ func (a *AggregationProcess) GetExpiryFromExpirePriorityQueue() time.Duration {
 	return a.inactiveExpiryTimeout
 }
 
-// GetRecords returns the string format flow records by FlowKey from flowKeyRecordMap.
-// Returns partial match if FlowKey is not complete.
-// Returns all the flow records if FlowKey is not provided.
-func (a *AggregationProcess) GetRecords(flowKey *FlowKey) []string {
+// GetRecords returns map format flow records given a flow key.
+// The key of the map is the element name and the value is the IE object.
+// Returns partially matched flow records if the flow key is not complete.
+// Returns all the flow records if the flow key is not provided.
+func (a *AggregationProcess) GetRecords(flowKey *FlowKey) []map[string]interface{} {
 	a.mutex.Lock()
 	defer a.mutex.Unlock()
 
-	var records []string
+	var records []map[string]interface{}
 	// Complete filter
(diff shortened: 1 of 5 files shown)


yanjunz97:
As Antonin suggests, the record should be able to be encoded to JSON format. I changed the return type of the `GetRecords` function to a map of interface. Currently the `GetString()` function in Record is not used. Shall we still keep it?

srikartati:
> As Antonin suggests, the record should be able to be encoded to JSON format. I changed the return type of the `GetRecords` function to a map of interface. Currently the `GetString()` function in Record is not used. Shall we still keep it?

I think JSON format is good right. We can leave it to the user to change in the format they want. If we need some other format, we can add it later. Lets remove for now?

yanjunz97:
> > As Antonin suggests, the record should be able to be encoded to JSON format. I changed the return type of the `GetRecords` function to a map of interface. Currently the `GetString()` function in Record is not used. Shall we still keep it?
> 
> I think JSON format is good right. We can leave it to the user to change in the format they want. If we need some other format, we can add it later. Lets remove for now?

Removed.

pkg/intermediate/aggregate.go
```diff
@@ -215,19 +215,19 @@ func (a *AggregationProcess) GetExpiryFromExpirePriorityQueue() time.Duration {
 	return a.inactiveExpiryTimeout
 }
 
-// GetRecords returns the string format flow records by FlowKey from flowKeyRecordMap.
-// Returns partial match if FlowKey is not complete.
-// Returns all the flow records if FlowKey is not provided.
-func (a *AggregationProcess) GetRecords(flowKey *FlowKey) []string {
+// GetRecords returns map format flow records given a flow key.
```
srikartati, line 218: Let's explain the map format clearly, where key is element name and value is the IE object.

pkg/intermediate/aggregate.go
```diff
@@ -215,19 +215,19 @@ func (a *AggregationProcess) GetExpiryFromExpirePriorityQueue() time.Duration {
...
+// Returns partial matched flow records if the flow key is not complete.
```
srikartati, line 219: "partially matched"

yanjunz97: Done

pkg/intermediate/aggregate.go
yanjunz97: Done

srikartati (approved):
lgtm

dreamtalen (approved):
LGTM

PR #276, opened 2022-02-22
Improve dateTime* type support

ivan4th:
Use `time.Time` to represent `dateTime*` types (breaking change, but I think it makes sense)
Support `dateTimeMicroseconds` and `dateTimeNanoseconds`

Helpful pointers (besides RFC 7011):
[Timesamp handling in the Wireshark dissector for IPFIX](https://github.com/wireshark/wireshark/blob/master/epan/dissectors/packet-netflow.c#L5431)
[NTP Timestamp format](https://datatracker.ietf.org/doc/html/rfc2030#section-3)

### DIFF
--- a/cmd/collector/collector.go
+++ b/cmd/collector/collector.go
@@ -87,7 +87,6 @@ func printIPFIXMessage(msg *entities.Message) {
 				case entities.Unsigned8:
 					fmt.Fprintf(&buf, "    %s: %v \n", elem.Name, ie.GetUnsigned8Value())
 				case entities.Unsigned16:
-
 					fmt.Fprintf(&buf, "    %s: %v \n", elem.Name, ie.GetUnsigned16Value())
 				case entities.Unsigned32:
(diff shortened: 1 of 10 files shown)


dreamtalen:
Thank your contribution @ivan4th! 
Since this is a performance-critical component and we're using it in another open source project at https://github.com/antrea-io/antrea/tree/main/pkg/agent/flowexporter, we may need more time to evaluate the performance impact from changing uint32 to DateTime type. 
May I ask your purpose for this PR: just change the type to meet the RFC 7011 standard (actually dateTimeSeconds data type is an unsigned 32-bit integer according to https://www.rfc-editor.org/rfc/rfc7011.html#section-6.1.7)? Or you need `DateTimeMicroseconds, DateTimeNanoseconds` in your project? 
Also, there are multiple test failures in the Github CI, could you please take a look at them? Please let us know if you need any help on refactoring the CI tests?
Thanks!

PR #277, opened 2022-02-22
Don't stop upon packet decode errors

ivan4th:
We shouldn't stop the collector if it happens to receive a dataset packet before the corresponding template


salv-orlando (approved):
Test

salv-orlando (commented):
test2

antoninbas:
Superseded by https://github.com/vmware/go-ipfix/pull/352

PR #278, opened 2022-02-22
Fixed the throughput burst issue

yanjunz97:
This PR fixes the throughput burst issues. 

Previously we reset `totalCount` and `deltaCount` static elements after the record is sent. As the throughput is calculated by `diff(octetTotalCount) / diff(flowEndSeconds)`, the existing `octetTotalCount` should not be reset to 0. 

The PR also adds reset of throughput after the the record is sent. This deals with the condition that a record is sent by both active and inactive expiry interval. The receiver will get two records with the same `flowEndSecond`. In this case, the record sent by inactive expiry interval timeout should have the throughput as 0.

### DIFF
--- a/pkg/intermediate/aggregate.go
+++ b/pkg/intermediate/aggregate.go
@@ -91,6 +91,9 @@ func InitAggregationProcess(input AggregationInput) (*AggregationProcess, error)
 		if (len(input.AggregateElements.StatsElements) != len(input.AggregateElements.AggregatedSourceStatsElements)) || (len(input.AggregateElements.StatsElements) != len(input.AggregateElements.AggregatedDestinationStatsElements)) {
 			return nil, fmt.Errorf("stats elements, source stats elements and destination stats elemenst length should be equal")
 		}
+		if (len(input.AggregateElements.ThroughputElements) != len(input.AggregateElements.SourceThroughputElements)) || (len(input.AggregateElements.ThroughputElements) != len(input.AggregateElements.DestinationThroughputElements)) {
+			return nil, fmt.Errorf("throughput elements, source throughput elements and destination throughput elemenst length should be equal")
+		}
 	}
 	return &AggregationProcess{
 		make(map[FlowKey]*AggregationFlowRecord),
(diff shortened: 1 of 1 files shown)


yanjunz97:
I added a commit to refactor the code which includes two changes:

- Reseting throughput is put in the same function which resets the statElements, as I did not come up with the case that these two functions may be called separately.
- Using a loop instead of repeating code to reset the value.

I think these changes may not have influences on performance but makes the code cleaner. But if there is any concern regarding the  change, I can revert it.

pkg/intermediate/aggregate.go
```diff
@@ -640,28 +643,37 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
...
+func (a *AggregationProcess) ResetStatAndThroughputElementsInRecord(record entities.Record) error {
 	statsElementList := a.aggregateElements.StatsElements
 	antreaSourceStatsElements := a.aggregateElements.AggregatedSourceStatsElements
 	antreaDestinationStatsElements := a.aggregateElements.AggregatedDestinationStatsElements
 	for i, element := range statsElementList {
-		if ieWithValue, _, exist := record.GetInfoElementWithValue(element); exist {
-			ieWithValue.ResetValue()
-		} else {
-			return fmt.Errorf("element with name %v in statsElements is not present in the record", element)
+		// TotalCount static elements should be skipped as they are used in the thoughput calculation.
```
dreamtalen, line 654: static -> statistic

dreamtalen (commented):
LGTM, thanks yanjun fixing this issue.

heanlan (approved):
Haven't review yet. Test approve.

pkg/intermediate/aggregate.go
heanlan, line 654:
```suggestion
		// TotalCount statistic elements should not be reset to zeroes as they are used in the
		// throughput calculation. 
```

yanjunz97: Done

yanjunz97: Done

heanlan (approved):
LGTM

heanlan (approved):

PR #290, opened 2022-12-01
Renew fake cert for tests

yanjunz97:


pkg/test/certs.go
```diff
@@ -15,134 +15,149 @@
 package test
 
 const (
-	// TODO: update the certs before 2022-02-01
+	// TODO: update the certs before 2025-02-01
```
dreamtalen, line 18: As the `-days 365` arguments in commands of README, it should be updated before 2023? Or maybe you should change the `-days` in README?

yanjunz97: I used `-days 1000` when generating these certificates. I thought 365 might be more common so I used it in the doc, but this inconsistency does introduce confusion. I updated the README to use `-days 1000`. Thanks!

dreamtalen (approved):
LGTM, thanks Yanjun fixing this.

PR #293, opened 2022-12-22
Upgrading old dependencies

antoninbas:
This is a maintenance commit, which takes care of upgrading dependencies:
* We upgrade to a supported version of Golang (v1.19)
* We upgrade most Go modules to their latest version
* We upgrade the base image for Docker containers to ubuntu:22.04

Signed-off-by: Antonin Bas <abas@vmware.com>


yanjunz97 (approved):
LGTM

PR #308, opened 2023-06-05
Add fields for Egress information support

dreamtalen:
Add two new fields egressName and egressIP for the visibility of Egress information in flow records.

Changes in Antrea: https://github.com/antrea-io/antrea/pull/5088

### DIFF
--- a/cmd/collector/collector.go
+++ b/cmd/collector/collector.go
@@ -87,7 +87,6 @@ func printIPFIXMessage(msg *entities.Message) {
 				case entities.Unsigned8:
 					fmt.Fprintf(&buf, "    %s: %v \n", elem.Name, ie.GetUnsigned8Value())
 				case entities.Unsigned16:
-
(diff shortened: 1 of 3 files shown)


pkg/intermediate/aggregate.go
```diff
@@ -338,6 +338,13 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
 	} else {
 		klog.Warning("FlowType does not exist in current record.")
 	}
+
+	// Ignore records from the Egress node for Pod-to-External flows
+	// if the Egress node is not the same as the Source node.
+	if isRecordFromEgressNode(flowType, record) {
```
antoninbas, line 344: why don't we skip sending it from the Agent instead?

dreamtalen: Thanks Antonin bringing it up, I misunderstood the words from Srikar 'at Egress node' as 'from Egress node.' I agree skip sending it from the Agent is a better approach.

antoninbas (approved):

heanlan (approved):

Issue #306, opened 2023-05-23
ipfix collector pod in error state when trying to deploy from main branch


Amoghrd:
**Describe the bug**
The IPFIX collector pod is not coming up and in error state with error `Error: unknown flag: --v` 

**To Reproduce**
Steps:
- Cloned the repo
- cd hack
- ./generate-manifest-collector.sh --mode dev --port 4739 --proto tcp > ../build/yamls/ipfix-collector.yaml
- oc apply -f build/yamls/ipfix-collector.yaml

**Expected behavior**
ipfix pod has to be up and running

**Actual Behavior**
<img width="1491" alt="Screenshot 2023-05-23 at 4 16 32 PM" src="https://github.com/vmware/go-ipfix/assets/65666095/1a215cd6-40f5-46f7-9d80-dfec04d624b9">

<img width="794" alt="Screenshot 2023-05-23 at 4 16 49 PM" src="https://github.com/vmware/go-ipfix/assets/65666095/1f9689c7-a0be-4470-88a2-6858c09ebb96">

**Versions**
Observed while deploying with the main branch

**Additional context**
The error is not hit when line [ipfix-collector.yaml#L51](https://github.com/vmware/go-ipfix/blob/main/build/yamls/ipfix-collector.yaml#L51) is removed and the pod comes up fine

PR #311, opened 2023-06-13
Fix collector klog flag not supported error

heanlan:
Add registration of klog flags

Fixes: #306

### DIFF
--- a/build/yamls/ipfix-collector.yaml
+++ b/build/yamls/ipfix-collector.yaml
@@ -48,7 +48,6 @@ spec:
     spec:
       containers:
       - args:
-        - --v=0
         - --ipfix.port=4739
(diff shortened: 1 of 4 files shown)


build/yamls/ipfix-collector.yaml
```diff
@@ -48,7 +48,6 @@ spec:
     spec:
       containers:
       - args:
-        - --v=0
```
yuntanghsu, line 51: Is it 0 a default value?

heanlan: yes, 0 is default log level

yuntanghsu (approved):

yanjunz97 (approved):
LGTM

PR #314, opened 2023-08-24
L7 Visibility support in Antrea

tushartathgur:
Adding L7 visibility fields to the Ipfix table.

- isL7: This field helps us differentiate an L7 flow from a non L7 flow, which would be helpful in Theia when trying to get data for only L7 flows
- httpVals: The field is added to contain all the http values and store them as a string with a delimiter (`Hostname<>URL<>UserAgent<>ContentType<>Method<>Protocol<>Status<>ContentLength`). This helps us to store multiple http Vals for a single flow which is possible in case of persistent http.

### DIFF
--- a/build/yamls/collector/templates/ipfix-collector.yaml
+++ b/build/yamls/collector/templates/ipfix-collector.yaml
@@ -44,7 +44,9 @@ spec:
     spec:
       containers:
       - args:
-        - --v=0
+        {{- if .Values.logVerbosity }}
+        - --v={{ .Values.logVerbosity }}
+        {{- end }}
         - --ipfix.port={{ .Values.port }}
         - --ipfix.transport={{ .Values.protocol }}
         image: projects.registry.vmware.com/antrea/ipfix-collector:{{ .Values.image_tag }}
--- a/build/yamls/collector/values.yaml
+++ b/build/yamls/collector/values.yaml
@@ -1,3 +1,5 @@
 protocol: tcp
 port: 4739
 image_tag: latest
+# -- Log verbosity switch for IPFIX collector.
+logVerbosity: 0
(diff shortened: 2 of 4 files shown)


pkg/registry/registry_antrea.csv
```diff
@@ -54,3 +54,5 @@ ElementID,Name,Abstract Data Type,Data Type Semantics,Status,Description,Units,R
 152,flowEndSecondsFromDestinationNode,unsigned32,,current,,,,,,,,56506,
 153,egressName,string,,current,,,,,,,,56506,
 154,egressIP,string,,current,,,,,,,,56506,
+155,isL7,string,,current,,,,,,,,56506,
```
antoninbas, line 57: what is the meaning of `isL7`? I think the format allows for quick description of each IE?

pkg/registry/registry_antrea.csv
```diff
@@ -54,3 +54,5 @@ ElementID,Name,Abstract Data Type,Data Type Semantics,Status,Description,Units,R
...
+156,httpVals,string,,current,,,,,,,,56506,
```
antoninbas, line 58: a similar question: is this going to be some JSON object? What will be in this object?

pkg/registry/registry_antrea.csv
tushartathgur: Thanks for pointing it out, I have added the meaning of the added fields in the description of the PR.

pkg/registry/registry_antrea.csv
tushartathgur: It is not going to be a json value, it will be a string containing all l7 http fields separated by a delimiter.

antoninbas:
@tushartathgur I think we can continue the conversation as part of your Antrea PR. There is more context there.

PR #315, opened 2023-09-15
Adding L7 visibility fields

tushartathgur:
Adding L7 visibility fields to the Ipfix table.

l7ProtocolName: field is used to store application layer protocol name, field will be empty if application layer flow is not exported.
httpVals: The field is added to contain all the http values and store them as a string of Json similar to labels. This helps us to store multiple http Vals for a single flow which is possible in case of persistent http.

Mistakenly removed #314

### DIFF
--- a/pkg/registry/registry_antrea.csv
+++ b/pkg/registry/registry_antrea.csv
@@ -54,3 +54,5 @@ ElementID,Name,Abstract Data Type,Data Type Semantics,Status,Description,Units,R
 152,flowEndSecondsFromDestinationNode,dateTimeSeconds,,current,,,,,,,,56506,
 153,egressName,string,,current,,,,,,,,56506,
 154,egressIP,string,,current,,,,,,,,56506,
+155,isL7,boolean,,current,,,,,,,,56506,
+156,httpVals,string,,current,,,,,,,,56506,
--- a/pkg/registry/registry_antrea.csv
+++ b/pkg/registry/registry_antrea.csv
@@ -54,5 +54,5 @@ ElementID,Name,Abstract Data Type,Data Type Semantics,Status,Description,Units,R
 152,flowEndSecondsFromDestinationNode,dateTimeSeconds,,current,,,,,,,,56506,
 153,egressName,string,,current,,,,,,,,56506,
 154,egressIP,string,,current,,,,,,,,56506,
-155,isL7,boolean,,current,,,,,,,,56506,
+155,l7ProtocolName,string,,current,,,,,,,,56506,
 156,httpVals,string,,current,,,,,,,,56506,
--- a/pkg/registry/registry_antrea.go
+++ b/pkg/registry/registry_antrea.go
@@ -72,8 +72,10 @@ func loadAntreaRegistry() {
 	registerInfoElement(*entities.NewInfoElement("throughputFromDestinationNode", 148, 4, 56506, 8), 56506)
 	registerInfoElement(*entities.NewInfoElement("reverseThroughputFromSourceNode", 149, 4, 56506, 8), 56506)
 	registerInfoElement(*entities.NewInfoElement("reverseThroughputFromDestinationNode", 150, 4, 56506, 8), 56506)
-	registerInfoElement(*entities.NewInfoElement("flowEndSecondsFromSourceNode", 151, 14, 56506, 4), 56506)
-	registerInfoElement(*entities.NewInfoElement("flowEndSecondsFromDestinationNode", 152, 14, 56506, 4), 56506)
+	registerInfoElement(*entities.NewInfoElement("flowEndSecondsFromSourceNode", 151, 3, 56506, 4), 56506)
+	registerInfoElement(*entities.NewInfoElement("flowEndSecondsFromDestinationNode", 152, 3, 56506, 4), 56506)
 	registerInfoElement(*entities.NewInfoElement("egressName", 153, 13, 56506, 65535), 56506)
 	registerInfoElement(*entities.NewInfoElement("egressIP", 154, 13, 56506, 65535), 56506)
+	registerInfoElement(*entities.NewInfoElement("isL7", 155, 11, 56506, 1), 56506)
+	registerInfoElement(*entities.NewInfoElement("httpVals", 156, 13, 56506, 65535), 56506)
 }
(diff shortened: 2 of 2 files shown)


pkg/registry/registry_antrea.go
```diff
@@ -72,8 +72,10 @@ func loadAntreaRegistry() {
 	registerInfoElement(*entities.NewInfoElement("throughputFromDestinationNode", 148, 4, 56506, 8), 56506)
 	registerInfoElement(*entities.NewInfoElement("reverseThroughputFromSourceNode", 149, 4, 56506, 8), 56506)
 	registerInfoElement(*entities.NewInfoElement("reverseThroughputFromDestinationNode", 150, 4, 56506, 8), 56506)
-	registerInfoElement(*entities.NewInfoElement("flowEndSecondsFromSourceNode", 151, 14, 56506, 4), 56506)
-	registerInfoElement(*entities.NewInfoElement("flowEndSecondsFromDestinationNode", 152, 14, 56506, 4), 56506)
+	registerInfoElement(*entities.NewInfoElement("flowEndSecondsFromSourceNode", 151, 3, 56506, 4), 56506)
```
dreamtalen, line 75: Could you help take a look at this @heanlan? I can't remember the reason behind the decision to use `uint32` for `flowEndSecondsFromSource/DestNode` instead of `dateTimeSeconds`.

dreamtalen (commented):
The L7 fields LGTM. 
Could you change the PR and commit message title? It appears to be identical to the Antrea one, which is actually implementing L7 visibility support.

pkg/registry/registry_antrea.go
heanlan: They should keep unchanged as uint32. We are following the same datatype with flowEndSeconds. https://github.com/vmware/go-ipfix/blob/ec54b43babc78f1c5f5e468333d4ebd72419455a/pkg/registry/registry_IANA.go#L148

elton-furtado (approved):

pkg/registry/registry_antrea.go
heanlan: This change should be reverted. The datatype should be change back to 14.

heanlan: Sorry for creating some confusions here. I just noticed there are some discrepancies in both Antrea and Go-ipfix code. `FlowEndSecondsFromSourceNode` and `FlowEndSecondsFromDestinationNode` should have the same datatype with `FlowEndSeconds`. And the datatype is `dateTimeSeconds`. @tushartathgur could you revert this change and also change the following lines to `dateTimeSeconds`
https://github.com/vmware/go-ipfix/blob/ec54b43babc78f1c5f5e468333d4ebd72419455a/pkg/registry/registry_antrea.csv#L53-L54

The antrea doc is wrong:
https://github.com/antrea-io/antrea/blob/4806be3c14e21a75dd3b8e876785fbc947439474/docs/network-flow-visibility.md?plain=1#L486-L487

tushartathgur: Sure, Creating a separate PR for this as this is not part of L7Visibility.
Though I am confused, the registry_antrea.go should create the elements according to their datatype, however for these two fields, it should not have generated DateTimeSeconds format if the datatype was unsigned32 !

```
registerInfoElement(*entities.NewInfoElement("flowEndSecondsFromSourceNode", 151, 14, 56506, 4), 56506)
registerInfoElement(*entities.NewInfoElement("flowEndSecondsFromDestinationNode", 152, 14, 56506, 4), 56506)
```

antoninbas (commented):
I still wonder if the `isL7` boolean should be replaced with some "enum" instead. For example: `l7FlowType: "http"` (an empty of missing IE would mean that we don't have L7 data). I wonder if that could be more future-proof, if we ever want to support other protocols?

What do you think @dreamtalen @heanlan @tushartathgur ?

pkg/registry/registry_antrea.go
heanlan: The underlying structure of DateTimeSeconds is uint32: https://www.rfc-editor.org/rfc/rfc7011.html#section-6.1.7

tushartathgur: I mean the registry_antrea.go is autogenerated, it should not have changed the unsigned32 to dateTimeSeconds

heanlan:
> I still wonder if the `isL7` boolean should be replaced with some "enum" instead. For example: `l7FlowType: "http"` (an empty of missing IE would mean that we don't have L7 data). I wonder if that could be more future-proof, if we ever want to support other protocols?

I agree it's better to make the change now compared to do it later.

pkg/registry/registry_antrea.go
heanlan: Let's keep these two lines of change, i.e. 14 -> 3, with this PR, as we want to correct the errors. As you mentioned, we can do the uint32 -> dateTimeSeconds changes in a separate PR if we would like to.

dreamtalen:
> > I still wonder if the `isL7` boolean should be replaced with some "enum" instead. For example: `l7FlowType: "http"` (an empty of missing IE would mean that we don't have L7 data). I wonder if that could be more future-proof, if we ever want to support other protocols?
> 
> I agree it's better to make the change now compared to do it later.

Sounds good to me as well.

tushartathgur:
> I still wonder if the `isL7` boolean should be replaced with some "enum" instead. For example: `l7FlowType: "http"` (an empty of missing IE would mean that we don't have L7 data). I wonder if that could be more future-proof, if we ever want to support other protocols?
> 
> What do you think @dreamtalen @heanlan @tushartathgur ?

Sound good to me, Adding in the next patch

pkg/registry/registry_antrea.csv
```diff
@@ -54,3 +54,5 @@ ElementID,Name,Abstract Data Type,Data Type Semantics,Status,Description,Units,R
 152,flowEndSecondsFromDestinationNode,unsigned32,,current,,,,,,,,56506,
 153,egressName,string,,current,,,,,,,,56506,
 154,egressIP,string,,current,,,,,,,,56506,
+155,l7FlowType,unsigned8,,current,Supported Actions(uint8 value): L7flowNotPresent(0) L7flowTypeHttp(1),,,,,,,56506,
```
antoninbas, line 57: I suggested `l7FlowType` as an example, but from a consistency standpoint with standard IPFIX IEs, I feel like `l7ProtocolName` is a better name

I also feel like using a short string (`"http"`, and maybe later `"grpc"`, `"ssh"`, etc) makes more sense than using an integer value, unless these integer values are standardized somewhere (I doubt it).

tushartathgur: I see, I got confused as it was mentioned enum, allow me to change it.

antoninbas (approved):
LGTM

heanlan (approved):
LGTM

PR #322, opened 2023-09-29
Improve handling of string IEs

antoninbas:
* Use the copy built-in to copy string bytes to buffer, instead of copying bytes one-by-one with a for loop. This is more efficient (confirmed with new Benchmark functions) and more correct (in case the string includes UTF-8 characters which use more than 1 byte).
* Increase max string length from 65,534 to 65,535.

### DIFF
--- a/pkg/entities/ie_test.go
+++ b/pkg/entities/ie_test.go
@@ -2,9 +2,11 @@ package entities
 
 import (
 	"net"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 var macAddress, _ = net.ParseMAC("aa:bb:cc:dd:ee:ff")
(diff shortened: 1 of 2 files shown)


antoninbas:
New benchmark results:
```
BenchmarkEncodeInfoElementValueToBuffShortString-12    	    2770	    435264 ns/op	       0 B/op	       0 allocs/op
BenchmarkEncodeInfoElementValueToBuffLongString-12     	    1341	    824074 ns/op	       0 B/op	       0 allocs/op
```
Without the use of `copy`:
```
BenchmarkEncodeInfoElementValueToBuffShortString-12    	    2154	    531415 ns/op	       0 B/op	       0 allocs/op
BenchmarkEncodeInfoElementValueToBuffLongString-12     	     152	   7948069 ns/op	       0 B/op	       0 allocs/op
```

pkg/entities/ie_test.go
```diff
@@ -72,3 +74,37 @@ func TestNewInfoElementWithValue(t *testing.T) {
...
+		index := 0
+		for j := 0; j < numCopies; j++ {
+			require.NoError(b, encodeInfoElementValueToBuff(element, buffer, index))
+			index += length
+		}
+	}
+}
+
+func BenchmarkEncodeInfoElementValueToBuffLongString(b *testing.B) {
+	// a long string has a max length of 65534
```
dreamtalen, line 96: Appears to be a typo, 65534->65535

dreamtalen (approved):
LGTM

dreamtalen (approved):
LGTM

heanlan (approved):
LGTM

PR #323, opened 2023-09-29
Avoid error logs in BenchmarkMultipleExportersToCollector

antoninbas:
* Gracefully shutdown exporter connections before stopping collector.
* Do not log an error for graceful client shutdown (EOF error).


heanlan (approved):
LGTM

About the previous integration test flakiness, I think there could be a bug in the throughput calculation logic. I'm checking on it.

dreamtalen (approved):
LGTM

PR #324, opened 2023-09-30
Fix aggregation bug for throughput common fields

antoninbas:
Just like for stats, throughput common fields should only be updated for the "latest" record (based on "flowEndSeconds").

This bug explains why the TestCollectorToIntermediate integration test has been flaky for a long time. In some cases, the record from the destination was processed before the record from the source (we have 2 aggregation workers, so records are not guaranteed to be processed in the order in which they are received). While the destination record has a larger "flowEndSeconds", the "throughput" and "reverseThroughput" common elements were overriden by values from the source record.


dreamtalen (approved):
LGTM

PR #325, opened 2023-10-02
Improvements in integration tests

antoninbas:
* Use more meaningful byte counts in test records (count and reverseCount have no reason to be the same).
* Generate test data packets automatically using a function, instead of having to generate packet data manually.


dreamtalen (approved):
LGTM

PR #328, opened 2023-10-02
Update CHANGELOG and VERSION for v0.8.0

dreamtalen:


heanlan (commented):
We need one more commit to update the VERSION, like https://github.com/vmware/go-ipfix/pull/319

dreamtalen:
> We need one more commit to update the VERSION, like #319

Done. I planned to open a separate PR to update the VERSION in the beginning

heanlan (approved):
LGTM

PR #329, opened 2023-10-03
Add new Go Benchmarks

antoninbas:
A few new benchmarks, which can be useful when validating improvements to the implementation.

### DIFF
--- a/pkg/entities/set_test.go
+++ b/pkg/entities/set_test.go
@@ -11,30 +11,61 @@ import (
 
 const (
 	testTemplateID = uint16(256)
+
+	testIPv4Addr1 = "10.0.0.1"
+	testIPv4Addr2 = "10.0.0.2"
+	testIPv6Addr1 = "2001:0:3238:dfe1:63::fefb"
+	testIPv6Addr2 = "2001:0:3238:dfe1:63::fefc"
 )
 
-func testAddRecordIPv4Addresses(t testing.TB) {
+func testAddRecordIPAddresses(t testing.TB, isIPv6 bool) {
+	var sourceIE, destinationIE *InfoElement
+	if isIPv6 {
+		sourceIE = NewInfoElement("sourceIPv6Address", 27, 19, 0, 16)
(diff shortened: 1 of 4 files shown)


dreamtalen (dismissed):
LGTM.
While trying this benchmark locally, I found the IPv4 AddRecord and AggregateMsg benchmarks are approximately 20% slower than the IPv6 one.
Have you also seen similar results Antonin?
```
BenchmarkAddRecordIPv4Addresses-10                     	  335524	      3345 ns/op
BenchmarkAddRecordIPv6Addresses-10                     	  404529	      2855 ns/op
```
```
BenchmarkAggregateMsgByFlowKey/ipv4-10         	   58531	     20586 ns/op
BenchmarkAggregateMsgByFlowKey/ipv6-10         	   82375	     15100 ns/op
```

antoninbas:
@dreamtalen For `BenchmarkAddRecord`, I have also observed that. It's because the IPv4 benchmarks is not the same as the IPv6 one (more checks). Although I will update it so that they do the same. It doesn't really make sense to have them be different.

For `BenchmarkAggregateMsgByFlowKey`, the IPv4 one is faster for me, as expected.
```
BenchmarkAggregateMsgByFlowKey/ipv4-12         	  181294	      6341 ns/op	     696 B/op	      12 allocs/op
BenchmarkAggregateMsgByFlowKey/ipv6-12         	  164910	      7327 ns/op	     792 B/op	      13 allocs/op
```

antoninbas:
@dreamtalen I unified the test code, PTAL

For `BenchmarkAggregateMsgByFlowKey`, you could give it another try (I recommend adding `-test.benchmem` as well). Your computer may have been running something else at the time.

dreamtalen (approved):
LGTM

dreamtalen:
> @dreamtalen I unified the test code, PTAL
> 
> For `BenchmarkAggregateMsgByFlowKey`, you could give it another try (I recommend adding `-test.benchmem` as well). Your computer may have been running something else at the time.

Thank you. As expected, the IPv4 one now performs slightly better in both benchmark tests.

PR #330, opened 2023-10-04
Create copies of net.IP and net.HardwareAddr when decoding IE values

antoninbas:
When using the provided slice (provided by the collector) directly, the garbage collector will not be able to release the underlying array while references to the net objects exist. The underlying array is allocated when reading the message from the TCP connection and can be 100s of bytes in size. The collector will typically run aggregation and keep references to the data record and its IEs for some time. During that time, we cannot release the packet buffer, while new packets keep coming in. To avoid the issue, we create a new slice, with a new underlying array, and the packet buffer can be releases as soon as all IEs are instantiated.

### DIFF
--- a/pkg/collector/process_test.go
+++ b/pkg/collector/process_test.go
@@ -19,12 +19,14 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"net"
+	"runtime"
 	"sync"
 	"testing"
 	"time"
 
 	"github.com/pion/dtls/v2"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/vmware/go-ipfix/pkg/entities"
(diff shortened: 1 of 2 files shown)


antoninbas:
Before the change:
```
=== RUN   TestTCPCollectingProcess_ReceiveDataRecordsMemoryUsage
I1004 13:18:40.576899   58679 tcp.go:39] Start TCP collecting process on 127.0.0.1:56807
    process_test.go:168: Data packet length: 33
I1004 13:18:40.713559   58679 process.go:121] stopping the collecting process
E1004 13:18:40.713533   58679 tcp.go:79] "Error when retrieving message length" err="read tcp 127.0.0.1:56807->127.0.0.1:56809: use of closed network connection"
    process_test.go:195: Live objects: 4432
    process_test.go:196: Bytes of allocated heap objects: 514984
```

After the change:
```
=== RUN   TestTCPCollectingProcess_ReceiveDataRecordsMemoryUsage
I1004 13:40:25.811470   62567 tcp.go:39] Start TCP collecting process on 127.0.0.1:56963
    process_test.go:169: Data packet length: 33
I1004 13:40:25.941870   62567 process.go:121] stopping the collecting process
    process_test.go:197: Live objects: 4427
    process_test.go:198: Bytes of allocated heap objects: 481208
```

The difference is roughly 33KB (1,000 packets x 33B per packet).
In a real deployment with 100s connections per second and data records that can be as large as 1000B (think Pods with labels), the difference can be quite significant.

pkg/collector/process_test.go
```diff
@@ -144,6 +146,60 @@ func TestTCPCollectingProcess_ReceiveDataRecord(t *testing.T) {
...
+	waitForCollectorReady(t, cp)
+
+	var conn net.Conn
+	collectorAddr := cp.GetAddress()
+	conn, err = net.Dial(collectorAddr.Network(), collectorAddr.String())
+	require.NoError(t, err)
+
+	const numRecords = 1000
+
+	ies := make([]entities.InfoElementWithValue, 0, numRecords)
```
heanlan, line 169: Apology if I didn't fully understand the test. Is there a need to collect and validate the `sourceIPv4Address ` ie values?

antoninbas: I'll write a comment. If these are not collected / used after calling `runtime.ReadMemStats`, everything will be GC'd.

heanlan (approved):
LGTM

PR #331, opened 2023-10-05
Improve memory performance of collector

antoninbas:
Avoid unnecessary slice creation when adding a record to a set. We implement this change for both data record and template records, but in practice most of the gains will come from data records.

Memory allocations (in bytes) are reduced by around 15% for BenchmarkCollector.

### DIFF
--- a/pkg/collector/process.go
+++ b/pkg/collector/process.go
@@ -280,7 +280,7 @@ func (cp *CollectingProcess) decodeTemplateSet(templateBuffer *bytes.Buffer, obs
 			return nil, err
 		}
 	}
-	err := templateSet.AddRecord(elementsWithValue, templateID)
+	err := templateSet.AddRecordV2(elementsWithValue, templateID)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


antoninbas:
Before change:
```
goos: darwin
goarch: amd64
pkg: github.com/vmware/go-ipfix/pkg/test
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkCollector/ipv4-12         	     304	   3809314 ns/op	 1832212 B/op	   45001 allocs/op
BenchmarkCollector/ipv6-12         	     304	   3852429 ns/op	 1922795 B/op	   45000 allocs/op
PASS
```

After change:
```
goos: darwin
goarch: amd64
pkg: github.com/vmware/go-ipfix/pkg/test
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkCollector/ipv4-12         	     306	   3785402 ns/op	 1512218 B/op	   44001 allocs/op
BenchmarkCollector/ipv6-12         	     310	   3828923 ns/op	 1602773 B/op	   44000 allocs/op
PASS
```

antoninbas:
@heanlan @dreamtalen let me know if you think this is not a correct change (functionality-wise)

dreamtalen (approved):
LGTM

heanlan (approved):
LGTM

Do we still need to keep the original `AddRecord`?

antoninbas:
> LGTM
> 
> Do we still need to keep the original `AddRecord`?

Since it's a public API of the library, and there was no strong reason to remove it, I just kept it

PR #333, opened 2023-10-22
Rename l7ProtocolName to appProtocol

tushartathgur:
Keeping the more standardized name for Layer 7 Protocol field

### DIFF
--- a/pkg/registry/registry_antrea.csv
+++ b/pkg/registry/registry_antrea.csv
@@ -54,5 +54,5 @@ ElementID,Name,Abstract Data Type,Data Type Semantics,Status,Description,Units,R
 152,flowEndSecondsFromDestinationNode,dateTimeSeconds,,current,,,,,,,,56506,
 153,egressName,string,,current,,,,,,,,56506,
 154,egressIP,string,,current,,,,,,,,56506,
-155,l7ProtocolName,string,,current,,,,,,,,56506,
+155,appProtocolName,string,,current,,,,,,,,56506,
 156,httpVals,string,,current,,,,,,,,56506,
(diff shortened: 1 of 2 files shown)


pkg/registry/registry_antrea.csv
```diff
@@ -54,5 +54,5 @@ ElementID,Name,Abstract Data Type,Data Type Semantics,Status,Description,Units,R
 152,flowEndSecondsFromDestinationNode,dateTimeSeconds,,current,,,,,,,,56506,
 153,egressName,string,,current,,,,,,,,56506,
 154,egressIP,string,,current,,,,,,,,56506,
-155,l7ProtocolName,string,,current,,,,,,,,56506,
+155,appProtocol,string,,current,,,,,,,,56506,
```
antoninbas, line 57: what about `appProtocolName`? I see that in standard IANA IE names, the `Name` suffix can be used explicitly to differentiate between string names and string ids (`vrfName`, `observationDomainName`, `observationDomainId`, ...)

tushartathgur: appProtocol is more streamlined with https://github.com/kubernetes/api/blob/fe172d7dd4b6501923938705cfb69ee0b907c32d/core/v1/types.go#L5058 
may be we can use appProtocolID when we will introduce application layer protocol ID in the infix tables?

antoninbas: I don't know if the K8s api is a good reference for this. It tends to use names exclusively, whereas IPFIX is a mix of names & ids

tushartathgur: changed!

antoninbas (approved):

PR #334, opened 2023-11-07
Update Flow aggregate field httpVals

tushartathgur:
if exporter sends httpval1 in first export and sends httpvals2 in second we should append both before sending further.
Also the function should take care of deduplicating same TxID items.

### DIFF
--- a/pkg/intermediate/aggregate.go
+++ b/pkg/intermediate/aggregate.go
@@ -16,6 +16,7 @@ package intermediate
 
 import (
 	"container/heap"
+	"encoding/json"
 	"fmt"
 	"net"
 	"strings"
@@ -542,6 +543,16 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
 					incomingVal := ieWithValue.GetStringValue()
 					existingIeWithValue.SetStringValue(incomingVal)
 				}
+			case "httpVals":
+				incomingVal := ieWithValue.GetStringValue()
+				existingVal := existingIeWithValue.GetStringValue()
+				updatedHttpVals, err := fillHttpVals(incomingVal, existingVal)
+				if err != nil {
+					klog.Errorf("httpVals could not be updated, err: %v", err)
+					existingIeWithValue.SetStringValue(incomingVal)
+				} else {
+					existingIeWithValue.SetStringValue(updatedHttpVals)
+				}
 			default:
 				klog.Errorf("Fields with name %v is not supported in aggregation fields list.", element)
 			}
@@ -983,4 +994,28 @@ func isCorrelationRequired(flowType uint8, record entities.Record) bool {
 		return true
 	}
 	return false
+}
+
+func fillHttpVals(incomingHttpVals, existingHttpVals string) (string, error) {
+	incomingHttpValsJson := make(map[int32]string)
+	existingHttpValsJson := make(map[int32]string)
+
+	if incomingHttpVals != "" {
+		if err := json.Unmarshal([]byte(incomingHttpVals), &incomingHttpValsJson); err != nil {
+			return "", fmt.Errorf("error parsing JSON: %v", err)
+		}
+	}
+	if existingHttpVals != "" {
+		if err := json.Unmarshal([]byte(existingHttpVals), &existingHttpValsJson); err != nil {
+			return "", fmt.Errorf("error parsing JSON: %v", err)
+		}
+	}
+	for key, value := range existingHttpValsJson {
+		incomingHttpValsJson[key] = value
+	}
+	updatedHttpVals, err := json.Marshal(incomingHttpValsJson)
+	if err != nil {
+		return "", fmt.Errorf("error converting JSON to string: %v", err)
+	}
+	return string(updatedHttpVals), nil
 }
--- a/pkg/intermediate/aggregate_test.go
+++ b/pkg/intermediate/aggregate_test.go
@@ -47,6 +47,7 @@ var (
 		"flowEndSeconds",
 		"flowEndReason",
 		"tcpState",
+		"httpVals",
 	}
 	statsElementList = []string{
 		"packetTotalCount",
@@ -165,7 +166,7 @@ func createDataMsgForSrc(t testing.TB, isIPv6 bool, isIntraNode bool, isUpdatedR
 	ie6 := entities.NewStringInfoElement(entities.NewInfoElement("sourcePodName", 101, 13, registry.AntreaEnterpriseID, 65535), srcPod)
 	ie7 := entities.NewStringInfoElement(entities.NewInfoElement("destinationPodName", 103, 13, registry.AntreaEnterpriseID, 65535), dstPod)
 	ie9 := entities.NewUnsigned16InfoElement(entities.NewInfoElement("destinationServicePort", 107, 2, registry.AntreaEnterpriseID, 2), uint16(4739))
-	var ie1, ie2, ie8, ie10, ie11, ie12, ie13, ie14, ie15, ie16 entities.InfoElementWithValue
+	var ie1, ie2, ie8, ie10, ie11, ie12, ie13, ie14, ie15, ie16, ie17, ie18 entities.InfoElementWithValue
 	if !isIPv6 {
 		ie1 = entities.NewIPAddressInfoElement(entities.NewInfoElement("sourceIPv4Address", 8, 18, 0, 4), net.ParseIP("10.0.0.1").To4())
 		ie2 = entities.NewIPAddressInfoElement(entities.NewInfoElement("destinationIPv4Address", 12, 18, 0, 4), net.ParseIP("10.0.0.2").To4())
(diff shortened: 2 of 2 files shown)


pkg/intermediate/aggregate.go
```diff
@@ -984,3 +993,29 @@ func isCorrelationRequired(flowType uint8, record entities.Record) bool {
...
+	existingHttpValsJson := make(map[int32]string)
+
+	if incomingHttpVals != "" {
+		if err := json.Unmarshal([]byte(incomingHttpVals), &incomingHttpValsJson); err != nil {
+			return "", fmt.Errorf("Error parsing JSON: %v", err)
+		}
+	}
+	if existingHttpVals != "" {
+		if err := json.Unmarshal([]byte(existingHttpVals), &existingHttpValsJson); err != nil {
+			return "", fmt.Errorf("Error parsing JSON: %v", err)
```
heanlan, line 1008: error message should not be capitalized

pkg/intermediate/aggregate.go
```diff
@@ -984,3 +993,29 @@ func isCorrelationRequired(flowType uint8, record entities.Record) bool {
...
+		}
+	}
+	if len(existingHttpValsJson) > 0 {
+		for key, value := range existingHttpValsJson {
+			incomingHttpValsJson[key] = value
```
heanlan, line 1013: Sorry for my ignorance, I'm a little bit confused about what we are trying to do here, why we fill value of `existingHttpValsJson` to `incomingHttpValsJson` if we have already unmarshal the `incomingHttpVals` in L1001

heanlan (commented):
Could you also update the aggregate_test.go file to include `httpVals`?

pkg/intermediate/aggregate.go
tushartathgur: This step is for deduplicating any keys present in both jsons and adding new keys not present in the incomingHttpValsJson

pkg/intermediate/aggregate.go
tushartathgur: Thanks! let me change it

pkg/intermediate/aggregate.go
```diff
@@ -984,3 +993,29 @@ func isCorrelationRequired(flowType uint8, record entities.Record) bool {
 	}
 	return false
 }
+
+func fillHttpVals(incomingHttpVals, existingHttpVals string) (string, error) {
+	incomingHttpValsJson := make(map[int32]string)
```
yuntanghsu, line 998: Forgive me if it is a dumb question. Transaction iD should be the same for the same event in different nodes?

pkg/intermediate/aggregate.go
```diff
@@ -542,6 +543,14 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
 					incomingVal := ieWithValue.GetStringValue()
 					existingIeWithValue.SetStringValue(incomingVal)
 				}
+			case "httpVals":
+				incomingVal := ieWithValue.GetStringValue()
+				existingVal := existingIeWithValue.GetStringValue()
+				updatedHttpVals, err := fillHttpVals(incomingVal, existingVal)
+				if err != nil {
+					return fmt.Errorf("httpVals could not be updated: %v", err)
```
yuntanghsu, line 551: Do you think we should print the error log instead? Or we should reset fields we changed above. Otherwise the existing connection might have wrong info comes from the incoming connection.

pkg/intermediate/aggregate.go
```diff
@@ -984,3 +993,29 @@ func isCorrelationRequired(flowType uint8, record entities.Record) bool {
...
+		}
+	}
+	if len(existingHttpValsJson) > 0 {
+		for key, value := range existingHttpValsJson {
+			incomingHttpValsJson[key] = value
+		}
+	}
+	updatedHttpVals, err := json.Marshal(incomingHttpValsJson)
+	if err != nil {
+		return "", fmt.Errorf("Error converting JSON to string: %v", err)
```
yuntanghsu, line 1018:
```suggestion
		return "", fmt.Errorf("error converting JSON to string: %v", err)
```

pkg/intermediate/aggregate.go
tushartathgur: Yes transaction ID should be same

pkg/intermediate/aggregate.go
tushartathgur: That is a good idea, allow me to change it

pkg/intermediate/aggregate.go
antoninbas: I think that's a key point, and the commit message should probably mention it.
Given that the HTTP values from both Nodes should eventually be consistent, I wonder if there is real value in that "complex" aggregation logic, or if we should just use the "latest" record (like we do for some other fields) and simply override the aggregated value.

tushartathgur: I am not able to reproduce it in my local setup, but the thought behind this design is if multiple exporters send the httpvals with different or overlapping TxIDs, we should deduplicate it.
if we only take the latest record for aggregation, we may not get all the TxIDs in a aggregated output given the aggregation interval is bigger than the exporter interval.

yuntanghsu: I think we won't be able to use the "latest" record in the current design.
The records from our Flow Exporter only contain partial httpVals as this field is reset after the connection is exported from the FE.
If we want to use the latest record in the FA, we might need to do the aggregation in the FE instead, which could end up increasing the packet size, transferred from FE to FA, by time?

antoninbas: Thanks @yuntanghsu, that makes sense.

pkg/intermediate/aggregate.go
```diff
@@ -984,3 +995,29 @@ func isCorrelationRequired(flowType uint8, record entities.Record) bool {
...
+		if err := json.Unmarshal([]byte(incomingHttpVals), &incomingHttpValsJson); err != nil {
+			return "", fmt.Errorf("error parsing JSON: %v", err)
+		}
+	}
+	if existingHttpVals != "" {
+		if err := json.Unmarshal([]byte(existingHttpVals), &existingHttpValsJson); err != nil {
+			return "", fmt.Errorf("error parsing JSON: %v", err)
+		}
+	}
+	if len(existingHttpValsJson) > 0 {
```
antoninbas, line 1013: the check is not useful: if the length is 0, the the loop will simply not execute

pkg/intermediate/aggregate.go
```diff
@@ -542,6 +543,16 @@ func (a *AggregationProcess) aggregateRecords(incomingRecord, existingRecord ent
 					incomingVal := ieWithValue.GetStringValue()
 					existingIeWithValue.SetStringValue(incomingVal)
 				}
+			case "httpVals":
+				incomingVal := ieWithValue.GetStringValue()
+				existingVal := existingIeWithValue.GetStringValue()
+				updatedHttpVals, err := fillHttpVals(incomingVal, existingVal)
+				if err != nil {
+					klog.Errorf("httpVals could not be updated from jsons, err: %v", err)
```
antoninbas, line 551:
```suggestion
					klog.Errorf("httpVals could not be updated: %v", err)
```

pkg/intermediate/aggregate.go
```diff
@@ -984,3 +995,29 @@ func isCorrelationRequired(flowType uint8, record entities.Record) bool {
 	}
 	return false
 }
+
+func fillHttpVals(incomingHttpVals, existingHttpVals string) (string, error) {
```
yuntanghsu, line 999: Can we add a unit test for this function to make sure we correlate two HttpVals correctly?

pkg/intermediate/aggregate_test.go
```diff
@@ -1127,3 +1136,35 @@ func runAggregationAndCheckResult(t *testing.T, ap *AggregationProcess, srcRecor
...
+			existingHttpVals: "{\"0\":\"{hostname:10.10.0.1,url:/public/,http_user_agent:curl/7.74.0,http_content_type:text/html,http_method:GET,protocol:HTTP/1.1,status:200,length:153}\"}",
+			updatedHttpVals:  "{\"0\":\"{hostname:10.10.0.1,url:/public/,http_user_agent:curl/7.74.0,http_content_type:text/html,http_method:GET,protocol:HTTP/1.1,status:200,length:153}\",\"1\":\"{hostname:10.10.0.1,url:/public/,http_user_agent:curl/7.74.0,http_content_type:text/html,http_method:GET,protocol:HTTP/1.1,status:200,length:153}\"}",
+		}, {
+			name:             "Existing httpVals empty",
+			incomingHttpVals: "{\"1\":\"{hostname:10.10.0.1,url:/public/,http_user_agent:curl/7.74.0,http_content_type:text/html,http_method:GET,protocol:HTTP/1.1,status:200,length:153}\"}",
+			existingHttpVals: "",
+			updatedHttpVals:  "{\"1\":\"{hostname:10.10.0.1,url:/public/,http_user_agent:curl/7.74.0,http_content_type:text/html,http_method:GET,protocol:HTTP/1.1,status:200,length:153}\"}",
+		}, {
+			name:             "overlapping httpVals empty",
```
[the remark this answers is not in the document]
tushartathgur: Allow me to change the name , "empty" just got copied from the last test case, I will change it too

PR #335, opened 2023-11-14
Adding make fmt support in go-ipfix

tushartathgur:
go-ipfix does not contain gofmt support, adding the support for it in Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ PROTOC_GEN_GO_VERSION  := v1.28.1
 GOLANGCI_LINT_VERSION  := v1.54.2
 GOLANGCI_LINT_BINDIR   := .golangci-bin
 GOLANGCI_LINT_BIN      := $(GOLANGCI_LINT_BINDIR)/$(GOLANGCI_LINT_VERSION)/golangci-lint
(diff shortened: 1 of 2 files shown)


antoninbas (approved):
LGTM
I have no issue with including this make target, even though it feels like `golangci-lint` should already handle this as it includes the `gofmt` linter.
We do have this target in Antrea, but I know I am not using it, and it was introduced prior to us using `golangci-lint`.

tushartathgur:
> LGTM I have no issue with including this make target, even though it feels like `golangci-lint` should already handle this as it includes the `gofmt` linter. We do have this target in Antrea, but I know I am not using it, and it was introduced prior to us using `golangci-lint`.

I tried using make golangci, but it throws error like "File is not `gofmt`-ed with `-s", so I added make fmt

antoninbas:
@tushartathgur that's expected.
`make golangci` reports errors, while `make golangci-fix` fixes them when possible

tushartathgur:
> @tushartathgur that's expected. `make golangci` reports errors, while `make golangci-fix` fixes them when possible

makes sense, I just tried that, apologies for overlooking it

PR #338, opened 2023-12-12
Improve IPFIX collector

yuntanghsu:
In this PR:
1. We disable printing records whenever we receive it. Instead, we store records in a string array.
2. Add http listener and handler to receive request to return or reset records.

The reason we do this change is because we want to reduce the time we use to retrieve logs.
Previously, using `kubectl logs` will take lots of time to get the logs. And the time depends on the number of records inside the IPFIX collector Pod. The time range is ~0.8s to 4s. After above changes, the time is reduced to ~80ms.

Functionality has been tested in this PR: https://github.com/antrea-io/antrea/actions/runs/7201229562/job/19617317869?pr=5770

### DIFF
--- a/cmd/collector/collector.go
+++ b/cmd/collector/collector.go
@@ -16,11 +16,15 @@ package main
 
 import (
 	"bytes"
+	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
+	"net/http"
 	"os"
 	"os/signal"
 	"strconv"
+	"sync"
 	"syscall"
 	"time"
 
@@ -42,6 +46,8 @@ var (
 	IPFIXAddr      string
 	IPFIXPort      uint16
 	IPFIXTransport string
+	flowRecords    []string
+	mutex          sync.Mutex
 )
 
 func initLoggingToFile(fs *pflag.FlagSet) {
@@ -60,7 +66,7 @@ func addIPFIXFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&IPFIXTransport, "ipfix.transport", "tcp", "IPFIX collector transport layer")
 }
 
-func printIPFIXMessage(msg *entities.Message) {
+func addIPFIXMessage(msg *entities.Message) {
 	var buf bytes.Buffer
 	fmt.Fprint(&buf, "\nIPFIX-HDR:\n")
 	fmt.Fprintf(&buf, "  version: %v,  Message Length: %v\n", msg.GetVersion(), msg.GetMessageLen())
@@ -126,7 +132,9 @@ func printIPFIXMessage(msg *entities.Message) {
 			}
 		}
 	}
-	klog.Infof(buf.String())
+	mutex.Lock()
+	defer mutex.Unlock()
+	flowRecords = append(flowRecords, buf.String())
 }
 
 func signalHandler(stopCh chan struct{}, messageReceived chan *entities.Message) {
@@ -136,7 +144,7 @@ func signalHandler(stopCh chan struct{}, messageReceived chan *entities.Message)
 	for {
 		select {
 		case msg := <-messageReceived:
-			printIPFIXMessage(msg)
+			addIPFIXMessage(msg)
 		case <-signalCh:
 			close(stopCh)
 			return
@@ -173,6 +181,25 @@ func run() error {
 		}
 	}()
 
+	var server *http.Server
+	// Start the HTTP server in a separate goroutine
+	go func() {
+		mux := http.NewServeMux()
+		mux.HandleFunc("/records", flowRecordHandler)
+		mux.HandleFunc("/reset", resetRecordHandler)
+
+		server = &http.Server{
+			Addr:              ":8080",
+			Handler:           mux,
+			ReadHeaderTimeout: 5 * time.Second,
+		}
+
+		klog.InfoS("Server listening on :8080")
+		if err := server.ListenAndServe(); err != http.ErrServerClosed {
+			klog.Fatalf("ListenAndServe error: %v", err)
+		}
+	}()
+
 	stopCh := make(chan struct{})
 	go signalHandler(stopCh, messageReceived)
 
@@ -180,6 +207,15 @@ func run() error {
 	// Stop the collector process
 	cp.Stop()
 	klog.Info("Stopping IPFIX collector")
+	// Create a context with a 5-second timeout
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	// Attempt to gracefully shut down the server
+	if err := server.Shutdown(shutdownCtx); err != nil {
+		klog.ErrorS(err, "Error during server shutdown")
+	} else {
+		klog.InfoS("Server gracefully stopped")
+	}
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


cmd/collector/collector.go
```diff
@@ -173,13 +181,42 @@ func run() error {
 		}
 	}()
 
+	// Create a new context with a cancel function
+	_, cancel := context.WithCancel(context.Background())
+	var server *http.Server
+	// Start the HTTP server in a separate goroutine
+	go func() {
+		mux := http.NewServeMux()
+		mux.HandleFunc("/record", flowRecordHandler)
```
antoninbas, line 190: s/record/records for the endpoint name

cmd/collector/collector.go
```diff
@@ -202,6 +239,38 @@ func newCollectorCommand() *cobra.Command {
 	return cmd
 }
 
+func getFlowRecords() []string {
+	return flowRecords
+}
+
+func flowRecordHandler(w http.ResponseWriter, r *http.Request) {
```
antoninbas, line 246: you should require the HTTP method to be GET in this handler:
```go
	if r.Method == "GET" {
		// ...
	} else {
		http.Error(w, "Invalid request method", http.StatusMethodNotAllowed)
	}
```

cmd/collector/collector.go
```diff
@@ -202,6 +239,38 @@ func newCollectorCommand() *cobra.Command {
...
+		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
+		return
+	}
+	// Set response headers
+	w.Header().Set("Content-Type", "application/json")
+	// Write JSON response
+	w.Write(jsonData)
+}
+
+func resetRecordHandler(w http.ResponseWriter, r *http.Request) {
```
antoninbas, line 265: same as above, but you should request the HTTP method to be POST in this handler

cmd/collector/collector.go
```diff
@@ -202,6 +239,38 @@ func newCollectorCommand() *cobra.Command {
 	return cmd
 }
 
+func getFlowRecords() []string {
```
antoninbas, line 242: I think this method just hurts readability (it's a basic accessor, and the caller has to acquire a low anyway). I'd suggest removing it. If the method was making a copy of the slice, it would be a different story.

cmd/collector/collector.go
```diff
@@ -173,13 +181,42 @@ func run() error {
...
+	// Create a new context with a cancel function
+	_, cancel := context.WithCancel(context.Background())
+	var server *http.Server
+	// Start the HTTP server in a separate goroutine
+	go func() {
+		mux := http.NewServeMux()
+		mux.HandleFunc("/record", flowRecordHandler)
+		mux.HandleFunc("/reset", resetRecordHandler)
+
+		server = &http.Server{
```
antoninbas, line 193: you should just set `ReadHeaderTimeout` to avoid the gosec warning

cmd/collector/collector.go
```diff
@@ -173,13 +181,42 @@ func run() error {
 		}
 	}()
 
+	// Create a new context with a cancel function
+	_, cancel := context.WithCancel(context.Background())
```
antoninbas, line 185: you do not use the created context, that's definitely not right
if you do not need a cancellable context, don't create one

cmd/collector/collector.go
yuntanghsu: Updated. Thanks.

cmd/collector/collector.go
yuntanghsu: Fixed.

cmd/collector/collector.go
yuntanghsu: Why should this method be POST?
Do you think we still need this Handler. After disabling printing the log, the retrieve time is decreased from ~4s to ~80ms.

cmd/collector/collector.go
yuntanghsu: Removed.

cmd/collector/collector.go
yuntanghsu: Fixed. Thanks.

cmd/collector/collector.go
yuntanghsu: Removed.

cmd/collector/collector.go
```diff
@@ -60,7 +66,9 @@ func addIPFIXFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&IPFIXTransport, "ipfix.transport", "tcp", "IPFIX collector transport layer")
 }
 
-func printIPFIXMessage(msg *entities.Message) {
+func addIPFIXMessage(msg *entities.Message) {
+	mutex.Lock()
+	defer mutex.Unlock()
```
antoninbas, line 71: it's much better if you only lock the mutex when you need it, i.e. for the last statement in this function

cmd/collector/collector.go
```diff
@@ -173,13 +181,38 @@ func run() error {
...
+
 	stopCh := make(chan struct{})
 	go signalHandler(stopCh, messageReceived)
 
 	<-stopCh
 	// Stop the collector process
 	cp.Stop()
 	klog.Info("Stopping IPFIX collector")
+	// Attempt to gracefully shut down the server
+	if err := server.Shutdown(context.Background()); err != nil {
```
antoninbas, line 211: here you probably want a context with a 5s timeout, just for the shutdown operation

cmd/collector/collector.go
```diff
@@ -173,13 +181,38 @@ func run() error {
...
+
+		server = &http.Server{
+			Addr:              ":8080",
+			Handler:           mux,
+			ReadHeaderTimeout: 5 * time.Second,
+		}
+
+		klog.InfoS("Server listening on :8080")
+		if err := server.ListenAndServe(); err != http.ErrServerClosed {
+			klog.ErrorS(err, "ListenAndServe error")
```
antoninbas, line 199: I think maybe exiting would be appropriate here

cmd/collector/collector.go
yuntanghsu: Updated.

cmd/collector/collector.go
yuntanghsu: Added. Thanks.

cmd/collector/collector.go
yuntanghsu: Fixed. I changed it to use Fatal instead.

cmd/collector/collector.go
```diff
@@ -173,13 +181,41 @@ func run() error {
...
+
+		server = &http.Server{
+			Addr:              ":8080",
+			Handler:           mux,
+			ReadHeaderTimeout: 5 * time.Second,
+		}
+
+		klog.InfoS("Server listening on :8080")
+		if err := server.ListenAndServe(); err != http.ErrServerClosed {
+			klog.Fatal(err, "ListenAndServe error")
```
antoninbas, line 199: `klog.Fatal` is not really mean to be used like this IIRC. This won't format very nicely. See https://go.dev/play/p/d7u96VgMq46

yuntanghsu: Changed it to Fatalf instead. Thanks for the suggestion.

antoninbas (approved):
LGTM

antoninbas:
I don't remember how tagging works for the collector image. @heanlan @dreamtalen will we need a new release of go-ipfix?

cmd/collector/collector.go
yuntanghsu: @antoninbas 
I think this concern hasn't been solved?

antoninbas: sorry I didn't see the question before
It should be POST because it does a mutation the server. That's an HTTP convention. You can lookup which HTTP verbs are most appropriate for different type of APIs.

> Do you think we still need this Handler. After disabling printing the log, the retrieve time is decreased from ~4s to ~80ms.

I thought you wrote 0.8s in the PR description, which would be 800ms?

Regardless, let's add the reset method, it's easy to do and if we don't end up using it in Antrea tests, so be it.
It's easier than opening a new PR in this repo later and publishing a new collector image tag...

yuntanghsu: 0.8s to 4s will be the range when we use "kubectl logs" to get the logs.
Under the current changes, this time will be approximately 80ms regardless of the number of records inside the IPFIX collector.

heanlan:
> will we need a new release of go-ipfix?

Yes, we need a new release

pkg/intermediate/aggregate.go
```diff
@@ -366,6 +366,16 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
 				}
 			}
 		} else {
+			// For inter-node traffic with deny np or drop anp, we may receive the record from the conntrack table first.
```
antoninbas, line 369: this should be in a different PR, it's unrelated to the collector changes and seems like a bug fix to me

antoninbas (approved):

PR #339, opened 2023-12-14
Bug fix for inter-node traffic

yuntanghsu:
In this PR, we do:
1. Set ReadyToSend to true for flows don’t need correlation.
2. Set areCorrelatedFieldsFilled to true for flows don’t belong to inter-node traffic. For flows need to do correlation, its areCorrelatedFieldsFilled will be set to true once the correlation job is finished.
3. Ignore or overwrite the correlationRequired record if any non-correlationRequired record with the same flowKey comes before or after the correlationRequired record.

### DIFF
--- a/pkg/intermediate/aggregate.go
+++ b/pkg/intermediate/aggregate.go
@@ -340,9 +340,22 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
 		klog.Warning("FlowType does not exist in current record.")
 	}
 	correlationRequired := isCorrelationRequired(flowType, record)
-
-	currTime := time.Now()
 	aggregationRecord, exist := a.flowKeyRecordMap[*flowKey]
+	if exist {
+		if flowTypeIE, _, exist := aggregationRecord.Record.GetInfoElementWithValue("flowType"); exist {
+			flowType = flowTypeIE.GetUnsigned8Value()
+		} else {
+			klog.Warning("FlowType does not exist in current record.")
+		}
+		prevCorrelationRequired := isCorrelationRequired(flowType, aggregationRecord.Record)
+		if prevCorrelationRequired && !correlationRequired {
+			delete(a.flowKeyRecordMap, *flowKey)
+			exist = false
+		} else if !prevCorrelationRequired && correlationRequired {
+			return nil
+		}
+	}
+	currTime := time.Now()
 	if exist {
 		if correlationRequired {
 			// Do correlation of records if record belongs to inter-node flow and
@@ -411,14 +424,9 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
 
 		if !correlationRequired {
 			aggregationRecord.ReadyToSend = true
-			// If no correlation is required for an Inter-Node record, K8s metadata is
-			// expected to be not completely filled. For Intra-Node flows and ToExternal
-			// flows, areCorrelatedFieldsFilled is set to true by default.
-			if flowType == registry.FlowTypeInterNode {
-				aggregationRecord.areCorrelatedFieldsFilled = false
-			} else {
-				aggregationRecord.areCorrelatedFieldsFilled = true
-			}
+		}
+		if flowType != registry.FlowTypeInterNode {
+			aggregationRecord.areCorrelatedFieldsFilled = true
 		}
 		aggregationRecord.areExternalFieldsFilled = false
 		// Push the record to the priority queue.


pkg/intermediate/aggregate.go
```diff
@@ -366,6 +366,16 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
 				}
 			}
 		} else {
+			// For inter-node traffic with deny np or drop anp, we may receive the record from the conntrack table first.
+			// In this case, the record from the conntrack table will be considered to do correlation job and the ReadyToSend
+			// is false until we receive another record from PacketIn from another node. However, the record From PacketIN
+			// is not correlationRequired, so we won't send this kind of connection due to this conflict.
+			aggregationRecord.ReadyToSend = true
```
antoninbas, line 373: I see similar logic later in the function, can we unify?

pkg/intermediate/aggregate.go
```diff
@@ -366,6 +366,16 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
 				}
 			}
 		} else {
+			// For inter-node traffic with deny np or drop anp, we may receive the record from the conntrack table first.
+			// In this case, the record from the conntrack table will be considered to do correlation job and the ReadyToSend
+			// is false until we receive another record from PacketIn from another node. However, the record From PacketIN
+			// is not correlationRequired, so we won't send this kind of connection due to this conflict.
```
antoninbas, line 372: I don't fully understand what's going on and I'll probably defer to @heanlan and @dreamtalen, but I do not like that the comment is so specific to the Antrea FE implementation, with mentions of conntrack and PacketIn. While I know that this code is pretty specific to Antrea, this seems *too* specific.

antoninbas (commented):
can you rebase main?

pkg/intermediate/aggregate.go
yuntanghsu: Yes. I've moved it to the bottom and set ReadyToSend and areCorrelatedFieldsFilled to true if the flow is not correlationRequired.

pkg/intermediate/aggregate.go
yuntanghsu: I've updated the comment. Thanks.

pkg/intermediate/aggregate.go
```diff
@@ -408,17 +408,10 @@ func (a *AggregationProcess) addOrUpdateRecordInMap(flowKey *FlowKey, record ent
...
-			// flows, areCorrelatedFieldsFilled is set to true by default.
-			if flowType == registry.FlowTypeInterNode {
-				aggregationRecord.areCorrelatedFieldsFilled = false
-			} else {
-				aggregationRecord.areCorrelatedFieldsFilled = true
-			}
+		// For intra-node and external traffic, areCorrelatedFieldsFilled is always true.
+		// For inter-node with allow np/anp action, its areExternalFieldsFilled will be set to true once the correlation job is finished.
+		if flowType != registry.FlowTypeInterNode {
+			aggregationRecord.areCorrelatedFieldsFilled = true
```
heanlan, line 414: It seems like `areCorrelatedFieldsFilled` is only being used in UT? If that's the case, do we still want to keep it? maybe I missed something

heanlan (commented):
I don't fully understand the code changes. Based on my understanding of the PR description, the conntrack records without deny NP metadata should not be sent out by FA, right? If so, their `correlationRequired` is true, and they will be marked as `ReadyToSend` in your code, which contradicts with the PR purpose?

```golang
	// For flows that do not need correlation, ReadyToSend should always be true.
	if !correlationRequired {
		aggregationRecord.ReadyToSend = true
	}
```

pkg/intermediate/aggregate.go
yuntanghsu: It will be set to true in line#355. And areCorrelatedFieldsFilled will be used in the flowaggregator.go in Antrea repo as well.

yuntanghsu:
> I don't fully understand the code changes. Based on my understanding of the PR description, the conntrack records without deny NP metadata should not be sent out by FA, right? However, their correlationRequired are true, and they will be marked as ReadyToSend in your code, which contradicts with the PR purpose?

No, for deny NP/or drop ANP, we will receive two records. (one is from PacketIn and the other one is from the conntrack table)
We should set the ReadyToSend to true whenever we receive the record from PacketIn.

heanlan (dismissed):

yuntanghsu:
I think I made a mistake.

For inter-node traffic with egress/ingress np with action drop, we will receive records from PacketIn and the conntrack table. For the ingress case, there is no issue as we will receive the records from both nodes and these two records are both correlationRequired.
For the egress case, the record from the conntrack table is correlationRequired and the record from PacketIn is not correlationRequired. If the record from the conntrack table arrive at the FA first, then the record will need to do correlation at FA. The ReadyToSend will be true forever as it keeps waiting to do correlation.


For the egress case, we can overwrite the existing record if it is from the conntrack table. But I think it would be enough to change the order where we append expired records before exporting them from the exporter. And the chance that a record from the conntrack will be exported earlier than a record from PacketIn is very low.

For these reasons, I think we can close this PR as these changes might not be really helpful.
@dreamtalen @heanlan What do you think?

antoninbas:
@yuntanghsu I want to confirm my understanding with you, because I feel like I am missing something.
For inter-Node flows with a drop policy, my understanding is as follows:
- egress drop (at source Node): we will only get a record from the source Node. That record is "generated" by a PacketIn event. This connection should never be committed to conntrack in the Antrea ct_zone (not on the source Node and of course not on the destination Node). If we see the connection in conntrack, this sounds like a bug to me.
- ingress drop (at destination Node): we will get 2 records, one from each Node. the record from the source Node comes from conntrack polling, the record from the destination Node is generated by a PacketIn event. The connection will never be committed to conntrack on the destination Node. The connection will be committed to conntrack on the source Node, but it will never make it to established state.

Let me know if theses statements don't match your understanding or observations.

yuntanghsu:
> egress drop (at source Node): we will only get a record from the source Node. That record is "generated" by a PacketIn event. This connection should never be committed to conntrack in the Antrea ct_zone (not on the source Node and of course not on the destination Node). If we see the connection in conntrack, this sounds like a bug to me

If the traffic is not a service traffic. That's correct. But I observed we will receive records from PacketIn and the conntrack table.

I1218 21:30:37.187849       1 deny_connections.go:115] "New deny connection added" connection={"ID":0,"Timeout":0,"StartTime":"2023-12-18T21:30:37.187791036Z","StopTime":"2023-12-18T21:30:37.187791036Z","LastExportTime":"2023-12-18T21:30:37.187791036Z","IsActive":true,"IsPresent":false,"ReadyToDelete":false,"Zone":0,"Mark":19,"StatusFlag":0,"Labels":null,"LabelsMask":null,"FlowKey":{"SourceAddress":"10.244.0.7","DestinationAddress":"10.244.1.8","Protocol":6,"SourcePort":57252,"DestinationPort":5201},"OriginalPackets":1,"OriginalBytes":60,"SourcePodNamespace":"testflowaggregator-0e73pgnx","SourcePodName":"perftest-a","DestinationPodNamespace":"","DestinationPodName":"","DestinationServicePortName":"testflowaggregator-0e73pgnx/perftest-e:","OriginalDestinationAddress":"10.96.176.70","OriginalDestinationPort":9999,"IngressNetworkPolicyName":"","IngressNetworkPolicyNamespace":"","IngressNetworkPolicyType":0,"IngressNetworkPolicyRuleName":"","IngressNetworkPolicyRuleAction":0,"EgressNetworkPolicyName":"test-flow-aggregator-anp-egress-drop","EgressNetworkPolicyNamespace":"testflowaggregator-0e73pgnx","EgressNetworkPolicyType":2,"EgressNetworkPolicyRuleName":"test-egress-rule-name","EgressNetworkPolicyRuleAction":2,"PrevPackets":0,"PrevBytes":0,"ReversePackets":0,"ReverseBytes":0,"PrevReversePackets":0,"PrevReverseBytes":0,"TCPState":"","PrevTCPState":"","FlowType":0,"EgressName":"","EgressIP":""}


I1218 21:30:38.106123       1 conntrack_connections.go:284] "New Antrea flow added" connection={"ID":2467408996,"Timeout":119,"StartTime":"2023-12-18T21:30:37.187206191Z","StopTime":"2023-12-18T21:30:38.097239785Z","LastExportTime":"2023-12-18T21:30:37.187206191Z","IsActive":true,"IsPresent":true,"ReadyToDelete":false,"Zone":65520,"Mark":19,"StatusFlag":424,"Labels":null,"LabelsMask":null,"FlowKey":{"SourceAddress":"10.244.0.7","DestinationAddress":"10.244.1.8","Protocol":6,"SourcePort":57252,"DestinationPort":5201},"OriginalPackets":1,"OriginalBytes":60,"SourcePodNamespace":"testflowaggregator-0e73pgnx","SourcePodName":"perftest-a","DestinationPodNamespace":"","DestinationPodName":"","DestinationServicePortName":"testflowaggregator-0e73pgnx/perftest-e:","OriginalDestinationAddress":"10.96.176.70","OriginalDestinationPort":9999,"IngressNetworkPolicyName":"","IngressNetworkPolicyNamespace":"","IngressNetworkPolicyType":0,"IngressNetworkPolicyRuleName":"","IngressNetworkPolicyRuleAction":0,"EgressNetworkPolicyName":"","EgressNetworkPolicyNamespace":"","EgressNetworkPolicyType":0,"EgressNetworkPolicyRuleName":"","EgressNetworkPolicyRuleAction":0,"PrevPackets":0,"PrevBytes":0,"ReversePackets":0,"ReverseBytes":0,"PrevReversePackets":0,"PrevReverseBytes":0,"TCPState":"SYN_SENT","PrevTCPState":"","FlowType":0,"EgressName":"","EgressIP":""}

antoninbas:
@yuntanghsu thanks, you are correct. For Service traffic, we will do DNAT (which commits to CT) before policy enforcement in Antrea. Although I wonder if this is the best approach, maybe I will bring it up at the next community meeting.

For ingress drop, I assume my understanding is correct?

With regards to your PR, I don't know what the best approach is. To me it seems that we should avoid exporting 2 separate records from the same Node in the FlowExporter in the first place. It is not right IMO to have the same connection in both connection stores. Maybe @heanlan and @dreamtalen have some inputs here.
**I didn't give it too much thought, but would it make sense to omit connections that were never established from the connection store?** For the ingress drop case, it would mean that we would only get 1 record (from the destination Node, triggered by PacketIn). It would make things a bit more symmetric in a way, and my assumption is that aggregation doesn't provide much value in this case.

yuntanghsu:
> For ingress drop, I assume my understanding is correct?
Yes, that's correct for non-service and service traffic.

I agree with you. I think the best solution approach is avoiding exporting the record in the conntrack in this case.
However, I found this is kind of difficult to do as we don't know that the record from the conntrack belongs to an allow connection or a deny connection until we see the record from PacketIn.

I think the other acceptable solution is changing the order where we append expired records before exporting them from our exporter. We should prioritize the record in the deny connection store and I think the chance that a record from the conntrack will be exported earlier than a record from PacketIn is very low.
I already made this change in https://github.com/antrea-io/antrea/pull/5770. The test is stable now.
Maybe we can release a new version of the IPFIX collector to unblock this PR https://github.com/antrea-io/antrea/pull/5770?

> I didn't give it too much thought, but would it make sense to omit connections that were never established from the connection store?
I think it's better to keep it? This kind of connection is still useful for detecting DoS attacks?

antoninbas:
@heanlan could you release vmware/go-ipfix 0.9.0, with Yun-Tang's improvements to the "reference" IPFIX collector

dreamtalen:
> @heanlan could you release vmware/go-ipfix 0.9.0, with Yun-Tang's improvements to the "reference" IPFIX collector

Anlan is on PTO this week, I can help to do a release

yuntanghsu:
Salvatore already release vmware/go-ipfix 0.8.2
https://github.com/vmware/go-ipfix/releases/tag/v0.8.2

antoninbas:
@dreamtalen can you do the changelog PR?

@yuntanghsu should we close this PR then?

dreamtalen:
> @dreamtalen can you do the changelog PR?

Yanjun has finished that :)

yuntanghsu:
Close this PR as we make an alternative fix in https://github.com/antrea-io/antrea/pull/5770.

PR #347, opened 2024-03-06
Stop referencing the VMware Harbor registry for ipfix-collector

antoninbas:
The ipfix-collector YAML manifest should no longer reference the VMware Harbor registry. The image should be pulled from DockerHub.

See https://github.com/antrea-io/antrea/issues/6003

### DIFF
--- a/build/yamls/collector/templates/ipfix-collector.yaml
+++ b/build/yamls/collector/templates/ipfix-collector.yaml
@@ -49,7 +49,7 @@ spec:
         {{- end }}
         - --ipfix.port={{ .Values.port }}
         - --ipfix.transport={{ .Values.protocol }}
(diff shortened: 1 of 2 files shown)


heanlan:
Although I don't think anyone is actively using the kafka and zookeeper images, should we also move them to Antrea DockerHub to avoid deadlinks?
image

antoninbas:
@heanlan this can be done in a future PR, the images should not be going away any time soon, but we won't be able to push new ones

heanlan (approved):

heanlan:
Would you need a new tag at the moment? @antoninbas

antoninbas:
@heanlan No need for a tag, thanks

PR #350, opened 2024-05-29
Return receive-only channel in GetMsgChan()

antoninbas:
This is a small improvement: the message channel retrieved from the collecting process should be "receive-only".


heanlan (approved):
LGTM

PR #351, opened 2024-05-31
Add more features to test IPFIX collector web API

antoninbas:
When querying records, one can now specify the max number of flows that should be returned (starting with the most recent one), as well as the format of the response (json or text).

We also add a size limit for the flowRecords slice, to avoid unbounded memory usage.

### DIFF
--- a/cmd/collector/collector.go
+++ b/cmd/collector/collector.go
@@ -40,6 +40,8 @@ import (
 
 const (
 	logToStdErrFlag = "logtostderr"
+	// An arbitrary limit, to prevent unbounded memory usage.
+	maxFlowRecords = 4096
 )
 
 var (
@@ -48,7 +50,13 @@ var (
 	IPFIXTransport string
 	flowRecords    []string
 	mutex          sync.Mutex
+
+	flowTextSeparator = bytes.Repeat([]byte("="), 80)
 )
+
+type jsonResponse struct {
+	FlowRecords []string `json:"flowRecords"`
+}
 
 func initLoggingToFile(fs *pflag.FlagSet) {
 	var err error
@@ -134,6 +142,11 @@ func addIPFIXMessage(msg *entities.Message) {
 	}
 	mutex.Lock()
 	defer mutex.Unlock()
+	if len(flowRecords) >= maxFlowRecords {
+		// Zero and remove first element.
+		flowRecords[0] = "" // Ensure first element can be garbage-collected.
+		flowRecords = flowRecords[1:]
+	}
 	flowRecords = append(flowRecords, buf.String())
 }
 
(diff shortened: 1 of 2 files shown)


cmd/collector/collector.go
```diff
@@ -240,21 +253,54 @@ func newCollectorCommand() *cobra.Command {
 
 func flowRecordHandler(w http.ResponseWriter, r *http.Request) {
 	if r.Method == "GET" {
+		countP := r.URL.Query().Get("count")
+		var count int
+		if countP != "" {
+			var err error
+			if count, err = strconv.Atoi(countP); err != nil || count < 0 {
```
yuntanghsu, line 260: Do we allow count=0 here?

cmd/collector/collector_test.go
```diff
@@ -0,0 +1,154 @@
+// Copyright 2024 VMware, Inc.
```
yuntanghsu, line 1: Not related to this PR.
Should we use Vmware or Broadcom. I saw some projects on gitlab start use Broadcom instead?

cmd/collector/collector_test.go
```diff
@@ -0,0 +1,154 @@
...
+		{
+			name:           "invalid format",
+			formatParam:    "foobar",
+			expectedStatus: http.StatusBadRequest,
+		},
+		{
+			name:          "both params set",
+			countParam:    "2",
+			formatParam:   "text",
+			expectedFlows: []string{"flow1", "flow2"},
```
yuntanghsu, line 85: Shouldn't it be []string{"flow2", "flow3"}

cmd/collector/collector.go
antoninbas: sure, why not, it will just return an empty list

cmd/collector/collector_test.go
```diff
@@ -0,0 +1,154 @@
...
+		require.NoError(t, err)
+		flowRecordHandler(rr, req)
+		resp := rr.Result()
+		defer resp.Body.Close()
+		require.Equal(t, expectedStatus, resp.StatusCode)
+		body, err := io.ReadAll(resp.Body)
+		require.NoError(t, err)
+
+		if expectedStatus != http.StatusOK {
+			return
```
yuntanghsu, line 120: I think this return will ignore the last 3 tests.

cmd/collector/collector_test.go
antoninbas: That's correct, I need to check why the test is passing

cmd/collector/collector_test.go
antoninbas: yes, that's correct, good catch
I forgot to call `t.Run` to define the subtests

cmd/collector/collector_test.go
antoninbas: I can switch to Broadcom

yuntanghsu (approved):
LGTM

PR #352, opened 2024-08-21
Fix how decoding errors are handled for UDP clients in collector

antoninbas:
Previously, in case of a decoding error (e.g., an unknown template ID in a data set), the goroutine responsible for decoding messages would be stopped, but the client would not be deleted. This would lead to a deadlock as subsequent messages could not be written to the packetChan channel for the client, effectively breaking the UDP collecting process completely. To avoid this issue, we log an error when decoding fails, but keep the client alive. Note that for IPFIX over UDP, there is no way to notify a single UDP client of an error (with TCP, we can close the connection if we want). This is also why the template refresh mechanism exists.

We also fix possible race conditions in the UDP collecting process: all access to the clients map should be protected by locking the mutex.

### DIFF
--- a/pkg/collector/tcp.go
+++ b/pkg/collector/tcp.go
@@ -66,7 +66,9 @@ func (cp *CollectingProcess) startTCPServer() {
 
 func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
 	address := conn.RemoteAddr().String()
-	client := cp.createClient()
+	// The channels stored in clientHandler are not needed for the TCP client, so we do not
+	// initialize them.
+	client := &clientHandler{}
 	func() {
 		cp.mutex.Lock()
 		defer cp.mutex.Unlock()
--- a/pkg/collector/tcp.go
+++ b/pkg/collector/tcp.go
@@ -54,7 +54,10 @@ func (cp *CollectingProcess) startTCPServer() {
 				}
 			}
 			cp.wg.Add(1)
-			go cp.handleTCPClient(conn)
+			go func() {
+				defer cp.wg.Done()
+				cp.handleTCPClient(conn)
+			}()
 		}
 	}(cp.stopChan)
 	<-cp.stopChan
@@ -64,11 +67,21 @@ func (cp *CollectingProcess) startTCPServer() {
 func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
 	address := conn.RemoteAddr().String()
 	client := cp.createClient()
-	cp.addClient(address, client)
-	defer cp.wg.Done()
+	func() {
+		cp.mutex.Lock()
+		defer cp.mutex.Unlock()
+		cp.clients[address] = client
+	}()
+	defer func() {
+		cp.mutex.Lock()
+		defer cp.mutex.Unlock()
+		delete(cp.clients, address)
+	}()
 	defer conn.Close()
+	reader := bufio.NewReader(conn)
+	cp.wg.Add(1)
 	go func() {
-		reader := bufio.NewReader(conn)
+		defer cp.wg.Done()
 		for {
 			length, err := getMessageLength(reader)
 			if errors.Is(err, io.EOF) {
@@ -77,18 +90,18 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
 			}
 			if err != nil {
 				klog.ErrorS(err, "Error when retrieving message length")
-				cp.deleteClient(address)
 				return
 			}
 			buff := make([]byte, length)
 			_, err = io.ReadFull(reader, buff)
 			if err != nil {
 				klog.ErrorS(err, "Error when reading the message")
-				cp.deleteClient(address)
 				return
 			}
 			message, err := cp.decodePacket(bytes.NewBuffer(buff), address)
 			if err != nil {
+				// TODO: should we close the connection instead and force the client to
+				// re-open it?
 				klog.ErrorS(err, "Error when decoding packet")
 				continue
 			}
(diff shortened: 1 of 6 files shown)


antoninbas:
This actually supersedes https://github.com/vmware/go-ipfix/pull/277, which was never merged (because of a CLA issue?)

pkg/collector/udp.go
```diff
@@ -107,45 +106,63 @@ func (cp *CollectingProcess) startUDPServer() {
 					return
 				}
 				klog.V(2).Infof("Receiving %d bytes from %s", size, address.String())
-				cp.handleUDPClient(address)
-				cp.clients[address.String()].packetChan <- bytes.NewBuffer(buff[0:size])
```
antoninbas, line 111: deadlock would happen here in case of previous decoding error (as no one is consuming from the channel anymore)

pkg/collector/udp.go
```diff
@@ -107,45 +106,63 @@ func (cp *CollectingProcess) startUDPServer() {
...
+		defer func() {
+			cp.mutex.Lock()
+			defer cp.mutex.Unlock()
+			delete(cp.clients, addr)
+		}()
+		for {
+			select {
+			case <-cp.stopChan:
+				klog.Infof("Collecting process from %s has stopped.", addr)
+				return
```
yuntanghsu, line 147: A dumb question, why don't we have deadlock if we return here?

antoninbas: deadlock is not possible when we delete the "client", which we do when we return from the function
we were already doing it before this change
when we delete the client handler, if we get a new UDP message from the same client, we will recreate the client handler and restart a goroutine to process packets

more specifically, in this case the `stopChan` channel is being closed, so the collecting process is being stopped anyway. But the description above also applies to the timeout case below.

pkg/collector/tcp.go
```diff
@@ -64,11 +67,21 @@ func (cp *CollectingProcess) startTCPServer() {
 func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
 	address := conn.RemoteAddr().String()
 	client := cp.createClient()
-	cp.addClient(address, client)
-	defer cp.wg.Done()
+	func() {
```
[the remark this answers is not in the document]
antoninbas: I removed the `addClient` and `deleteClient` methods as they are not very useful and they make it too easy to write racy code, at least for the UDP case

In any case, the fact that we were not deleting the client when the client connection was closed was a bug IMO.

The only other change I did here was to move the `defer cp.wg.Done()` call to the parent function, as it's more readable and less error prone to keep the `Add` and corresponding `Done` calls in the same place.

pkg/collector/udp.go
```diff
@@ -107,45 +106,63 @@ func (cp *CollectingProcess) startUDPServer() {
 					return
 				}
 				klog.V(2).Infof("Receiving %d bytes from %s", size, address.String())
-				cp.handleUDPClient(address)
-				cp.clients[address.String()].packetChan <- bytes.NewBuffer(buff[0:size])
+				cp.handleUDPMessage(address, buff[0:size])
```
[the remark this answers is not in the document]
antoninbas: Yes, I can make that change, It will be more consistent with the TCP case.

pkg/collector/tcp.go
```diff
@@ -77,18 +90,18 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
...
 			buff := make([]byte, length)
 			_, err = io.ReadFull(reader, buff)
 			if err != nil {
 				klog.ErrorS(err, "Error when reading the message")
-				cp.deleteClient(address)
 				return
 			}
 			message, err := cp.decodePacket(bytes.NewBuffer(buff), address)
 			if err != nil {
+				// TODO: should we close the connection instead and force the client to
```
heanlan, line 103: What's the pros for letting the client to re-open the connection?

pkg/collector/udp.go
```diff
@@ -107,45 +110,63 @@ func (cp *CollectingProcess) startUDPServer() {
...
+		defer ticker.Stop()
+		defer func() {
+			cp.mutex.Lock()
+			defer cp.mutex.Unlock()
+			delete(cp.clients, addr)
+		}()
+		for {
+			select {
+			case <-cp.stopChan:
+				klog.Infof("Collecting process from %s has stopped.", addr)
```
heanlan, line 150: I haven't been actively following Antrea latest changes. Do we still want to gradually switch to structured loggings?

pkg/collector/tcp.go
antoninbas: Let the client know that an invalid IPFIX message was sent and force it to re-open the connection
But I didn't make the change in this PR

pkg/collector/udp.go
antoninbas: We always use structured logging for new log messages in Antrea. It would not be a bad idea to switch to structured logging in this library, but it is orthogonal to this bug fix. Because the library is not that big, unlike Antrea, and doesn't have a lot of concurrent contributions, the best approach would be a dedicated PR to transition all existing log messages to structured logging.

heanlan (approved):
LGTM

pkg/collector/udp.go
yuntanghsu: You're right. The old channel can still read one message and exit as expected.

yuntanghsu (approved):
LGTM

pkg/collector/udp.go
antoninbas: Actually, I think you're right. There is still a possibility of deadlock. Let me see how we can address that.

antoninbas: I pushed a new commit. I couldn't think of a simpler solution than adding an extra goroutine. PTAL and let me know if it solves the issue you pointed out.

yuntanghsu (approved):

PR #353, opened 2024-08-23
Use correct default value for templateRefreshTimeout

antoninbas:
The correct recommended default is 600s, not 1800s. 1800s is the recommended default for templateLifeTime (3 times
templateRefreshTimeout).

We also update the UDP collector process to timeout the client after templateLifeTime instead of templateRefreshTimeout. This makes more sense as templateLifeTime is a collector process configuration parameter, while templateRefreshTimeout is for the exporter process.


yuntanghsu (approved):
LGTM

PR #354, opened 2024-08-23
Fix connection close in exporter process

antoninbas:
There were a couple of issues:

* ConnToCollector is exposed to the consumers of the library, and in Antrea we call ConnToCollector whenever sending a record fails. So the library itself should not call ConnToCollector as this is inherently racy. For example, it is possible for templateRefCh to be closed twice, despite the isChanClosed test (check-then-act race). We could add a mutex and a boolean, but I feel like it is better to just let consumer call ConnToCollector.
* When the exporter process is sending over UDP, there was a typo where a `break` was used instead of `return`. The `break` statement breaks from the select case, and not from the surrounding for loop, so the goroutine would never terminate.

### DIFF
--- a/pkg/exporter/process.go
+++ b/pkg/exporter/process.go
@@ -20,7 +20,6 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net"
 	"sync"
 	"time"
@@ -32,7 +31,6 @@ import (
 )
 
 const startTemplateID uint16 = 255
-const defaultCheckConnInterval = 10 * time.Second
 const defaultJSONBufferLen = 5000
 
 type templateValue struct {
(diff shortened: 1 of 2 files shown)


antoninbas:
@heanlan @yuntanghsu this is a bit of a disruptive change, so I hope you can provide some feedback

yuntanghsu (approved):
LGTM.

In the description, I think it should be CloseConnToCollector instead of ConnToCollector? I don't find the usage for ConnToCollector...

yuntanghsu:
I think we will need to add CloseConnToCollector in the Stop() for IPFIXEXporter in antrea repo?

antoninbas:
> I think we will need to add CloseConnToCollector in the Stop() for IPFIXEXporter in antrea repo?

Yes I think it would be better. Right now if the FA configuration is updated to remove the IPFIX exporter, we will not close the connection, which seems wrong. It's independent of the change in this PR though, but it should be addressed at some point.

PR #364, opened 2024-09-10
Fix invalid function comment for InitExportingProcess

antoninbas:
The default for the template refresh timeout is 600s, not 1800s. This should have been modified as part of #353.


heanlan (approved):
LGTM

PR #368, opened 2024-09-12
Fix Docker builds for ipfix-collector and kafka-consumer

antoninbas:
The go version needs to be updated in the Dockerfiles.

We also make sure that the Github workflow which validates the Docker builds is run for pull requests.


heanlan (approved):
LGTM

PR #370, opened 2024-09-12
Update Github Actions with Dependabot

antoninbas:
This repo currently uses actions which use a deprecated Node.js version. Dependabot will make sure that we use up-to-date actions.


heanlan (approved):
LGTM

PR #373, opened 2024-09-20
Stop using projects.registry.vmware.com registry for Kafka images

antoninbas:
- For confluent images (Kafka and Zookeeper), we now use the standard images from the confluentinc docker.io organization.
- For the kafka-consumer image, we use antrea/kafka-consumer.


heanlan (approved):
LGTM

PR #379, opened 2024-10-24
Add clusterId IE to Antrea registry

antoninbas:
As a way to identify the K8s cluster from which the IPFIX record was exported.

The IE is of type string to allow for enough flexibility. This could be a qualified name or a UUID (for example). It should uniquely identify the cluster (to an external IPFIX collector).


antoninbas:
@heanlan I was hesitating to name this "clusterUUID" and use the `octetArray` type. This would reduce the size of the value from 36 bytes down to 16 bytes. However, it doesn't seem like we really support `octetArray` in the library. Using the `string` type but not the typical string representation of a UUID also seems suboptimal.
I thought that a `string` "clusterId" field would be a bit more flexible, but I am not sure what is best. Let me know what you think.
For reference, this is the corresponding draft PR for Antrea: https://github.com/antrea-io/antrea/pull/6769

heanlan (approved):
LGTM

yuntanghsu (approved):
LGTM

PR #380, opened 2024-11-01
Ability to support unknown IEs when decoding template sets

antoninbas:
Introduce a "decoding mode" configuration for the collector process, which can take one of 3 values: Strict, LenientKeepUnknown, LenientDropUnknown. When the mode is Strict, unknown IEs are rejected. This is the default mode and matches the behavior prior to this patch. When the mode is LenientKeepUnknown, unknown IEs are accepted and will be preserved in data records. When the mode is LenientDropUnknown, unknown IEs are accepted but will not be included in data records (creating a mismatch between the data record and the template).

When decoding is not strict and the IE is not found in the registry, the name and type of the IE are not known. Name is kept as an empty string. Type is always set to OctetArray. The length is known as it is included in the template record.

This change required adding support for the OctetArray type (fixed-length or variable-length).

### DIFF
--- a/pkg/entities/set.go
+++ b/pkg/entities/set.go
@@ -68,7 +68,7 @@ type set struct {
 	length       int
 }
 
-func NewSet(isDecoding bool) Set {
+func NewSet(isDecoding bool) *set {
 	if isDecoding {
 		return &set{
 			records:    make([]Record, 0),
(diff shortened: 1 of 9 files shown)


pkg/entities/set.go
```diff
@@ -189,3 +189,34 @@ func (s *set) createHeader(setType ContentType, templateID uint16) {
 		binary.BigEndian.PutUint16(s.headerBuffer[0:2], templateID)
 	}
 }
+
+// MakeTemplateSet is a convenience function which creates a template Set with a single Record.
+func MakeTemplateSet(templateID uint16, ies []*InfoElement) (*set, error) {
```
heanlan, line 194: Do we want to add these two functions to Set interface?

antoninbas: These functions create a `Set` so I don't think we can add them to the interface?

heanlan: right, my brain is damaged

PR #381, opened 2024-11-14
Fix UDP template expiry in collector process

antoninbas:
Prior to this change, for every template received by the UDP collector, a new goroutine with a ticker was created, responsible for "expiring" the template after the proper timeout. After expiration, the ticker was stopped and te goroutine would complete (the ticker would only ever tick once, so really a timer should have been used?). If a client were to send a refreshed template, the old goroutine / ticker would not be stopped, and the new template would be deleted by the old goroutine, which is invalid. Additionally, starting a new goroutine every time a template message is received is a bit expensive.

With the new implementation, we use a timer created with time.AfterFunc. All the timers are handled by the go runtime with a single goroutine, making this very lightweight. The timers are stored as part of the template map. When a template is refreshed, we reset the existing timer. When a template is deleted, we stop the timer.

This is the first in a series of patch to improve template management in the collector process.

### DIFF
--- a/pkg/collector/clock.go
+++ b/pkg/collector/clock.go
@@ -1,0 +1,156 @@
+// Copyright 2024 VMware, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collector
+
+import (
+	"sync"
+	"time"
+)
+
+// timer allows for injecting fake or real timers into code that needs to do arbitrary things based
+// on time. We do not include the C() method, as we only support timers created with AfterFunc.
+type timer interface {
+	Stop() bool
+	Reset(d time.Duration) bool
+}
+
+// clock allows for injecting fake or real clocks into code that needs to do arbitrary things based
+// on time. We only support a very limited interface at the moment, with only the methods required
+// by CollectingProcess.
+type clock interface {
+	Now() time.Time
+	AfterFunc(d time.Duration, f func()) timer
+}
+
+// realClock implements the clock interface using functions from the time package.
+type realClock struct{}
+
+func (realClock) Now() time.Time {
+	return time.Now()
+}
+
+func (realClock) AfterFunc(d time.Duration, f func()) timer {
+	return time.AfterFunc(d, f)
+}
+
+type fakeTimer struct {
+	targetTime time.Time
+	f          func()
+	clock      *fakeClock
+}
+
+func (t *fakeTimer) Stop() bool {
+	clock := t.clock
+	clock.m.Lock()
+	defer clock.m.Unlock()
+	newTimers := make([]*fakeTimer, 0, len(clock.timers))
+	fired := true
+	for i := range clock.timers {
+		if clock.timers[i] != t {
+			newTimers = append(newTimers, t)
+			continue
+		}
+		// timer is found so it hasn't been fired yet
+		fired = false
+	}
+	clock.timers = newTimers
+	return !fired
+}
+
+func (t *fakeTimer) Reset(d time.Duration) bool {
+	clock := t.clock
+	clock.m.Lock()
+	defer clock.m.Unlock()
+	fired := true
+	for i := range clock.timers {
+		if clock.timers[i] != t {
+			continue
+		}
+		// timer is found so it hasn't been fired yet
+		fired = false
+		t.targetTime = clock.now.Add(d)
+	}
+	return !fired
+}
+
+// fakeClock implements the clock interface as a virtual clock meant to be used in tests. Time can
+// be advanced arbitrarily, but does not change on its own.
+type fakeClock struct {
+	m           sync.RWMutex
+	isAdvancing bool
+	now         time.Time
+	timers      []*fakeTimer
+}
+
+func newFakeClock(t time.Time) *fakeClock {
+	return &fakeClock{
+		now: t,
+	}
+}
+
+func (c *fakeClock) Now() time.Time {
+	c.m.RLock()
+	defer c.m.RUnlock()
+	return c.now
+}
+
+func (c *fakeClock) AfterFunc(d time.Duration, f func()) timer {
+	if d <= 0 {
+		panic("negative duration not supported")
+	}
+	c.m.Lock()
+	defer c.m.Unlock()
+	t := &fakeTimer{
+		targetTime: c.now.Add(d),
+		f:          f,
+		clock:      c,
+	}
+	c.timers = append(c.timers, t)
+	return t
+}
+
+func (c *fakeClock) Step(d time.Duration) {
+	if d < 0 {
+		panic("invalid duration")
+	}
+	timerFuncs := []func(){}
+	func() {
+		c.m.Lock()
+		defer c.m.Unlock()
+		if c.isAdvancing {
+			panic("concurrent calls to Step() not allowed")
+		}
+		c.isAdvancing = true
+		c.now = c.now.Add(d)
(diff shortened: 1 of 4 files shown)


pkg/collector/process.go
```diff
@@ -361,47 +375,94 @@ func (cp *CollectingProcess) decodeDataSet(dataBuffer *bytes.Buffer, obsDomainID
...
+			//   returns false, Reset neither waits for the prior f to complete before
+			//   returning nor does it guarantee that the subsequent goroutine running f
+			//   does not run concurrently with the prior one. If the caller needs to
+			//   know whether the prior execution of f is completed, it must coordinate
+			//   with f explicitly.
+			// In our case, when f executes, we have to verify that the record is indeed
+			// scheduled for deletion by checking expiryTime. We cannot just
+			// automatically delete the template.
+
+			// No reason to try to stop the timer in this case, even though it would be
```
heanlan, line 414: Trying to follow the logic, is it because the expiryTimer is already nil, so there's no point to set stopTimer to true? Because I was thinking about L444, wondering if there's a case when stopTimer is true but expiryTimer is nil.

antoninbas: this code executes as part of the function invoked by the timer. Broadly speaking, if the function has been invoked by the timer, then the timer has already fired and there is no need to stop it. There are some edge cases with `AfterFunc` which are described in the documentation snippet above.
Essentially there are 2 cases:
- the timer fires because we need to expire the template. This is the "normal" case. The timer has fired and there is no need to stop it.
- the timer fires, but there is a concurrent call to `addTemplate` (and to `AfterFunc`). This is the edge case described in the doc. We rely on the condition below (`!tpl.expiryTime.After(now)`) to determine whether we should just ignore this firing of the timer. The synchronization provided by `cp.mutex` is very important here. We will wait until the call to `addTemplate` returns and releases the lock. If the condition is false (which will be the case), we will not call `expiryTimer.Stop()` in `deleteTemplateWithConds` (even if `stopTimer` was set to true), as the template entry has been updated and the current run of the timer function should be ignored and be a no-op.

Calling `expiryTimer.Stop()` even if the timer has already been stopped is harmless. But it doesn't work well with the fake clock unfortunately. If the fake clock implementation is fixed, I may remove `stopTimer`.

The `nil` check at line L444 is more for defensive programming and for the TCP case.

pkg/collector/process.go
```diff
@@ -361,47 +375,94 @@ func (cp *CollectingProcess) decodeDataSet(dataBuffer *bytes.Buffer, obsDomainID
...
+// deleteTemplate returns true iff a template was actually deleted.
+func (cp *CollectingProcess) deleteTemplate(obsDomainID uint32, templateID uint16) bool {
+	return cp.deleteTemplateWithConds(obsDomainID, templateID, true)
+}
+
+// deleteTemplateWithConds returns true iff a template was actually deleted.
+func (cp *CollectingProcess) deleteTemplateWithConds(obsDomainID uint32, templateID uint16, stopTimer bool, condFns ...func(*template) bool) bool {
+	cp.mutex.Lock()
+	defer cp.mutex.Unlock()
+	template, ok := cp.templatesMap[obsDomainID][templateID]
```
yuntanghsu, line 435: Wouldn't the templateID be deleted before, and the timer cannot be stopped when we can't find the templateID in templatesMap?

antoninbas: which case / code path are you referring to?
the only case where `stopTimer` should be `false` is if the caller knows for certain than the timer has already expired (taking into account that `condFns` should all be true).

yuntanghsu: I mean, when we remove the `templateID` from a `templatesMap`, shouldn't we call the `expiryTimer.Stop()`? otherwise; wouldn't expiryTimer keep running?

yuntanghsu: Could you also remind me where we call `deleteTemplate`? If we call `deleteTemplate` after a template is expired, deleteTemplate will return false and the `expiryTimer. Stop ()` won't be called?

antoninbas: > I mean, when we remove the templateID from a templatesMap, shouldn't we call the expiryTimer.Stop()? otherwise; wouldn't expiryTimer keep running?

We do this a few lines below:
```go
	if stopTimer && template.expiryTimer != nil {
		template.expiryTimer.Stop()
	}
```

> Could you also remind me where we call deleteTemplate? If we call deleteTemplate after a template is expired, deleteTemplate will return false and the expiryTimer. Stop () won't be called?

`deleteTemplate` is not called right now actually. I have a follow up patch which deletes templates when TCP connections are closed, but I wanted to split PRs to keep them small. Sorry for the confusion. At the moment we call `deleteTemplateWithConds` when expiring a template. During this call, we do not need to Stop the timer, as we are guaranteed that the timer has already fired (see comments in `addTemplate`).

yuntanghsu: Thanks for the explanation. So stopTimer is mainly for TCP when we want to delete the template before it is expired.

yuntanghsu (approved):
LGTM

antoninbas:
@heanlan @yuntanghsu Because there was some legitimate confusion around the use of `stopTimer` and the correctness of the code, I have added a second commit which removes this parameter. The only reason why I added `stopTimer` was because of limitations in the fake clock implementation provided by k8s.io/utils. But it's usually not a good idea to modify logic to accommodate unit tests, even though the code was still correct. In the new commit, I no longer use the fake clock provided by k8s.io/utils, and I define a custom clock interface which satisfies the needs of the `CollectingProcess`, along with a corresponding fake clock implementation for the interface, which can be used in unit tests. PTAL, hopefully the code will make more sense now. More details in the commit message for the new commit.

pkg/collector/clock.go
```diff
@@ -0,0 +1,156 @@
...
+
+func (c *fakeClock) Step(d time.Duration) {
+	if d < 0 {
+		panic("invalid duration")
+	}
+	timerFuncs := []func(){}
+	func() {
+		c.m.Lock()
+		defer c.m.Unlock()
+		if c.isAdvancing {
```
yuntanghsu, line 132: Just curious, why don't we directly use lock at the beginning of this func? Or we want the concurrent calls to return immediately instead of waiting to be executed?

pkg/collector/clock.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func (realClock) AfterFunc(d time.Duration, f func()) timer {
+	return time.AfterFunc(d, f)
+}
+
+type fakeTimer struct {
+	targetTime time.Time
+	f          func()
+	clock      *fakeClock
```
yuntanghsu, line 51: The clock here is for deleting itself from the parent clock in convenience?

pkg/collector/process.go
```diff
@@ -361,47 +374,95 @@ func (cp *CollectingProcess) decodeDataSet(dataBuffer *bytes.Buffer, obsDomainID
...
+	if !ok {
+		return false
+	}
+	for _, condFn := range condFns {
+		if !condFn(template) {
+			return false
 		}
-	}()
+	}
+	// expiryTimer will be nil when the protocol is UDP.
```
yuntanghsu, line 440: Could you remind me when will the expiryTimer be set to nil when the protocol is UDP?

pkg/collector/clock.go
antoninbas: We can't lock at the beginning of the function and for the duration of the function, because we cannot call the AfterFunc f functions while holding the lock. This is what the K8s fake clock for testing does, and it means the f functions cannot call `Now`, or `timer.Stop`, as explained in the commit message.

At the same time, it would break functionality of the clock to allow a concurrent call to `Step`. So we need a boolean flag  (`c.isAdvancing`) which enables us to panic if that happens. This is more of a sanity check to avoid invalid usages of the fake clock.

pkg/collector/clock.go
antoninbas: It's not a convenience, it's required for functionality (implementation of `Stop`). But yes, it is so the timer can remove itself from the clock's list of timers.

pkg/collector/process.go
antoninbas: It's a typo, it should be TCP

antoninbas: Fixed, thanks for catching it

yuntanghsu (approved):
LGTM

PR #383, opened 2024-11-20
Delete existing template if a new invalid one is received

antoninbas:
This is particularly useful for UDP collection, as there is no feedback mechanism to let the sender know that the new template is invalid (while with TCP, we can close the connection). If we keep the old template and the sender sends data records which use the new template, we would try to decode them according to the old template, which would cause issues. Instead we will ignore data records for that observation domain and template ID until a new valid template is received.

### DIFF
--- a/pkg/collector/process.go
+++ b/pkg/collector/process.go
@@ -261,12 +261,7 @@ func (cp *CollectingProcess) decodeTemplateSet(templateBuffer *bytes.Buffer, obs
 		return nil, err
 	}
 
-	templateSet := entities.NewSet(true)
-	if err := templateSet.PrepareSet(entities.Template, templateID); err != nil {
-		return nil, err
-	}
-	elementsWithValue := make([]entities.InfoElementWithValue, int(fieldCount))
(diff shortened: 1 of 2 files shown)


yuntanghsu:
I think the new template will have new template_id? Then it should be good if new template has any issues?
Or it is possible that user update its old template and use the same template_id?

User should be responsible for the missing date when they provide invalid template when using UDP, right?

antoninbas:
> Or it is possible that user update its old template and use the same template_id?

We always use the same template ID in Antrea, even when we add new fields / IEs

> User should be responsible for the missing data when they provide invalid template when using UDP, right?

Yes, but they key change in this PR is to make sure we won't try to decode a "new" data record using an "old" template, which can actually cause a panic.
With UDP, when the exporter sends a template, it will assume that it was accepted and will proceed with sending data records right away. That creates issues if the template was actually rejected. I think the best course of action (assuming the same template ID was used of course, otherwise there is no issue) is to forget the old template in that case and have a clean state.

yuntanghsu (approved):
LGTM

PR #384, opened 2024-11-22
Close TCP connection in case of error

antoninbas:
This is better than keeping the connection open, as it sends a signal to the exporter that something is wrong.

This is also one of the four possible behaviors described by RFC7011 for handling malformed IPFIX messages for TCP and SCTP connections. See https://datatracker.ietf.org/doc/html/rfc7011#section-9.1. Note that implementing one of these four options is not mandatory. However, the RFC also states that:

> the Collecting Process SHOULD stop processing IPFIX Messages
> from clearly malfunctioning Exporting Processes (e.g., those from
> which the last few IPFIX Messages have been malformed).

Closing the TCP connection is the easiest thing for us to implement, and it makes sense to use IMO.

### DIFF
--- a/pkg/collector/tcp.go
+++ b/pkg/collector/tcp.go
@@ -81,9 +81,14 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
 	}()
 	defer conn.Close()
 	reader := bufio.NewReader(conn)
+	doneCh := make(chan struct{})
 	cp.wg.Add(1)
+	// We read from the connection in a separate goroutine, so we can stop immediately when
+	// cp.StopChan is closed. An alternative would be to use a read deadline, and check
+	// cp.StopChan at every iteration.
 	go func() {
 		defer cp.wg.Done()
+		defer close(doneCh)
 		for {
 			length, err := getMessageLength(reader)
 			if errors.Is(err, io.EOF) {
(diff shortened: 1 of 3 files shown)


pkg/collector/tcp.go
```diff
@@ -81,9 +81,14 @@ func (cp *CollectingProcess) handleTCPClient(conn net.Conn) {
...
 	defer conn.Close()
 	reader := bufio.NewReader(conn)
+	doneCh := make(chan struct{})
 	cp.wg.Add(1)
+	// We read from the connection in a separate goroutine, so we can stop immediately when
+	// cp.StopChan is closed. An alternative would be to use a read deadline, and check
+	// cp.StopChan at every iteration.
 	go func() {
 		defer cp.wg.Done()
+		defer close(doneCh)
```
antoninbas, line 91: this could also be considered a bug fix. Prior to this commit, `return` statements in the `for` loop would cause this goroutine to return, but we would block on `cp.stopChan` indefinitely (so I think there would be a goroutine leak, as the parent goroutine, which invoked `handleTCPClient` would never return).

yuntanghsu: Wouldn't it still be blocked in startTCPServer?

yuntanghsu (approved):
LGTM

PR #385, opened 2024-11-23
Support resetting expired timer in fake clock

antoninbas:
This is how timers from the standard library behave.

### DIFF
--- a/pkg/collector/clock.go
+++ b/pkg/collector/clock.go
@@ -82,6 +82,10 @@ func (t *fakeTimer) Reset(d time.Duration) bool {
 		fired = false
 		t.targetTime = clock.now.Add(d)
 	}
+	if fired {
+		t.targetTime = clock.now.Add(d)
+		clock.timers = append(clock.timers, t)
+	}
 	return !fired
 }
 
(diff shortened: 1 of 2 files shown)


pkg/collector/clock.go
```diff
@@ -82,6 +82,10 @@ func (t *fakeTimer) Reset(d time.Duration) bool {
 		fired = false
 		t.targetTime = clock.now.Add(d)
```
yuntanghsu, line 83: Not related to this PR, but I'm wondering if we don't have duplicate timer in the clock, maybe we can break here?

pkg/collector/clock.go
```diff
@@ -82,6 +82,10 @@ func (t *fakeTimer) Reset(d time.Duration) bool {
...
 	}
+	if fired {
+		t.targetTime = clock.now.Add(d)
```
yuntanghsu, line 86: Looks like we need to update the targetTime whether we found the timer or not? Do you think it's better to move it out of the bracket?

yuntanghsu (approved):
LGTM,
Just two nits.

yuntanghsu (approved):

PR #386, opened 2024-12-04
Manage templates per session instead of globally

antoninbas:
As per the IPFIX specification, a template is specific to a transport session. Observation domains are not guaranteed to be unique across transport sessions.

When a transport session is closed (TCP connection closed or UDP session timeout), all associated templates are removed. Note that prior to this change, templates associated to a TCP connection were never cleaned up, causing a memory leak.

### DIFF
--- a/pkg/collector/clock.go
+++ b/pkg/collector/clock.go
@@ -22,8 +22,17 @@ import (
 // timer allows for injecting fake or real timers into code that needs to do arbitrary things based
 // on time. We do not include the C() method, as we only support timers created with AfterFunc.
 type timer interface {
+	C() <-chan time.Time
 	Stop() bool
 	Reset(d time.Duration) bool
+}
+
+type realTimer struct {
+	*time.Timer
+}
+
+func (t *realTimer) C() <-chan time.Time {
+	return t.Timer.C
 }
 
 // clock allows for injecting fake or real clocks into code that needs to do arbitrary things based
@@ -32,6 +41,7 @@ type timer interface {
 type clock interface {
 	Now() time.Time
 	AfterFunc(d time.Duration, f func()) timer
+	NewTimer(d time.Duration) timer
 }
 
 // realClock implements the clock interface using functions from the time package.
@@ -42,13 +52,26 @@ func (realClock) Now() time.Time {
 }
 
 func (realClock) AfterFunc(d time.Duration, f func()) timer {
-	return time.AfterFunc(d, f)
+	return &realTimer{
+		Timer: time.AfterFunc(d, f),
+	}
+}
+
+func (realClock) NewTimer(d time.Duration) timer {
+	return &realTimer{
+		Timer: time.NewTimer(d),
+	}
 }
 
 type fakeTimer struct {
 	targetTime time.Time
 	f          func()
+	ch         chan time.Time
 	clock      *fakeClock
+}
+
+func (t *fakeTimer) C() <-chan time.Time {
+	return t.ch
 }
 
 func (t *fakeTimer) Stop() bool {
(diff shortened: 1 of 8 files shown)


pkg/collector/clock.go
```diff
@@ -140,18 +184,41 @@ func (c *fakeClock) Step(d time.Duration) {
...
+				if !c.m.TryRLock() {
+					return true
+				}
+				defer c.m.RUnlock()
+				// If timer has been stopped or reset, do not fire.
+				if t.targetTime.IsZero() || t.targetTime.After(c.now) {
+					return false
+				}
+				select {
+				case t.ch <- c.now:
```
yuntanghsu, line 211: I think t.ch is used to check if a timer is fired? Then why don't we need it after executing a function at line 199?
The other question is timer.C() won't return when a time is stopped?

pkg/collector/clock.go
```diff
@@ -140,18 +184,41 @@ func (c *fakeClock) Step(d time.Duration) {
...
+					return false
+				default:
+				}
+				return true
+			}
+			// Spin until we can acquire a read lock and write to the C channel: this
```
yuntanghsu, line 217: Just out of curiosity, why do we only consider the concurrent calls when t.f is nil?

pkg/collector/clock.go
```diff
@@ -70,6 +94,9 @@ func (t *fakeTimer) Stop() bool {
 }
 
 func (t *fakeTimer) Reset(d time.Duration) bool {
+	if d <= 0 {
```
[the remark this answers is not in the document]
antoninbas: If I accept `d < 0`, I have to potentially call the step function here, which would just add complexity. This can be supported at a later time if needed.

pkg/collector/clock.go
```diff
@@ -124,11 +151,28 @@ func (c *fakeClock) AfterFunc(d time.Duration, f func()) timer {
...
+		// time package.
+		ch:    make(chan time.Time),
+		clock: c,
+	}
+	c.timers = append(c.timers, t)
+	return t
+}
+
 func (c *fakeClock) Step(d time.Duration) {
 	if d < 0 {
```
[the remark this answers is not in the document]
antoninbas: Actually, there is no harm in supporting `d == 0`. This is a way to run expired timers without advancing clock time (even though that is not super useful).

pkg/collector/clock.go
antoninbas: There is no `t.ch` for function-based timers (created with `AfterFunc`).

> The other question is timer.C() won't return when a time is stopped?

From the `time` documentation:

> For a chan-based timer created with NewTimer(d), as of Go 1.23, any receive from t.C after Stop has returned is guaranteed to block rather than receive a stale time value from before the Stop; if the program has not received from t.C already and the timer is running, Stop is guaranteed to return true. 

So it is correct, the channel will block if the timer has been stopped.

pkg/collector/clock.go
antoninbas: The behavior is different for AfterFun-based timers (`t.f != nil`) and channel-based timers (`t.f == nil`). For details, refer to https://pkg.go.dev/time#Timer.Stop. The specified behavior for channel-based timers (Go >= 1.23) is harder to achieve, and the code is more complex.

pkg/collector/clock.go
yuntanghsu: Thanks for the explanation

yuntanghsu (approved):
LGTM

antoninbas:
Thanks for the review @yuntanghsu !

PR #388, opened 2024-12-16
Remove CloseMsgChan() method from collector process

antoninbas:
Having a seprate method to close an "output" channel is an anti-pattern IMO, and it is not clear how library consumers are supposed to use it, in particular in relation to the Stop() method.

Instead, we can close the channel when the Stop() method is closed, and all goroutines which may write to the channel have been stopped. Goroutines which read from the channel can use this as a signal to stop - or can be signalled directly by the goroutine which called Stop() in the first place.


yuntanghsu (approved):
LGTM

PR #389, opened 2024-12-18
Support Records as input for AggregationProcess

antoninbas:
This enables consumers of this library to do some preprocessing on data records included in messages received by the collector, prior to passing on the records to the AggregationProcess. Without this change, a consumer has to either mutate the record objects directly (which may require adding new methods to the Record interface, and these methods may not be very meaningful or may be very specialized), or has to duplicate records and wrap them in a new data set and a new message, which is cumbersome and expensive. With this change, a consumer can duplicate records (quite cheaply, by reusing the IE list), and pass them directly to the AggregationProcess.

Note that we maintain backwards-compatibility. When initializing the AggregationProcess, one can choose whether data will be fed as messages or individual records.


yuntanghsu (approved):
LGTM

PR #390, opened 2024-12-19
Improve IPFIX Set implementation

antoninbas:
* Optimize ResetSet to avoid memory allocations
* Add new interface method AddRecordV3 which takes a Record as a parameter (instead of a list of IEs). AddRecordV3 can avoid all memory allocations.

We also add a new GetRecordType method to the Record interface.

### DIFF
--- a/pkg/entities/set.go
+++ b/pkg/entities/set.go
@@ -34,6 +34,7 @@ const (
 	SetHeaderLen  int    = 4
 )
 
+// ContenType is used for both sets and records.
 type ContentType uint8
 
 const (
@@ -45,6 +46,8 @@ const (
 
 type Set interface {
 	PrepareSet(setType ContentType, templateID uint16) error
+	// Call ResetSet followd by a new call to PrepareSet in order to reuse an existing Set,
+	// instead of instantiating a new one.
 	ResetSet()
 	GetHeaderBuffer() []byte
 	GetSetLength() int
@@ -56,6 +59,11 @@ type Set interface {
 	// one. This can result in fewer memory allocations. The caller should not modify the
 	// contents of the slice after calling AddRecordV2.
 	AddRecordV2(elements []InfoElementWithValue, templateID uint16) error
+	// Unlike other AddRecord* variants, AddRecordV3 takes an actual existing Record as an input
+	// parameter, instead of a list of Information Elements with values. When calling
+	// AddRecordV3, the Set effectively takes ownership of the Record, and the Record should no
+	// longer be modified by the caller.
+	AddRecordV3(record Record) error
 	GetRecords() []Record
 	GetNumberOfRecords() uint32
 }
@@ -63,6 +71,7 @@ type Set interface {
 type set struct {
 	headerBuffer []byte
 	setType      ContentType
+	templateID   uint16
 	records      []Record
 	isDecoding   bool
 	length       int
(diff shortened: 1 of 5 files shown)


antoninbas:
Benchmark results without ResetSet improvements:
```
C02F16ASMD6R:go-ipfix abas$ go test -run=xxx -bench=BenchmarkSet -benchmem ./pkg/entities/
goos: darwin
goarch: amd64
pkg: github.com/vmware/go-ipfix/pkg/entities
BenchmarkSet-16    	 4014678	       303.4 ns/op	     500 B/op	       6 allocs/op
PASS
ok  	github.com/vmware/go-ipfix/pkg/entities	1.861s
```

And with the improvements:
```
C02F16ASMD6R:go-ipfix abas$ go test -run=xxx -bench=BenchmarkSet -benchmem ./pkg/entities/
goos: darwin
goarch: amd64
pkg: github.com/vmware/go-ipfix/pkg/entities
BenchmarkSet-16    	16009188	        74.63 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/vmware/go-ipfix/pkg/entities	1.631s
```

pkg/entities/set.go
```diff
@@ -45,6 +46,8 @@ const (
 
 type Set interface {
 	PrepareSet(setType ContentType, templateID uint16) error
+	// Call ResetSet followd by a new call to PrepareSet in order to reuse an existing Set,
```
yuntanghsu, line 49:
```suggestion
	// Call ResetSet followed by a new call to PrepareSet in order to reuse an existing Set,
```

pkg/entities/set.go
```diff
@@ -99,13 +109,14 @@ func (s *set) PrepareSet(setType ContentType, templateID uint16) error {
...
 	if !s.isDecoding {
-		s.headerBuffer = nil
-		s.headerBuffer = make([]byte, SetHeaderLen)
+		clear(s.headerBuffer)
 		s.length = SetHeaderLen
 	}
 	s.setType = Undefined
-	s.records = nil
-	s.records = make([]Record, 0)
+	// Clear before shrinking the slice so that existing elements are eligible for garbage collection.
```
yuntanghsu, line 116: Do we need to reset the s.length here? I'm not quite sure what's the case when it isDecoding.

pkg/entities/set.go
```diff
@@ -174,6 +185,20 @@ func (s *set) AddRecordV2(elements []InfoElementWithValue, templateID uint16) er
 	return nil
 }
 
+func (s *set) AddRecordV3(record Record) error {
+	// Sanity check: we need to make sure that the record is allowed to be added.
+	recordType := record.GetRecordType()
+	if recordType != s.setType {
+		return fmt.Errorf("Record and Set types don't match")
```
yuntanghsu, line 192:
```suggestion
		return fmt.Errorf("record and Set types don't match")
```

pkg/entities/set.go
```diff
@@ -174,6 +185,20 @@ func (s *set) AddRecordV2(elements []InfoElementWithValue, templateID uint16) er
...
+	}
+	if recordType == Data && record.GetTemplateID() != s.templateID {
+		return fmt.Errorf("All Data Records in the same Data Set must have the same template ID")
```
yuntanghsu, line 195:
```suggestion
		return fmt.Errorf("all Data Records in the same Data Set must have the same template ID")
```

pkg/entities/set.go
antoninbas: I can reset it to 0, even though it's a change to the existing behavior. I agree that it would make more sense.
I am not a big fan of the `isDecoding` flag everywhere. We should have `Marhsal` / `Unmarshal` methods as appropriate, but that will have to be for a later time...

antoninbas:
@yuntanghsu Thanks for the review, I addressed comments

yuntanghsu (approved):
LGTM

PR #391, opened 2024-12-20
Add SendDataRecords to IPFIX Exporter

antoninbas:
As a more efficient and covenient alternative to SendSet.

* SendDataRecords takes a list of data Records (for the same template ID), and breaks them up into multiple IPFIX messages (with one data Set per message) as needed.
* SendDataRecords can take a reusable buffer as a parameter, to avoid repeated memory allocations.

We also introduce the MaxMsgSize input for the Exporter. It will be used by SendDataRecords to determine how to break up the list of Records into multiple messages. When using UDP, MaxMsgSize should be set based on the PMTU (Path MTU) to avoid fragmentation.

### DIFF
--- a/pkg/exporter/msg.go
+++ b/pkg/exporter/msg.go
@@ -15,6 +15,7 @@
 package exporter
 
 import (
+	"bytes"
 	"fmt"
 	"time"
 
@@ -22,6 +23,16 @@ import (
 )
 
 func CreateIPFIXMsg(set entities.Set, obsDomainID uint32, seqNumber uint32, exportTime time.Time) ([]byte, error) {
+	var buf bytes.Buffer
+	if _, err := WriteIPFIXMsgToBuffer(set, obsDomainID, seqNumber, exportTime, &buf); err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+// WriteIPFIXMsgToBuffer serializes the set as an IPFIX message and writes it to the provided
+// buffer. It returns the message length and an error if applicable.
+func WriteIPFIXMsgToBuffer(set entities.Set, obsDomainID uint32, seqNumber uint32, exportTime time.Time, buf *bytes.Buffer) (int, error) {
 	// Create a new message and use it to send the set.
 	msg := entities.NewMessage(false)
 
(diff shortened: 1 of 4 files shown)


pkg/exporter/msg.go
```diff
@@ -30,7 +41,7 @@ func CreateIPFIXMsg(set entities.Set, obsDomainID uint32, seqNumber uint32, expo
 	msgLen := entities.MsgHeaderLength + set.GetSetLength()
 	if msgLen > entities.MaxSocketMsgSize {
```
yuntanghsu, line 42: Should we use expProc.maxMsgSize here? This should be the same case when input.MaxMsgSize=0, but I'm kind of confused about what if user set their own MaxMsgSize?

pkg/exporter/process.go
```diff
@@ -249,21 +259,96 @@ func (ep *ExportingProcess) SendSet(set entities.Set) (int, error) {
...
+				if err := set.AddRecordV3(record); err != nil {
+					return recordsSent, bytesSent, fmt.Errorf("error when adding record to data set: %w", err)
+				}
+				numRecordsInSet += 1
+				length += recordLength
+				continue
+			}
+			// There is no record in the set currently, yet this new record cannot fit!
+			if numRecordsInSet == 0 {
+				return recordsSent, bytesSent, fmt.Errorf("record exceeds max size")
```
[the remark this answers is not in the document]
antoninbas: - If the message exceeds the maximum size set by the user, then we return an error here (and the user can choose how to handle this error)
- If the message does not exceed the maximum size, but the maximum size was not configured properly by the user (it exceeds the Path MTU for UDP export), then the message will be fragmented.

The second case is what the comment describes, not the first case.

pkg/exporter/msg.go
antoninbas: I chose to keep this function as is, as it is a freestanding function (also used by tests) and not an exporter method. This function only enforces the default upper bound. The user-provided upper-bound is enforced in the exporter before calling this function.

There would be no real harm in removing this check altogether but I thought I would keep it just for cases where the function is called by tests.

yuntanghsu (approved):
LGTM

PR #394, opened 2025-01-04
Improve received stats in collector

antoninbas:
GetNumRecordsReceived() was returning the number of IPFIX messages. This was misleading as a message is different from a record, and a message may include multiple records.

We add support for more stats (num messages, num template sets, num data sets, num template records, num data records). At the moment, not all of these stats are exposed through public methods.

GetNumRecordsReceived() is updated to return the number of received data records. GetNumMessagesReceived() is introduced to return the number of received IPFIX messages, which matches the previous implementation of GetNumRecordsReceived().

Finally, we update the collector code so that stats are incremented before sending the message on the channel, instead of after. This seems more correct to me, but also it avoid a possible race condition in unit tests, which assume that by the time a message has been received on the channel (by the test code), stats have already been updated.

### DIFF
--- a/pkg/collector/process.go
+++ b/pkg/collector/process.go
@@ -84,11 +84,16 @@ type CollectingProcess struct {
 	// decoding.
 	decodingMode DecodingMode
 	// caCert, serverCert and serverKey are for storing encryption info when using TLS/DTLS
-	caCert               []byte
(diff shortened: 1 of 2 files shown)


yuntanghsu (approved):
LGTM

PR #395, opened 2025-01-08
Improve IPFIX header decoding in collector

antoninbas:
Stop using util.Decode to decode header data

util.Decode calls binary.Read in a loop and each call allocates a byte slice.
Instead we can use binary.BigEndian methods directly.


pkg/collector/process.go
```diff
@@ -393,22 +454,22 @@ func getMessageLength(reader *bufio.Reader) (int, error) {
 	if err != nil {
 		return 0, err
 	}
-	var msgLen uint16
-	err = util.Decode(bytes.NewBuffer(partialHeader[2:]), binary.BigEndian, &msgLen)
```
antoninbas, line 397: this should avoid a memory allocation as well

antoninbas:
Benchmark results.

Before:
```
$ go test -run=xxx -bench=BenchmarkCollectingProcess_DecodeDataRecord -benchmem ./pkg/collector/...
goos: darwin
goarch: amd64
pkg: github.com/vmware/go-ipfix/pkg/collector
BenchmarkCollectingProcess_DecodeDataRecord-16    	  270014	      3914 ns/op	     504 B/op	      21 allocs/op
PASS
ok  	github.com/vmware/go-ipfix/pkg/collector	17.955s
```

After:
```
$ go test -run=xxx -bench=BenchmarkCollectingProcess_DecodeDataRecord -benchmem ./pkg/collector/...
goos: darwin
goarch: amd64
pkg: github.com/vmware/go-ipfix/pkg/collector
BenchmarkCollectingProcess_DecodeDataRecord-16    	  349524	      3494 ns/op	     488 B/op	      14 allocs/op
PASS
ok  	github.com/vmware/go-ipfix/pkg/collector	22.516s
```

I was expecting a more significant improvement for CPU time, but at least the reduction in allocs/op is as expected (-7 as we decode 5 fields in the message header and 2 fields in the set header).

yuntanghsu (approved):
LGTM

antoninbas:
@yuntanghsu thanks for the review, I had to resolve a merge conflict so I need a new approval

yuntanghsu (approved):

PR #396, opened 2025-01-08
Support configurable decoding mode in reference collector program

antoninbas:
By introducing the "--decoding-mode" flag.

LenientKeepUnknown can be used in order to accept templates with unknown IEs. Note that these IEs will be skipped when serializing data records.

### DIFF
--- a/cmd/collector/collector.go
+++ b/cmd/collector/collector.go
@@ -17,13 +17,16 @@ package main
 import (
 	"bytes"
 	"context"
+	"encoding/hex"
 	"encoding/json"
 	"flag"
 	"fmt"
 	"net/http"
 	"os"
 	"os/signal"
+	"slices"
 	"strconv"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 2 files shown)


cmd/collector/collector.go
```diff
@@ -170,6 +186,9 @@ func run() error {
 	// Load the IPFIX global registry
 	registry.LoadRegistry()
 	// Initialize collecting process
+	if !slices.Contains(validDecodingModes, decodingMode) {
```
yuntanghsu, line 189: I think we can delete
```
	if decodingMode == "" {
		decodingMode = DecodingModeStrict
	}
```
in initCollectingProcess

yuntanghsu (commented):
LGTM.  just one nit

cmd/collector/collector.go
antoninbas: I don't think so. `pkg/collector` is used as a library, including by Antrea. It makes sense to have a defaulting mechanism there.
`cmd/collector` is just a reference implementation used for testing & development.

yuntanghsu (approved):
LGTM

PR #397, opened 2025-01-09
Prevent unnecessary memory allocations in WriteIPFIXMsgToBuffer

antoninbas:
Calling GetBuffer on a record allocates a new buffer which is then copied to the message buffer. This is not necessary as we can directly encode the IE value to the message buffer.
The recommended way to do that is to call AvailableBuffer on the bytes.Buffer object to get a []byte buffer with enough capacity, then call Append-style methods to add data to the buffer (without allocating new memory).

### DIFF
--- a/pkg/entities/ie.go
+++ b/pkg/entities/ie.go
@@ -488,8 +488,8 @@ func EncodeToIEDataType(dataType IEDataType, val interface{}) ([]byte, error) {
 		if !ok {
 			return nil, fmt.Errorf("val argument %v is not of type net.IP for this element", val)
 		}
-		if ipv4Add := v.To4(); ipv4Add != nil {
-			return ipv4Add, nil
+		if ipv4Addr := v.To4(); ipv4Addr != nil {
+			return ipv4Addr, nil
 		} else {
 			return nil, fmt.Errorf("provided IP %v does not belong to IPv4 address family", v)
 		}
@@ -499,8 +499,8 @@ func EncodeToIEDataType(dataType IEDataType, val interface{}) ([]byte, error) {
 		if !ok {
 			return nil, fmt.Errorf("val argument %v is not of type net.IP for this element", val)
 		}
-		if ipv6Add := v.To16(); ipv6Add != nil {
-			return ipv6Add, nil
+		if ipv6Addr := v.To16(); ipv6Addr != nil {
+			return ipv6Addr, nil
 		} else {
 			return nil, fmt.Errorf("provided IPv6 address %v is not of correct length", v)
 		}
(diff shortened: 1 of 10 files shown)


antoninbas:
Before change:
```
C02F16ASMD6R:go-ipfix abas$ go test -run=xxx -bench=BenchmarkWriteIPFIXMsgToBuffer -benchmem ./pkg/exporter/
goos: darwin
goarch: amd64
pkg: github.com/vmware/go-ipfix/pkg/exporter
BenchmarkWriteIPFIXMsgToBuffer-16    	  291412	      3545 ns/op	      80 B/op	      10 allocs/op
PASS
ok  	github.com/vmware/go-ipfix/pkg/exporter	51.294s
```

After change:
```
C02F16ASMD6R:go-ipfix abas$ go test -run=xxx -bench=BenchmarkWriteIPFIXMsgToBuffer -benchmem ./pkg/exporter/
goos: darwin
goarch: amd64
pkg: github.com/vmware/go-ipfix/pkg/exporter
BenchmarkWriteIPFIXMsgToBuffer-16    	  378382	      3050 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/vmware/go-ipfix/pkg/exporter	37.357s
```

pkg/entities/ie.go
```diff
@@ -620,3 +620,94 @@ func encodeInfoElementValueToBuff(element InfoElementWithValue, buffer []byte, i
...
+		v := element.GetOctetArrayValue()
+		ieLen := element.GetInfoElement().Len
+		if ieLen < VariableLength {
+			// fixed length case
+			if len(v) != int(ieLen) {
+				return nil, fmt.Errorf("invalid value for fixed-length octet array: length mismatch")
+			}
+			buffer = append(buffer, v...)
+		} else if len(v) < 255 {
+			buffer = append(buffer, byte(len(v)))
```
yuntanghsu, line 637: Forgive me if it is a dumb question. Why do we need to append the length or 255 before the element value?

antoninbas: No worries. For variable-length elements, the encoding is as follows:
- if the length is < 255, write the length to the first byte, then write the binary data
- if the length is >= 255 and < 65,536, write 255 to the first byte, write the actual length to the next 2 bytes (uint16), then write the binary data
- if the length is >= 65,536, it is not supported

This is just IPFIX encoding. 255 is a sentinel value for the first byte.

yuntanghsu: Got it. I was confused with the usage of ieLen and found it just a number to define whether it is a fixed-length or variable-length case/

yuntanghsu (approved):
LGTM

pkg/entities/ie.go
antoninbas: Yes, in the fixed-length case, `ieLen` matches the actual length. In the variable-length case, it is set to 65,535.

PR #398, opened 2025-01-09
Do not embed Record interface into baseRecord

antoninbas:
As far as I know this is an anti-pattern, and I am not sure why it was done this way. baseRecord is just here to help reduce code duplication in dataRecord and templateRecord, which both implement the Record interface.

When embedding Recod in baseRecord, we 1) obfuscate things, making it harder to detect at compile time that one of the record types is missing a method implementation, and 2) increase the size of the baseRecord struct needlessly (after this change the size is reduced from 88B down to 72B on my machine).


yuntanghsu (approved):
LGTM

PR #407, opened 2025-02-11
Add Path MTU to Exporter inputs

antoninbas:
It is not straightforward for the consumer of the library to determine the maximum message size (for UDP) just based on the PMTU, as the amount to deduct from the PMTU depends on the protocol used for the connection to the exporter (IPv4 or IPv6). The protocol is only known after resolving the collector address, which happens in InitExportingProcess. As a workaround, we add PMTU to the exporter inputs, and let the exporter compute the maximum message size. It is still possible to request a specific maximum message size. If both are provided and the exporter determines that the values are incompatible, an error will be returned.

### DIFF
--- a/pkg/exporter/process.go
+++ b/pkg/exporter/process.go
@@ -48,8 +48,7 @@ type templateValue struct {
 //  3. Supports only TCP and UDP; one session at a time. SCTP is not supported.
 //  4. UDP needs to send PMTU size packets as per RFC7011. In order to guarantee
 //     this, maxMsgSize should be set correctly. maxMsgSize is the maximum
-//     payload (IPFIX message) size, not the maximum packet size. You need to
-//     compute maxMsgSize based on the desired maximum packet size. If
+//     payload (IPFIX message) size, not the maximum packet size. If
 //     maxMsgSize is not set correctly, the message may be fragmented.
 type ExportingProcess struct {
 	connToCollector net.Conn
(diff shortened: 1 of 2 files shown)


pkg/exporter/process.go
```diff
@@ -94,17 +93,68 @@ type ExporterInput struct {
...
+			return entities.MaxSocketMsgSize, nil
+		} else {
+			return requestedSize, nil
+		}
+	}
+	// UDP protocol
+	if pathMTU == 0 {
+		if requestedSize == 0 {
+			klog.InfoS("Neither max IPFIX message size nor PMTU were provided, defaulting to min message size", "messageSize", entities.MinSupportedMsgSize)
+			return entities.MinSupportedMsgSize, nil
```
yuntanghsu, line 126: Why do we use MinSupportedMsgSize for udp but using MaxSocketMsgSize for tcp when requestedSize is 0? Or we want to set it to a minimum value to prevent threshold? But I think we returned MaxSocketMsgSize when MaxMsgSize is 0 no matter which protocol is provided previously?

antoninbas: Because UDP is based on datagrams and when we call `ep.connToCollector.Write`, the data we provide will be send immediately as one packet / datagram. The IPFIX spec stays that these UDP packets should not be fragmented, so when the PMTU is not provided and when the user doesn't provide an explicit message size, we default to a very conservative value (`MinSupportedMsgSize`) of 512B which should always avoid fragmentation.

For TCP, the message size is not really important as TCP is segment based. When we call `ep.connToCollector.Write`, we add data to the TCP data stream, and we don't really control packet boundaries. `MaxSocketMsgSize` is set to 65KB, and that's actually the maximum size for an IPFIX message, given that the IPFIX message length is encoded in the header as a uint16. Any message size which is less than `MaxSocketMsgSize` will work just fine, and when no value is provided by the user, there is no real reason to use a smaller value.

yuntanghsu (approved):
LGTM

PR #409, opened 2025-02-12
Add buffered exporter for IPFIX messages

antoninbas:
SendDataRecords was added in #391 as a way to export data records more efficiently. However, this method is not convenient to use for clients which need to mutate the records after sending them (as is the case when aggregation is used). To work around this issue, we introduce a buffered exporter. The buffered exporter wraps the standard exporter process. Whenever a data record is added, it is serialized to a byte buffer immediately, and the buffer is sent when it reaches the maximum message size.

Note that for convenience the buffered exporter also supports JSON records, but in this case it just acts as a passthrough to the regular ("unbuffered") exporter process.

### DIFF
--- a/pkg/exporter/buffered.go
+++ b/pkg/exporter/buffered.go
@@ -86,7 +86,7 @@ func (e *BufferedIPFIXExporter) addDataRecord(record entities.Record) error {
 	return m.addRecord(record)
 }
 
-// AddRecord adds a record to be sent to the destination collector. If it is a template record, the
+// AddRecord adds a record to be sent to the destination collector. If it is a template record, then
 // it will be sent to the collector right away. If it is a data record, it will be added to the
 // buffer. If adding the record to the buffer would cause the buffer length to exceed the max
 // message size, the buffer is flushed first. Note that because data records are serialized to the
--- a/pkg/exporter/buffered.go
+++ b/pkg/exporter/buffered.go
@@ -1,0 +1,186 @@
+// Copyright 2025 VMware, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package exporter
+
+import (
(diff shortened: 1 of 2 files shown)


pkg/exporter/buffered.go
```diff
@@ -0,0 +1,186 @@
...
+	m, ok := e.messages[templateID]
+	if ok {
+		return m.addRecord(record)
+	}
+	m = newBufferedMessage(e.ep, templateID)
+	e.messages[templateID] = m
+	return m.addRecord(record)
+}
+
+// AddRecord adds a record to be sent to the destination collector. If it is a template record, the
```
yuntanghsu, line 89:
```suggestion
// AddRecord adds a record to be sent to the destination collector. If it is a template record, then
```

pkg/exporter/buffered.go
```diff
@@ -0,0 +1,186 @@
...
+// before sending them. BufferedIPFIXExporter is not safe for usage by multiple goroutines. There
+// should be a single BufferedIPFIXExporter created for a given ExportingProcess.
+// While the BufferedIPFIXExporter supports sending JSON records, in that case it mostly acts as a
+// passthrough to the underlying ExportingProcess (no actual buffering).
+type BufferedIPFIXExporter struct {
+	ep          *ExportingProcess
+	templateSet entities.Set
+	// maps templateID to the corresponding buffer for data records. Note that entries are never
+	// deleted from this map.
+	messages   map[uint16]*bufferedMessage
```
yuntanghsu, line 36: I think this BufferedIPFIXExporter is the middle layer we were talking about? It is used for aggregation after we collect the record and before we export the data to IPFIX?
I'm just curious about if we don't expose messages, how do user mutate the record?

antoninbas: If you are referring to https://github.com/vmware/go-ipfix/pull/389, then no this is an orthogonal change. This is the middle layer in Antrea: https://github.com/antrea-io/antrea/blob/main/pkg/flowaggregator/preprocessor.go. In the end, I couldn't go with my original approach, and I ended up "duplicating" records (in practice this is not that expensive) and creating new messges.

This code (`BufferedIPFIXExporter`) is unrelated. It is a wrapper around the regular exporter process which allows for better performance by buffering data records and generating IPFIX messages on the fly. It takes records as inputs, but this is what is produced by the aggregation process (not messages).
Think of this as a buffered writer in the standard Go library: https://pkg.go.dev/bufio#NewWriter. It wraps another writer object and provides improved performance thanks to buffering / batching. But it typically requires the addition of a `Flush` method.

yuntanghsu: Got it. I misunderstood these two idea.. Thanks for the explanation

yuntanghsu (approved):
LGTM

PR #410, opened 2025-02-14
Improve collector program for data sets with multiple records

antoninbas:
When multiple data records are present in a set / IPFIX message, we need to "split" them so that the API returns them as individual records.

We also drop template records to simplify the implementation.


yuntanghsu (approved):
LGTM

PR #411, opened 2025-02-14
Fix template record handling in BufferedExporter

antoninbas:
The record was dropped instead of being added to the template set. Additionally, when a template is sent, buffered messages for the same template ID should be flushed first.

### DIFF
--- a/pkg/exporter/buffered.go
+++ b/pkg/exporter/buffered.go
@@ -68,8 +68,21 @@ func NewBufferedIPFIXExporter(ep *ExportingProcess) *BufferedIPFIXExporter {
 }
 
 func (e *BufferedIPFIXExporter) addTemplateRecord(record entities.Record) error {
+	templateID := record.GetTemplateID()
+	// If the templateID already exists, we should send corresponding buffered data records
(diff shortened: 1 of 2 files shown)


pkg/exporter/buffered.go
```diff
@@ -68,8 +68,21 @@ func NewBufferedIPFIXExporter(ep *ExportingProcess) *BufferedIPFIXExporter {
 }
 
 func (e *BufferedIPFIXExporter) addTemplateRecord(record entities.Record) error {
+	templateID := record.GetTemplateID()
+	// If the templateID already exists, we should send corresponding buffered data records
+	// immediately, as they may not match the new template definition.
+	if m, ok := e.messages[templateID]; ok {
+		if err := m.flush(); err != nil {
```
yuntanghsu, line 75: Should we return after flushing? Otherwise I think we will have duplicate records?

antoninbas: no, the new template should be sent and will override the old template, both locally in the exporter (`templateMap`) and at the destination collector

yuntanghsu (approved):
LGTM

PR #417, opened 2025-03-20
Fix Record length computation

antoninbas:
In several cases, GetRecordLength() was returning an invalid value:
- for "decoded" records, the method would always return 0.
- when mutating an IE value with a variable length (e.g., a string element), the cached record length would no longer be correct and the method would return an invalid value.

To avoid this issue, we stop caching the length as part of the Record object, and instead we recompute the length whenever GetRecordLength() is invoked. Typically, this method is only invoked once for a given Record, at the time of exporting the Record, so there should not be any performane impact in a real-life scenario.

We also remove the buffer field from DataRecord (we keep it for TemplateRecord), and we remove the isDecoding field for all records, as it was only used for length tracking and is confusing. For now, we do not modify the Set implementation, although in the long term we should remove isDecoding everywhere and use a standard Marshal / Unmarshal approach.

### DIFF
--- a/pkg/entities/record.go
+++ b/pkg/entities/record.go
@@ -46,42 +46,30 @@ type Record interface {
 }
 
 type baseRecord struct {
-	buffer             []byte
 	fieldCount         uint16
 	templateID         uint16
 	orderedElementList []InfoElementWithValue
-	isDecoding         bool
-	len                int
 }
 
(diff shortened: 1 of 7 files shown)


yuntanghsu (approved):
LGTM

PR #418, opened 2025-03-21
Fix sequence number incrementing in IPFIX exporter

antoninbas:
The sequence number should be the total number of IPFIX Data Records sent prior to the current message.


yuntanghsu (approved):
LGTM

PR #423, opened 2025-04-15
Support for more TLS config options in Collector and Exporter

antoninbas:
We support providing a list of cipher suites, as well as a min TLS protocol version.
For DTLS, the version cannot be configured (only 1.2 is supported).

### DIFF
--- a/pkg/exporter/process.go
+++ b/pkg/exporter/process.go
@@ -634,11 +634,6 @@ func (ep *ExportingProcess) dataRecSanityCheck(rec entities.Record) error {
 }
 
 func createClientConfig(config *ExporterTLSClientConfig) (*tls.Config, error) {
-	roots := x509.NewCertPool()
-	ok := roots.AppendCertsFromPEM(config.CAData)
-	if !ok {
-		return nil, fmt.Errorf("failed to parse root certificate")
-	}
 	tlsMinVersion := config.MinVersion
 	// This should already be the default value for tls.Config, but we duplicate the earlier
 	// implementation, which was explicitly setting it to 1.2.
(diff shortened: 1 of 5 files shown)


pkg/exporter/process.go
```diff
@@ -172,17 +181,26 @@ func InitExportingProcess(input ExporterInput) (*ExportingProcess, error) {
 		} else if input.CollectorProtocol == "udp" { // use DTLS
 			// TODO: support client authentication
 			if len(tlsConfig.CertData) > 0 || len(tlsConfig.KeyData) > 0 {
-				klog.Error("Client-authentication is not supported yet for DTLS, cert and key data will be ignored")
+				return nil, fmt.Errorf("client-authentication is not supported yet for DTLS")
+			}
+			if tlsConfig.MinVersion != 0 && tlsConfig.MinVersion != tls.VersionTLS12 {
```
yuntanghsu, line 186: I think the tls version would be decided by server? Do we need to specify the minTlsVersion for client? Or it's not harmful to set it on both side?

yuntanghsu (dismissed):

pkg/exporter/process.go
```diff
@@ -616,26 +634,35 @@ func (ep *ExportingProcess) dataRecSanityCheck(rec entities.Record) error {
...
+	if config.CAData != nil {
+		roots := x509.NewCertPool()
+		ok := roots.AppendCertsFromPEM(config.CAData)
+		if !ok {
+			return nil, fmt.Errorf("failed to parse root certificate")
+		}
+		tlsConfig.RootCAs = roots
+	}
+	// Don't use a client certificate if config.CertData == nil.
+	if config.CertData != nil {
```
yuntanghsu, line 660: Do we need to check KeyData != nil as well?

pkg/exporter/process.go
```diff
@@ -616,26 +634,35 @@ func (ep *ExportingProcess) dataRecSanityCheck(rec entities.Record) error {
...
+	}
+	// #nosec G402: client is in charge of setting the min TLS version. We use 1.2 as the
+	// default, which is secure.
+	tlsConfig := &tls.Config{
 		ServerName:   config.ServerName,
-	}, nil
+		CipherSuites: config.CipherSuites,
+		MinVersion:   tlsMinVersion,
+	}
+	// Use system roots if config.CAData == nil.
```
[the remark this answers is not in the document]
antoninbas: If `config.CAData == nil`, then we do not set `tlsConfig.RootCAs`, so it is `nil`. In that case Go will use the system CAs by default:
> If RootCAs is nil, TLS uses the host's root CA set.

pkg/exporter/process.go
yuntanghsu: I saw you do this check for the collector. Maybe we should do the same way for consistency? But I don't have strong opinion on that.

PR #427, opened 2025-06-07
Do not close the udp connection to the collector if there are errors while sending

vlet:
Fixed the problem with permanently closing the connection to the collector (ipfix/udp mode), if the collector was unavailable for some time. Since the mechanism for repeated reconnection is not implemented, the exporter no longer closes the connection to the collector, but only reports the problem in the log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.  The format
 Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
+### Changed
+- Do not close the UDP connection to the collector if there are errors while sending (@vlet)
 ## 0.14.0 03-24-2025
 ### Fixed
 - Fix Record length computation so that IPFIX UDP messages can be sized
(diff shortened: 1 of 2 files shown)


vlet:
> Thanks for opening this PR. I had a similar change about a year ago, but ended up reverting it after a while: #360 At least for TCP I decided that it wasn't necessarily the way to go. I don't have a strong opinion for UDP, but it does introduce some asymmetry. By the way, what kind of transient error(s) are you trying to address?

I use netobserv-ebpf-agent outside of the kubernetes environment on linux routers as a systemd service. Netflow data in ipfix/udp format is sent to the nfcapd collector along with other netflow flows from network equipment. When the nfcapd service is restarted or the collector server is rebooted, the udp port stops listening for some time. When receiving data, the OS generates an icmp response port unreacheable, which leads to the fact that netobserv-ebpf-agent considers the connection closed and stops sending netflow data until the netobserv-ebpf-agent service is manually restarted.

On the other hand, network equipment (cisco, mikrotik ...) ignores such icmp responses from the server and sends netflow data endlessly. If the udp protocol is selected, it means that losses and delivery order are not critical. Probably, the implementation on network equipment follows this principle.

As a workaround, I configured blocking of outgoing icmp responses on the collector's firewall, it helped, but I want a more reliable solution.

If netobserv-ebpf-agent periodically reconnected to the collector, this problem would not exist. For a TCP connection, this is the only correct way, but in the case of UDP, this is not necessary, since formally in UDP there is no procedure for establishing/closing a connection.

PR #433, opened 2025-07-02
Add new IEs for K8s object UUIDs to Antrea registry

antoninbas:
The data type is octetArray, with a fixed size of 16 bytes.

We are using "UUID" and not "UID" for the following reasons:
  - we want to use a more compact representation than the default string
    representation for K8s UIDs. Using "UUID" highlights the difference.
  - we already have 2 existing IEs, ingressNetworkPolicyUID and
    egressNetworkPolicyUID, which use a string representation. We do not
    want to change the definition of these IEs, even though they may not
    be actively used.
  - in the future, we may want to introduce new IEs to encode the string
    representation of the UID, so it makes sense to reserve "UID" for this
    scenario.

A new column (Data size) is added to the Antrea registry definition file, registry_antrea.csv, so that a fixed length can be provided, even though this column does not exist in the IANA csv file. In the IANA csv file, that information is included as part of the IE description, which is not practical.

Note that registry_IANA.go is also updated (automatically), as new IEs have been registered since the last time the source code was generated.


yuntanghsu (commented):
LGTM

yuntanghsu (approved):

PR #434, opened 2025-07-04
UDP: refresh connection at the same time as the templates refresh

jotak:
Hi,
here's a suggestion to refresh the UDP connection using the same templates refreshing ticker , what do you think about it?
A typical use case is when the collector host restarts, the connectivity might be lost. Recreating the connection would restore it. I think it makes sense to tie that with the templates refresh, since it's typically the same cause that would require resending the templates (hosts restart)


jotak:
I'll close this PR since it gets no attention, but happy to reopen if it's considered useful

